PR #263, opened 2015-01-08
[sailfish-browser] Enable overlay content when it is fully opened

rainemak:
Favorite grid and history list are enabled only after overlay
is atTop state (fully opened).


rojkov:
LGTM

PR #264, opened 2015-01-08
[sailfish-browser] Position overlay flickables to beginning when overlay is hidden

rainemak:
Position both favorite grid and history list to the beginning when
overlay reaches bottom.


rojkov:
LGTM, tested with a number of  favorites taking more then one screen

PR #265, opened 2015-01-08
[sailfish-browser] Treat window as CoverWindow when web content is shown in cover

rainemak:


rojkov:
LGTM, no black frame visible with this patch applied

PR #266, opened 2015-01-09
First usage improvaments

rainemak:
Contains two commits:

[sailfish-browser] Mark first time usage as done when a tab is activated
[sailfish-browser] Load incoming url also when first use is not done


rojkov:
LGTM given that opening an URL from an external app with Overlay on top when browser is running is supposed to be in another fix.

PR #269, opened 2015-01-09
[sailfish-browser] Do not limit dragging of overlay in fullscreen mode

rainemak:
Overlay cannot be opened by dragging meaning that dragging can be only used
to lower the overlay. Thus, we should not disable dragging in fullscreen
mode.

For instance, when playing a video in fullscreen and a new tab requested
from cover it should be possible to close the overlay by dragging it.


rojkov:
LGTM

PR #270, opened 2015-01-10
Do not force focus on url field

kimmoli:


rainemak:
Thanks @kimmoli for your first PR.

Alternatively you can close virtual keyboard by panning the favorite list.

Personally I'd like it more like it is now and this is what was requested by design. But I see the point that when using heavily favorites current approach might be annoying. At some point of time we had a pull up gesture for the toolbar to open the overlay and with the gesture we could have skipped focusing of the search field.

Will check this with design again.

PR #272, opened 2015-01-16
[sailfish-browser] Add error handling for favorite icons.

rainemak:
In case the server is not responding or favorite icon is not found
from server icon will be replaced by default icon.


mikkoharju:
LGTM

PR #274, opened 2015-01-19
Fix initial download url

rainemak:
Subject: [sailfish-browser] Do not add initial url to the tab model

Initial url can be a download url.

Subject: [sailfish-browser] At startup wait that Download Manager is initialized

Do not mark DeclarativeWebContainer as initialized before mozilla context,
download manager, web container itself, and tab model are completed/initialized.


rojkov:
LGTM. A new empty tab opens, the pdf is not downloaded again.

PR #275, opened 2015-01-19
[sailfish-browser] Switch to grabbed web content while popping tab page

rainemak:
Previously there was a glimpse of previous active tab visible in the
tab index that will be the new active tab. This happened because
activeTab property change changed texture source to use incorrect
sourceItem. Thus, let's jump to grabbed web content while tab page is animated
away from the pagestack.


rojkov:
LGTM

rainemak:
Obsolete

PR #276, opened 2015-01-20
[unittests] Check conditionally title change in tst_webview

rainemak:
Added conditional checking of the title change when loading
a url in tst_webview.


rojkov:
LGTM
All tests in tst_webview pass.

PR #282, opened 2015-01-30
Notify engine about VKB state

rojkov:
This PR depends on xulrunner 31.3.0.4 and the commit
https://github.com/nemomobile-packages/embedlite-components/commit/1ffd093cbf31bf9bc90b603cdf2b74e3b7d08e20

### DIFF
--- a/src/declarativewebcontainer.cpp
+++ b/src/declarativewebcontainer.cpp
@@ -47,9 +47,6 @@ DeclarativeWebContainer::DeclarativeWebContainer(QQuickItem *parent)
     , m_portrait(true)
     , m_fullScreenMode(false)
     , m_fullScreenHeight(0.0)
-    , m_inputPanelVisible(false)
-    , m_inputPanelHeight(0.0)
-    , m_inputPanelOpenHeight(0.0)
     , m_toolbarHeight(0.0)
     , m_tabId(0)
     , m_loading(false)
@@ -66,7 +63,6 @@ DeclarativeWebContainer::DeclarativeWebContainer(QQuickItem *parent)
     connect(DownloadManager::instance(), SIGNAL(downloadStarted()), this, SLOT(onDownloadStarted()));
     connect(QMozContext::GetInstance(), SIGNAL(onInitialized()), this, SLOT(initialize()));
     connect(this, SIGNAL(portraitChanged()), this, SLOT(resetHeight()));
-    connect(this, SIGNAL(enabledChanged()), this, SLOT(handleEnabledChanged()));
 
     QString cacheLocation = QStandardPaths::writableLocation(QStandardPaths::CacheLocation);
     QDir dir(cacheLocation);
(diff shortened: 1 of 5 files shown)


src/declarativewebcontainer.cpp
```diff
@@ -872,3 +875,39 @@ void DeclarativeWebContainer::sendVkbOpenCompositionMetrics()
         m_webPage->sendAsyncMessage("embedui:vkbOpenCompositionMetrics", data);
     }
 }
+
+void DeclarativeWebContainer::notifyVkbFullyOpen()
+{
+    int vkbHeight = 0;
+    int winHeight = 0;
+
+    if (width() < height()) {
```
rainemak: ```if (!m_portrait) {```

src/declarativewebcontainer.cpp
```diff
@@ -872,3 +875,39 @@ void DeclarativeWebContainer::sendVkbOpenCompositionMetrics()
         m_webPage->sendAsyncMessage("embedui:vkbOpenCompositionMetrics", data);
     }
 }
+
+void DeclarativeWebContainer::notifyVkbFullyOpen()
+{
```
rainemak:
```
if (!window()) {
    return;
}
```
Not 100% sure if this is possible even in theory.

src/declarativewebcontainer.cpp
```diff
@@ -77,6 +78,7 @@ DeclarativeWebContainer::DeclarativeWebContainer(QQuickItem *parent)
     connect(this, SIGNAL(heightChanged()), this, SLOT(sendVkbOpenCompositionMetrics()));
     connect(this, SIGNAL(widthChanged()), this, SLOT(sendVkbOpenCompositionMetrics()));
```
rainemak: Could we now remove these connections?

rainemak:
Looks to me that sendVkbOpenCompositionMetrics calculates incorrectly

rojkov:
I've addressed the comments and in addition to them I've dropped updateVkbHeight() and added orientation change handling.

The corresponding patch in embedlite-components is https://github.com/rojkov/embedlite-components/commit/e1636fe7ad38b9fc18e3651bee2a42da3c89e698

rainemak:
Ran through some random web pages. Here are some observations:

- m_inputPanelOpenHeight should be reset to zero when input panel is focused as virtual keyboard size can vary (digits, landscape/portrait)
- setter of inputPanelHeight (read notify) could be removed from QML side. We could update actual target values immediately when QInputMethod::visible() changes (per orientation).
- If above approach would work, then WebView could binding directly to Qt.inputMethod.visible ~ WebView.qml@298
- regarding the zooming delay, we could calculate target viewport size when virtual keyboard starts opening (when im visibility changes). Send that information immediately to the engine side and handle fully opened state there

rojkov:
I've dropped the inputPanel related properties. So far no regressions.

Sending vieport size right after VKB starts to open isn't better perfromance-wise:
1. currently metrics updates in AZPC are propagated synchronously to sailfish-browser C++ code without conversions
2. handling the updates in embedhepler.js would require data transformations C++ <-> JS.

rojkov:
clean up open PRs

PR #283, opened 2015-02-04
Tiling texture to indicate private browsing mode

roppola:
this time not directly to master :)


rainemak:
@siteshwar here is @roppola 's latest private browsing texture.

LGTM

PR #285, opened 2015-02-06
Build with multiple cores when using mb2.

rburchell:
jobs is non-standard and won't use multiple cores when building under mb2, unlike _smp_mflags.


siteshwar:
Ship it!

rainemak:
LGTM

PR #286, opened 2015-02-08
[sailfish-browser] Add support for private browsing

### LINKED ISSUE #202: Private browsing support
Enable private browsing support.

As we have a history that is at browser side we cannot write to the history when private browsing is enabled.

Preference:
"browser.privatebrowsing.autostart"

Some pointers to Gecko:

* https://github.com/tmeshkova/gecko-dev/blob/embedlite/embedding/components/windowwatcher/nsWindowWatcher.cpp#L424

* https://github.com/tmeshkova/gecko-dev/blob/embedlite/dom/ipc/TabChild.cpp#L1012

* https://github.com/tmeshkova/gecko-dev/blob/embedlite/dom/ipc/TabChild.cpp#L812

siteshwar:
Fixes #202


src/declarativewebcontainer.cpp
```diff
@@ -187,7 +209,10 @@ bool DeclarativeWebContainer::background() const
 
 bool DeclarativeWebContainer::loading() const
 {
-    return m_webPage ? m_webPage->loading() : m_model->count();
+    if (m_webPage)
+        return m_webPage->loading();
```
rojkov: coding style: we use curly brackets even for one line blocks. It may save time when qDebug() is inserted before the statement mistakenly without brackets.

src/declarativewebcontainer.cpp
```diff
@@ -739,6 +778,13 @@ void DeclarativeWebContainer::setActiveTabData()
     }
 }
 
+void DeclarativeWebContainer::setWebPages() {
+    if (m_privateMode)
```
rojkov: brackets

src/pages/components/Overlay.qml
```diff
@@ -415,15 +442,23 @@ PanelBackground {
...
 
             Browser.TabView {
                 model: webView.tabModel
                 portrait: tabPage.isPortrait
+                privateMode: webView.privateMode
 
                 onHide: pageStack.pop()
+
+                onPrivateModeChanged: {
+                    webView.privateMode = privateMode;
```
rojkov: There is a belief that explicit statement separator (the semicolons) do cognitive overload with information noise and thus make developers less productive. The majority of SailfishOS developers are the believers so we'd better not use semicolons in QML for the sake of consistency.

siteshwar:
@rojkov Fixed through c2766a8074fa6288fe73b8bb9ed806e60f3480c2

rojkov:
The code looks good to me. I've noticed two problems when trying the patch:

 1. No "private Browsing" label in landscape, but there's no such label in the UI specs. Might not be needed.
 2. If to a) open a web page in the normal mode, b) to open another webpage in the private mode, c) to return to normal mode back again and d) to tap the first page in the TabsView then the Overlay with Favorites and SearchField is shown instead of the webpage.

siteshwar:
I agree with point 1 but not able to reproduce point 2. Is this issue always reproducible for you ?

rojkov:
Yes, it's reproducible in 100%

siteshwar:
I tried several times but was not able to reproduce this issue. I am running Sailfish 1.1.1.27 on my device which may be different from your development device environment.

rojkov:
Now it looks good to me.

Though without https://github.com/sailfishos/sailfish-browser/pull/288 accepted persistent web views get cleared upon returning to the normal mode.

src/declarativewebcontainer.h
```diff
@@ -101,6 +106,7 @@ class DeclarativeWebContainer : public QQuickItem {
     QString thumbnailPath() const;
 
     bool isActiveTab(int tabId);
+
```
rainemak: Extra

src/pages/components/ConfigDialog.qml
```diff
@@ -1,6 +1,6 @@
 /****************************************************************************
 **
-** Copyright (C) 2013
+** Copyright (C) 2014
```
rainemak: Maybe even 2014, 2015 :)

src/pages/components/Overlay.qml
```diff
@@ -158,6 +158,29 @@ PanelBackground {
...
+        id: privateModeLabel
+
+        PrivateModeTexture {
+            anchors.fill: parent
+        }
+
+        //: Label for private browsing above address bar
+        //% "Private browsing"
+        text: qsTrId("sailfish_browser-la-private_mode")
+        visible: searchField.visible && webView.privateMode
```
rainemak: Fade this out.

```
opacity: searchField.visible && webView.privateMode ? 1.0 : 0.0
visible: opacity > 0.0
...
font.pixelSize: Theme.fontSizeLarge
...
Behavior on opacity { FadeAnimation {} }
```

src/pages/components/PrivateModeTexture.qml
```diff
@@ -0,0 +1,19 @@
...
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.1
+
+Image {
+    anchors.fill: parent
+    fillMode: Image.Tile
+    source: "graphic-diagonal-line-texture.png"
+}
+
```
rainemak: Extra line

src/pages/components/PrivateModeTexture.qml
```diff
@@ -0,0 +1,19 @@
...
+**
+****************************************************************************/
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.1
+
+Image {
```
rainemak:
```
property bool active
opacity: active ? 1.0 : 0.0
visible: opacity > 0.0
```
...
source: "graphic-diagonal-line-texture.png"
Behavior on opacity { FadeAnimation {} }
```

File: src/pages/components/Overlay.qml [hunk 9]
```diff
@@ -211,6 +234,10 @@ PanelBackground {
             height: toolBar.toolsHeight + historyList.height
             clip: true
 
+            PrivateModeTexture {
+                visible: toolBar.visible && webView.privateMode
```
rainemak: active: toolBar.visible && webView.privateMode

src/declarativewebcontainer.cpp
```diff
@@ -59,14 +60,26 @@ DeclarativeWebContainer::DeclarativeWebContainer(QQuickItem *parent)
...
 {
-    m_webPages.reset(new WebPages(this));
     setFlag(QQuickItem::ItemHasContents, true);
+    setPrivateMode(m_settingManager->autostartPrivateBrowsing());
+
+    m_normalWebPages.reset(new WebPages(this));
+    m_privateWebPages.reset(new WebPages(this));
+
+    m_persistentTabModel.reset(new PersistentTabModel());
+    m_privateTabModel.reset(new PrivateTabModel());
```
rojkov: Here it makes sense to set instances' parent to `this`

m_persistentTabModel.reset(new PersistentTabModel(this));
m_privateTabModel.reset(new PrivateTabModel(this));

src/pages/components/Overlay.qml
```diff
@@ -158,6 +158,29 @@ PanelBackground {
         source: "image://theme/graphic-gradient-edge"
     }
 
+    Label {
+        id: privateModeLabel
+
+        PrivateModeTexture {
+            anchors.fill: parent
```
rainemak: Not filling the space above overlay. Visible when dragging overlay down.

Should be filling Rectangle @ BrowserPage.qml 123. Could be added as a children of it.

From out side of the Overlay the texture is active when on private mode and toolBar not visible...
```
active: !overlay.toolBar.visible && webView.privateMode
```

src/pages/components/PrivateModeTexture.qml
```diff
@@ -0,0 +1,19 @@
...
+** Copyright (C) 2015 Jolla Ltd.
+** Contact: Siteshwar Vashisht <siteshwar@gmail.com>
+**
+****************************************************************************/
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.1
```
rainemak: import Sailfish.Silica 1.0 for FadeAnimation

src/pages/components/Overlay.qml
```diff
@@ -158,6 +158,29 @@ PanelBackground {
         source: "image://theme/graphic-gradient-edge"
     }
 
+    Label {
+        id: privateModeLabel
```
rainemak: The label could be also at the same place as the PrivateModeTexture should be, see below.

@ BrowserPage.qml after the MouseArea ~133

```
        Browser.PrivateModeTexture {
            anchors.fill: parent
            active: !overlay.toolBar.visible && webView.privateMode

            Label {
                id: privateModeLabel
                x: Theme.paddingLarge
                // Allow only half of the width
                width: parent.width / 2
                truncationMode: TruncationMode.Fade
                anchors {
                    bottom: parent.bottom
                    bottomMargin: (overlay.toolBar.toolsHeight - height) / 2
                    horizontalCenter: parent.horizontalCenter
                }

                //: Label for private browsing above address bar
                //% "Private browsing"
                text: "Private browsing"
                color: Theme.highlightColor
                font.pixelSize: Theme.fontSizeLarge
            }
        }
    }
```

src/pages/components/PrivateModeTexture.qml
```diff
@@ -0,0 +1,19 @@
...
+
+Image {
+    anchors.fill: parent
+    fillMode: Image.Tile
+    source: "graphic-diagonal-line-texture.png"
```
rainemak: visible: opacity > 0.0

src/pages/components/Overlay.qml
rainemak: Actually above should be sibling of the Rectangle.... So ```anchors.fill: contentDimmer``` given that Rectangle at BrowserPage 123 has an id of contentDimmer.

src/pages/components/Overlay.qml
rainemak:
```
opacity: toolBar.visible && webView.privateMode ? toolBar.opacity : 0.0
visible: opacity > 0.0
```

src/pages/components/ConfigDialog.qml
VDVsx: Correct form used in oss licenses is 2014-2015 (year of first realease - year of last release)

rainemak:
Two generic comments:
- Don't like that fact that private tabs are closed immediately when browsing mode is changed. Instead I'd activate the active page when mode is changed without closing the tabs. I think that private tabs should be closed after browser is closed.

- A normal tab can be accessed from private mode like:
    1) Close all tabs
    2) Active "normal" browsing
    3) Enter m.engadget.com
    4) Go to tabs view and switch to private browsing
    5) Pop the tabs view from the pagestack with left-right swipe gesture

    * It should not be possible to pop the tab view unless there are pages available in respected mode.

src/declarativewebcontainer.cpp
```diff
@@ -59,14 +60,26 @@ DeclarativeWebContainer::DeclarativeWebContainer(QQuickItem *parent)
...
+
+    setWebPages();
+
     connect(DownloadManager::instance(), SIGNAL(initializedChanged()), this, SLOT(initialize()));
     connect(DownloadManager::instance(), SIGNAL(downloadStarted()), this, SLOT(onDownloadStarted()));
     connect(QMozContext::GetInstance(), SIGNAL(onInitialized()), this, SLOT(initialize()));
     connect(this, SIGNAL(portraitChanged()), this, SLOT(resetHeight()));
     connect(this, SIGNAL(enabledChanged()), this, SLOT(handleEnabledChanged()));
 
+    connect(this, SIGNAL(privateModeChanged()), this, SLOT(updateMode()));
```
rainemak: updateMode() should be called directly from setPrivateMode before emitting change signal. When change signal is emitted all properties should match to that mode.

src/persistenttabmodel.cpp
```diff
@@ -0,0 +1,150 @@
...
+int PersistentTabModel::nextLinkId() {
+    return DBManager::instance()->nextLinkId();
+}
+
+void PersistentTabModel::updateTab(int tabId, QString url, QString title, QString path) {
+    DBManager::instance()->updateTab(tabId, url, "", "");
+}
+
+void PersistentTabModel::navigateTo(int tabId, QString url, QString title, QString path) {
+    DBManager::instance()->navigateTo(tabId, url, "", "");
```
rojkov: I've noticed build warnings about unused title and path. Could you please add Q_UNUSED() to updateTab() and navigateTo()?

rojkov:
LGTM

rojkov:
Not to forget:

Now we have two instances of WebPages (and two WebPagesQueue instances) per mode. As result we can't have a single value for maxLivePages and we can't easily suspend/resume a view when switching to another mode. It makes sense to have only one instance of WebPages for DeclarativeWebContainer, but this requires that all tabs (private and non-private) have unique IDs across the app. Theoretically persistent tabIds can reach 10000.
This uniqueness can be achieved either by replacing integer IDs with UUIDs or by introducing a single source of new tabIds.

src/pages/BrowserPage.qml
```diff
@@ -131,6 +132,33 @@ Page {
...
+
+        Browser.PrivateModeTexture {
+            id: privateModeTexture
+            anchors.fill: contentDimmer
+            visible: webView.privateMode && !overlay.animator.allowContentUse
+        }
+    }
+
+    Label {
+        x: Theme.paddingLarge
```
rainemak: Actually this should be:
```x: (contentDimmer.width - implicitWidth) / 2```

File: src/pages/BrowserPage.qml [hunk 18]
```diff
@@ -131,6 +132,33 @@ Page {
...
+        // Allow only half of the width
+        width: parent.width / 2
+        truncationMode: TruncationMode.Fade
+        opacity: privateModeTexture.visible ? 1.0 : 0.0
+        anchors {
+            bottom: contentDimmer.bottom
+            bottomMargin: (overlay.toolBar.toolsHeight - height) / 2
+            horizontalCenter: contentDimmer.horizontalCenter
```
rainemak: Due to above remove this horizontalCentering.

siteshwar:
Done!

src/declarativewebcontainer.cpp
```diff
@@ -380,6 +408,7 @@ bool DeclarativeWebContainer::activatePage(int tabId, bool force, int parentId)
         connect(m_webPage, SIGNAL(backgroundChanged()), this, SIGNAL(backgroundChanged()), Qt::UniqueConnection);
         return activationData.activated;
     }
+
```
rainemak: Empty line.

src/declarativewebcontainer.cpp
```diff
@@ -78,6 +86,7 @@ DeclarativeWebContainer::DeclarativeWebContainer(QQuickItem *parent)
     connect(this, SIGNAL(widthChanged()), this, SLOT(sendVkbOpenCompositionMetrics()));
     connect(this, SIGNAL(foregroundChanged()), this, SLOT(updateWindowFlags()));
 
+
```
rainemak: Empty line

src/pages/components/Overlay.qml
```diff
@@ -415,15 +419,23 @@ PanelBackground {
...
                 model: webView.tabModel
                 portrait: tabPage.isPortrait
+                privateMode: webView.privateMode
 
                 onHide: pageStack.pop()
+
+                onPrivateModeChanged: {
+                    webView.privateMode = privateMode
+                    tabPage.activeTabIndex =  webView.tabModel.activeTabIndex
+                    tabPage.activeWebPage = webView.contentItem
```
rainemak: if (webView.tabModel.count === 0) {
                        overlay.enterNewTabUrl(PageStackAction.Immediate)
                    }

rainemak:
Added some minor comments and some of them you already fixed. 

WebView unit tests are occasionally failing. There should be a test mode that would not use dconf value or mock version of settings manager.

Other than that LGTM

src/pages/components/Overlay.qml
siteshwar: With above code consider following sequence of events :

1) Normal browsing active
2) Restart browser
3) Switch to private mode
4) Switch back to normal mode
4) Page pop the page

the box to enter url is visible even when tabmodel in normal browsing mode is not empty.

I propose below code :
```
                    if (webView.tabModel.count === 0) {
                        overlay.enterNewTabUrl(PageStackAction.Immediate)
                    } else {
                        if (overlayAnimator.state !== "fullscreenWebPage") {
                            dismiss()
                        }
                    }
```

siteshwar:
We can explicitly set privateMode: false in ```tests/auto/tst_webview/tst_webview.qml```. Should that be sufficient ?

src/pages/components/Overlay.qml
rainemak: Agree.. 

Actually like...
```
                    if (webView.tabModel.count === 0) {
                        overlay.enterNewTabUrl(PageStackAction.Immediate)
                    } else if (!overlayAnimator.atBottom) {
                         dismiss()
                    }
```
I started thinking that with private/non-private modes we could actually stay in tab view when the last tab is closed from tab view regardless of the mode. That would imply that whenever tab count changes to zero (regardless of the mode) and we're at tab we'd execute enterNewTabUrl. When privateMode changes we could only dismiss overlay if tab count more than zero. This approach would be imo more symmetric.

Let's see what @roppola thinks :)

siteshwar: Fixed through 77d1ffb0098f7eb9a7d03514c5b146edc6fb8af0 and I agree that we should stay in  tabview after closing all tabs.

src/declarativewebcontainer.h
```diff
@@ -182,12 +188,18 @@ private slots:
...
 
     QPointer<DeclarativeWebPage> m_webPage;
     QPointer<DeclarativeTabModel> m_model;
     QPointer<QQmlComponent> m_webPageComponent;
     QPointer<SettingManager> m_settingManager;
-    QScopedPointer<WebPages> m_webPages;
+    QPointer<WebPages> m_webPages;
+
+    QScopedPointer<DeclarativeTabModel> m_persistentTabModel;
+    QScopedPointer<DeclarativeTabModel> m_privateTabModel;
```
rojkov: Setting parent for scoped members was a bad idea.
Please replace QScopedPointer with QPointer. Scoping is not needed if parent is set for an instance. Otherwise Qt will try to delete the member object twice when DeclarativeWebContainer instance is destroying => crash.

rainemak:
Let's merge this now. LGTM

@rojkov will do some updates after.

PR #288, opened 2015-02-10
[webcontainer] Actually disconnect signals from TabModel

rojkov:
The disconnect method with a single argument doesn't actually disconnect signals even though it compiles


rainemak:
LGTM

PR #289, opened 2015-02-11
[sailfish-browser] Add 'activateNewTabView' dbus service method

blammit:


src/dbusadaptor.h
```diff
@@ -25,6 +25,7 @@ class DBusAdaptor : public QDBusAbstractAdaptor
 
 public slots:
     void openUrl(QStringList args);
+    void openNewTab();
```
rainemak: I think this should be something like enterNewTabUrl() or activateNewTabView() as this doesn't yet open new tab instead just brings up the view where user can enter the new tab url or activate a favorite.

I'd like to see openNewTab(QString url) reversed for actually opening an external link to a new tab without activating the overlay.

rainemak:
Does not work if browser is not running when calling the dbus method.

src/dbusadaptor.h
blammit: Yep that sounds good, I've changed it to activateNewTabView().

blammit:
D-Bus activation works okay for me when the app is not running, if there's no check for whether the interface exists before the call is made, e.g.

dbus-send --print-reply --dest=org.sailfishos.browser / org.sailfishos.browser.activateNewTabView

rainemak:
LGTM

PR #296, opened 2015-02-17
[sailfish-browser] Use grid layout for tab view on large screen

rainemak:
On large screen
- 2 columns on portrait
- 3 columns on landscape

### DIFF
--- a/src/pages/components/TabView.qml
+++ b/src/pages/components/TabView.qml
@@ -17,6 +17,7 @@ SilicaGridView {
     id: tabView
 
     property bool portrait
+    readonly property bool largeScreen: Screen.sizeCategory > Screen.Medium
 
     signal hide
     signal enterNewTabUrl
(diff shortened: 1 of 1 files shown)


siteshwar:
LGTM

rojkov:
LGTM

src/pages/components/TabView.qml
```diff
@@ -39,12 +42,15 @@ SilicaGridView {
     delegate: TabItem {
         id: tabItem
 
+        readonly property bool largeLeftMargin: index % columns === 0
+        readonly property bool largeRightMargin: columns > 1 ? index % columns === (columns - 1) : true
```
kananoja: "index % columns === (columns - 1)" should be enough. No need for the tertiary operation afaics.

kananoja: I would rename the properties to something like this:
- "largeLeftMargin" -> "firstColumn"
- "largeRightMargin" -> "lastColumn"
Maybe it's just me, but I find it easier to understand the margin bindings (below) if they were named like this.

rainemak: Thanks @kananoja will do suggested changes. firstColumn and lastColumn are better names.

rainemak:
Updated as per comments.

kananoja:
LGTM

PR #298, opened 2015-02-20
[sailfish-browser] Adjust favorite grid for large screen

rainemak:


src/pages/components/FavoriteGrid.qml
```diff
@@ -51,17 +75,13 @@ SilicaGridView {
...
-    // for the context menu when it's open.
-    property int minOffsetIndex: contextMenu ? currentIndex - (currentIndex % columns) + columns : 0
-    property int yOffset: contextMenu ? contextMenu.height : 0
-
     delegate: ListItem {
         id: container
 
         property Item remorse
+        property real offsetY: largeScreen
+                               ? (((y % (favoriteGrid.pageHeight-container.contentHeight+1)) / (favoriteGrid.pageHeight-container.contentHeight+1)) - 0.5) * (Theme.paddingLarge*4)
```
kananoja: This will cause the actual grid item (ListItem) and the visible item (FavoriteItem) y coordinate to diverge, and eventually  the visible item will "disappear" before it reaches the bottom of the screen.

kananoja: Don't like the patch too much, but can live with it for a while.

rainemak:
I feel you @kananoja .

PR #301, opened 2015-02-24
[sailfish-browser] Remove redundant tabsCleared signal JB#26644

rojkov:
The slot DeclarativeWebContainer::onTabsCleared() is redundant as
for every deleted webview DeclarativeWebContainer::releasePage() is
called and in this method we can do the same resets upon emptying
the current tabs model (DeclarativeWebContainer::m_model).

Since DeclarativeWebContainer::onTabsCleared() is the only client
for DeclarativeTabModel::tabsCleared() signal the signal has been
removed as well.

Also DBManager::removeAllTabs() has been removed as unused.

### DIFF
--- a/src/dbworker.cpp
+++ b/src/dbworker.cpp
@@ -571,7 +571,6 @@ HistoryResult DBWorker::addToBrowserHistory(QString url, QString title)
 
 void DBWorker::clearHistory()
 {
-    int oldTabCount = tabCount();
     QSqlQuery query = prepare("DELETE FROM browser_history;");
     execute(query);
     removeAllTabs();
@@ -580,10 +579,6 @@ void DBWorker::clearHistory()
 
     QList<Link> linkList;
     emit historyAvailable(linkList);
-    if (oldTabCount != 0) {
-        QList<Tab> tabList;
-        emit tabsAvailable(tabList);
-    }
 }
 
 int DBWorker::addToTabHistory(int tabId, int linkId)
--- a/src/declarativewebcontainer.cpp
+++ b/src/declarativewebcontainer.cpp
@@ -639,11 +639,6 @@ void DeclarativeWebContainer::releasePage(int tabId, bool virtualize)
         m_webPages->release(tabId, virtualize);
         // Successfully destroyed. Emit relevant property changes.
         if (!m_webPage || m_model->count() == 0) {
-            if (m_tabId != 0) {
-                m_tabId = 0;
-                emit tabIdChanged();
-            }
-
             if (m_canGoBack) {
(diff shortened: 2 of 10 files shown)


siteshwar:
LGTM

src/dbworker.cpp
```diff
@@ -580,10 +579,6 @@ void DBWorker::clearHistory()
 
     QList<Link> linkList;
     emit historyAvailable(linkList);
-    if (oldTabCount != 0) {
-        QList<Tab> tabList;
-        emit tabsAvailable(tabList);
-    }
```
rainemak: Running tabs should be cleared when this happens. Currently only meaningful call that ends up here is in SettingsManager. The call is triggered when you clear history from the settings.

I'd leave this over here so that tabmodel's onTabsAvailable slot can clear tabs.

src/declarativewebcontainer.cpp
```diff
@@ -655,8 +638,12 @@ void DeclarativeWebContainer::releasePage(int tabId, bool virtualize)
     if (m_webPages) {
         m_webPages->release(tabId, virtualize);
         // Successfully destroyed. Emit relevant property changes.
-        if (!m_webPage) {
-            m_tabId = 0;
+        if (!m_webPage || m_model->count() == 0) {
+            if (m_tabId != 0) {
+                m_tabId = 0;
+                emit tabIdChanged();
```
rainemak: Previously sequence emitted this signal as a last signal. Updating data over here is fine.

So, when tabId has changed all data always matches to the tabId, no wrong title/url etc.

src/persistenttabmodel.cpp
```diff
@@ -36,10 +36,12 @@ void PersistentTabModel::tabsAvailable(QList<Tab> tabs)
 {
     beginResetModel();
     int oldCount = count();
-    m_tabs.clear();
-    m_tabs = tabs;
```
rainemak: Actually... would following do the same trick. I think there was and is possible error, at least code is error phone. 

```
// Clear always previous tabs
clear();
if (m_tabs.count() > 0) {
    m_tabs = tabs;
...
```
```

src/dbworker.cpp
rojkov: These lines are redundant as tabsAvailable() gets emitted by this moment inside removeAllTabs() which is called in the line 576

rainemak: True.

rojkov:
Updated. Now previous tabs are always cleared and tabIs is changed after resetting title and url.

rainemak:
LGTM

PR #302, opened 2015-02-27
[sailfish-browser] Fix rotation when overlays are opened over the browse...

adenexter:
...r.

Overwrite the window page orientation so the orientation reported to the
compositor isn't forced to portrait when the window isn't active.

@rainemak


rainemak:
LGTM

PR #304, opened 2015-03-02
[sailfish-browser] While activating a tab do not use page's title and url

rainemak:
While we are in process of activating a tab, do not pass title and url
of the active page through as it is already out dated. Previously this
caused an issue when cached url / title (m_url / m_title) got updated
and matching change signals were emitted right after new tab data became
available. When QML side tried to read updated value we returned
the value of the previous page.

As all title and url updates are done through updateTitle and updateUrl, when setWebPage
got called m_title and m_url were already up-to-date. Thus, change signals
didn't get emitted again.

### DIFF
--- a/src/declarativewebcontainer.cpp
+++ b/src/declarativewebcontainer.cpp
@@ -46,6 +46,7 @@ DeclarativeWebContainer::DeclarativeWebContainer(QQuickItem *parent)
     , m_popupActive(false)
     , m_portrait(true)
     , m_fullScreenMode(false)
+    , m_activatingTab(false)
     , m_fullScreenHeight(0.0)
     , m_inputPanelVisible(false)
     , m_inputPanelHeight(0.0)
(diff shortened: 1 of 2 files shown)


rojkov:
Looks good, but wouldn't it be better to use less stateful approach, i.e. not to use potentially outdated data with

```
diff --git a/src/pages/components/Overlay.qml b/src/pages/components/Overlay.qml
index 6b38433..de148e8 100644
--- a/src/pages/components/Overlay.qml
+++ b/src/pages/components/Overlay.qml
@@ -220,7 +220,7 @@ PanelBackground {
 
                 url: webView.contentItem && webView.contentItem.url || ""
                 findText: searchField.text
-                bookmarked: bookmarkModel.count && bookmarkModel.contains(webView.url)
+                bookmarked: webView.webPage && bookmarkModel.count && bookmarkModel.contains(webView.webPage.url)
                 opacity: (overlay.y - webView.fullscreenHeight/2)  / (webView.fullscreenHeight/2 - toolBar.height)
                 visible: opacity > 0.0
                 secondaryToolsActive: overlayAnimator.secondaryTools
```

rainemak:
Yes, considered above by using the "url" of the toolbar (effectively the same). But as the root of the error is in WebContainer and that should be fixed as well.

rojkov:
LGTM

PR #307, opened 2015-03-11
[sailfish-browser] Limit history size to 2000 entries

### LINKED ISSUE #199: Remove oldest browsing history entries from database
Oldest entries should be periodically cleaned from browser's history database. Either keep only X amount of latest entries or timestamp entries and remove entries older than X days.

2-3months time should be enough, while it can still be a lot.

As history database queries are executed in another thread, clean-up could be for instance done when browser is started.

siteshwar:
Fixes #199

### DIFF
--- a/src/dbworker.cpp
+++ b/src/dbworker.cpp
@@ -31,6 +31,8 @@
 #define QUOTE(arg) #arg
 #define STR(arg) QUOTE(arg)
 
+#define MAX_BROWSER_HISTORY_SIZE 2000
+
 static const char * const create_table_tab =
         "CREATE TABLE tab (tab_id INTEGER PRIMARY KEY,\n"
         "tab_history_id INTEGER\n"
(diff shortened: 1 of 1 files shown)


src/dbworker.cpp
```diff
@@ -107,6 +107,13 @@ void DBWorker::init()
             QSqlQuery query = prepare(db_schema[i]);
             execute(query);
         }
+    } else {
+        // Limit history size to 2000 entries
+        QSqlQuery cleanupHistory = prepare("DELETE FROM browser_history WHERE id NOT IN (SELECT id from "\
+                                           "browser_history ORDER BY date DESC LIMIT 2000);");
```
rojkov: I'd better define the number in a macros closer to the top of the file

siteshwar: Fixed through 2b83f0f

src/dbworker.cpp
```diff
@@ -107,6 +109,13 @@ void DBWorker::init()
             QSqlQuery query = prepare(db_schema[i]);
             execute(query);
         }
+    } else {
+        // Limit history size to 2000 entries
+        QSqlQuery cleanupHistory = prepare(QString("DELETE FROM browser_history WHERE id NOT IN (SELECT id from "\
+                                                   "browser_history ORDER BY date DESC LIMIT %1);").arg(MAX_BROWSER_HISTORY_SIZE));
```
rojkov: micro-optimization: a better way to extrapolate a string with a constant is to use the STR macro. This way the string would be extrapolated in build time, not run time.

siteshwar: I thought about it but decided not to do it as it might not look very readable.

rojkov: That's a valid point.

But I feel it's acceptable to go for a slight optimization as this SQL is simple enough to grasp even with some noise added.

rojkov:
LGTM

PR #308, opened 2015-03-11
Upload tab's session history from local DB to Gecko

rojkov:
Warning: this is a preliminary implementation of uploading session history to the engine. Do not merge.  There are broken use cases related to downloading and some of them are still not known.

Depends on
https://github.com/rojkov/embedlite-components/commit/aa2e5eca8a4ae3ba7a66367e804de9ec3a450add
https://github.com/rojkov/qtmozembed/commit/6b61c57605922fba3185916e362cf957eb63590e


src/declarativewebcontainer.cpp
```diff
@@ -266,31 +261,11 @@ void DeclarativeWebContainer::setInputPanelHeight(qreal height)
     }
 }
 
-bool DeclarativeWebContainer::canGoForward() const
```
rainemak: I'd keep this two but read directly from the current page. No need to do caching over here.

PR #309, opened 2015-03-18
Bookmarks update

rojkov:
Also this PR makes the bookmarks ordered.

### DIFF
--- a/content/bookmarks.json
+++ b/content/bookmarks.json
@@ -1,19 +1,13 @@
 [
     {
         "favicon": "image://theme/icon-m-service-jolla",
-        "title": "User Guide",
-        "url": "http://jolla.com/guide/"
+        "title": "Jolla",
+        "url": "http://jolla.com/"
     },
     {
         "favicon": "image://theme/icon-m-service-jolla",
-        "title": "How-to tutorial videos",
-        "url": "http://m.youtube.com/playlist?list=PLQgR2jhO_J0y8YSSvVd-Mg9LM88W0aIpD&desktop_uri=%2Fplaylist%3Flist%3DPLQgR2jh
-O_J0y8YSSvVd-Mg9LM88W0aIpD"
-    },
-    {
-        "favicon": "image://theme/icon-m-service-jolla",
-        "title": "Jolla Questions & Answers",
-        "url": "https://together.jolla.com/questions/"
+        "title": "How-to videos",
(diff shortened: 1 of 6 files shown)


src/declarativebookmarkmodel.cpp
```diff
@@ -47,12 +51,12 @@ void DeclarativeBookmarkModel::removeBookmark(const QString& url)
         return;
     }
 
-    if (bookmarks.contains(url)) {
-        int index = bookmarkUrls.indexOf(url);
+    if (bookmarkIndexes.keys().contains(url)) {
```
rainemak: What if we'd drop also bookmarkIndexes QMap. And remove the first occurrence by first checking indexOf(url) if greater than zero then remove it.

We're always removing first occurrence.

rainemak: What if we'd drop also bookmarkIndexes QMap. And remove the first occurrence by first checking bookmarkIndex(url) if greater than zero then remove it. We'd need a simple function bookmarkIndex function to find the occurrence.

We're always removing first occurrence.

rainemak: And above bookmarkIndex function could be used in DeclarativeBookmarkModel::contains(const QString& url). However, we'd get a performance hit when comparing to map.

rojkov: Basically the performace is the only reason why I decided to stick to QMap: O(N) vs O(log2 N).
Do we want to sacrifice that?

content/bookmarks.json
```diff
@@ -1,28 +1,32 @@
...
+        "title": "Jolla",
+        "url": "http://jolla.com/"
     },
     {
         "favicon": "image://theme/icon-m-service-jolla",
-        "title": "How-to tutorial videos",
-        "url": "http://m.youtube.com/playlist?list=PLQgR2jhO_J0y8YSSvVd-Mg9LM88W0aIpD&desktop_uri=%2Fplaylist%3Flist%3DPLQgR2jh
-O_J0y8YSSvVd-Mg9LM88W0aIpD"
+        "title": "How-to videos",
+        "url": "http://m.youtube.com/playlist?list=PLQgR2jhO_J0y8YSSvVd-Mg9LM88W0aIpD&desktop_uri=%2Fplaylist%3Flist%3DPLQgR2jhO_J0y8YSSvVd-Mg9LM88W0aIpD"
```
rainemak: Should be:
https://www.youtube.com/playlist?list=PLQgR2jhO_J0zrl6kLWusekAFr9aHR9jcH

rainemak:
Add some small comments. Other than that LGTM.

Let's keep bookmark indexes as a separate entity.

rojkov:
outdated URL is a good catch! I somehow missed it.
Added comments.

PR #315, opened 2015-03-31
[sailfish-browser] Don't instantiate an abstract class. JB#27654

rojkov:
For some reason QML thinks that a property declaration is the same as class instantiation. And since TabModel is an abstract class QML fails to load BrowserContextMenu.qml.


siteshwar:
LGTM

rainemak:
LGTM

PR #322, opened 2015-05-13
[sailfish-browser] Fix favorite grid overlay edges for bigger screens.

rainemak:
Closing the favorite grid overlay from the space between right edge of screen and right-most icon is not allowing dragging gesture, similar at the left edge.


rojkov:
LGTM

PR #323, opened 2015-05-25
[sailfish-browser] Removed remorse timer from delete bookmark. Contri…

antseppa:
…butes to JB#27571

### DIFF
--- a/src/pages/components/FavoriteContextMenu.qml
+++ b/src/pages/components/FavoriteContextMenu.qml
@@ -50,7 +50,7 @@ Component {
             //: "Remove favorited / bookmarked web page"
             //% "Remove favorite"
             text: qsTrId("sailfish_browser-me-remove_favorite")
-            onClicked: delegate.remove()
+            onClicked: bookmarkModel.removeBookmark(url)
         }
     }
 }
--- a/src/pages/components/FavoriteGrid.qml
+++ b/src/pages/components/FavoriteGrid.qml
@@ -61,18 +61,6 @@ SilicaGridView {
     delegate: ListItem {
         id: container
 
-        property Item remorse
-
-        function remove() {
-            //: Remorse timer for removing bookmark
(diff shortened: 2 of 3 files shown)


src/pages/components/FavoriteGrid.qml
```diff
@@ -61,16 +61,8 @@ SilicaGridView {
     delegate: ListItem {
         id: container
 
-        property Item remorse
-
         function remove() {
```
rainemak: I'd also remove this function and move ```bookmarkModel.removeBookmark(url)``` to the FavoriteContextMenu. Matter of taste as currently FavoriteContextMenu is not using ids from elsewhere and that's also good. Not big deal.

For safety, add url parameter for the remove function is you don't remove this. Now url refers to the model.url but it looks quite odd (maybe I have added at some point of time) as call is coming through FavoriteContextMenu. Url is passed to the context menu in the show.

antseppa:
Updated: removed remove() function and moved bookmarkModel.removeBookmark(url) to context menu.

antseppa:
Updated: added a commit which updates indices in bookmarkIndexes map after a bookmark has been removed. Without this, consecutive calls to DeclarativeBookmarkModel::removeBookmark either crashed browser or caused wrong bookmark to be deleted if the index of the bookmark was greater than previous one.

src/pages/components/FavoriteContextMenu.qml
```diff
@@ -50,7 +50,9 @@ Component {
             //: "Remove favorited / bookmarked web page"
             //% "Remove favorite"
             text: qsTrId("sailfish_browser-me-remove_favorite")
-            onClicked: delegate.remove()
+            onClicked: {
+                bookmarkModel.removeBookmark(url)
+            }
```
rainemak: No need for brackets.

rainemak:
LGTM

PR #326, opened 2015-06-01
Enable favorite item's context menu on Tablet.

rojkov:


tworaz:
LGTM

src/pages/components/FavoriteContextMenu.qml
```diff
@@ -20,6 +20,9 @@ Component {
         property string title
         property string url
 
+        // FavoriteGrid doesn't fill the entire page thus set width to FavoriteGrid's parent
+        width: _flickable !== null ? _flickable.parent.width : (parent ? parent.width : 0)
+
```
rainemak: Menu position is not updated correctly when rotating the device so that context menu is open.

rojkov: Updated PR:
If the binding for x is not broken (see below) then the position is correct upon orientation change.

rainemak:
LGTM. Looking good!

PR #334, opened 2015-06-04
[sailfish-browser] Avoid dynamic updates of TabItem background. 

rojkov:
I still don't understand how exactly, but this patch seems to fix the crashes when creating a tab in private mode and switch back and forth to normal mode.


rainemak:
LGTM

PR #342, opened 2015-06-24
[sailfish-browser] Transfers icon updated. Fixes JB#30041

kananoja:


rainemak:
LGTM. Could you bump up browser version to 1.3.12 from the spec before tagging.

PR #344, opened 2015-06-25
[sailfish-browser] Open content dimmer when private browsing mode changed. Fixes JB#30226

rainemak:
Fix revealed following problems that needed to fixed in context of
this PR.

1) When browsing mode is changed, wait for a tab. Waiting is cleared from the
old model.

2) Share some logic for both model change and model count change for showing
overlay and updating the cover.

3) When switching from normal browsing to private browsing (not tabs) and back
to normal browsing enteringNewTabUrl was left to true. This made browser
page activation failing to show chrome. When the overlay is dismissed,
set the enteringNewTabUrl to false.

4) When changing from normal browsing to private browsing, the overlay was
dismissed so that chrome was revealed. This broken chrome showing animation.
So, when switching between browsing modes and overlay is dismissed it is
dismissed so that chrome is hidden. Thus, allowing browser page
activation to animate the chrome back to visible.

### DIFF
--- a/src/declarativewebcontainer.cpp
+++ b/src/declarativewebcontainer.cpp
@@ -191,6 +191,7 @@ void DeclarativeWebContainer::setTabModel(DeclarativeTabModel *model)
         if (m_model) {
             disconnect(m_model, 0, 0, 0);
             oldCount = m_model->count();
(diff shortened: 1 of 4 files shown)


rojkov:
LGTM

PR #347, opened 2015-06-26
No need to make background opaque when entering new tab URL. JB#30228

rojkov:


rainemak:
When opening the overlay so that we're entering a url, background should be dimmed. When the new tab url is given, dimmer is opened after the first frame swap.

PR #348, opened 2015-06-26
[sailfish-browser] Don't bring browser to foreground if page status changes. Fixes JB#30260

rainemak:
Fixes also activating browser from new tab cover action or from the cover area.

Chrome focus handling related manual test case steps added.


rojkov:
LGTM

PR #349, opened 2015-06-26
[sailfish-browser] Activate current tab if needed upon dismissing 'Enter URL' form

rojkov:


rainemak:
LGTM

rainemak:
Good catch. Thanks!

PR #352, opened 2015-06-29
[sailfish-browser] Update private mode texture. Fixes JB#30228

rainemak:
Opacity of the texture is still 10% but diagonal stripes are
black and white so that it's visible on top of light and dark
background.

Thanks @roppola !


rojkov:
Looks cool to me!

PR #353, opened 2015-06-30
[sailfish-browser] Communicate to the compositor that the window has nested children. Contributes to JB#30188

adenexter:
…nested children. Contributes to JB#30188


rainemak:
If there is no better way to communicate nested children than wayland property => LGTM

PR #358, opened 2015-07-01
Minor cleanups

rojkov:
The first commit removes unused previousLink and nextLink properties from the `Tab` struct.

The second one simply renames DeclarativeWebPage::m_tab to DeclarativeWebPage::m_initialTab in order to stress the fact that this tab data is static and never changes after WebPage's initialization.


rainemak:
Nice set of cleanups!

```void DBManager::getTab(int tabId);``` caught my eye as well. Removing that renders also ```DBWorker::getTab``` to void.

rojkov:
Removed `getTab()` and related signals.

rainemak:
If tests are green. LGTM.

PR #359, opened 2015-07-02
[sailfish-browser] Focus search field after browser page gets activated. Fixes JB#30418

rainemak:
Silica Page is a focus scope so we should not focus the search field
until the browser page is active.

Before this fix virtual keyboard opened already on the tab page when
all tabs were closed and you started pop gesture from the tab page. Due
to this and virtual keyboard observer there was a white stripe blinking
at the bottom of the screen.


rojkov:
LGTM

PR #361, opened 2015-07-02
[sailfish-browser] Add new tab cover action back to the tab view. Fixes JB#30443

rainemak:
This workaround adds the new tab cover action back to the cover in case
tab page is active when browser pushed to switcher. Other sub pages are not
handled (sharing and web prompts). For web prompts we should not have
new tab cover action as triggering the action would reject the prompt.

Remove this when the tab view cover is done.

### DIFF
--- a/src/pages/components/Overlay.qml
+++ b/src/pages/components/Overlay.qml
@@ -412,6 +412,8 @@ Background {
             id: tabPage
             property int activeTabIndex
 
+            onStatusChanged: browserPage.tabPageActive = (status == PageStatus.Active)
+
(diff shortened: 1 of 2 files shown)


src/pages/components/Overlay.qml
```diff
@@ -412,6 +412,8 @@ Background {
             id: tabPage
             property int activeTabIndex
 
+            onStatusChanged: browserPage.tabPageActive = (status == PageStatus.Active)
```
rojkov: LGTM

PR #363, opened 2015-07-03
[sailfish-browser] Improve bookmark icon parsing. Contributes to JB#30209

rainemak:
Increase upper limit of parsed icon, allow touch icon picking
without size, and allow high quality (large) parsed icons.

### DIFF
--- a/src/pages/components/WebView.qml
+++ b/src/pages/components/WebView.qml
@@ -189,21 +189,20 @@ WebContainer {
                             if (size) {
                                 sizes.push(size)
                             }
-                        } else {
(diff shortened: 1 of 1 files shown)


src/pages/components/WebView.qml
```diff
@@ -189,21 +189,20 @@ WebContainer {
                             if (size) {
                                 sizes.push(size)
                             }
-                        } else {
-                            sizes = data.href.match(/\d+/)
```
rainemak: Don't read anything from the url. This was big no no as the number might have been from sha1 (cdn url).

rojkov:
LGTM

PR #365, opened 2015-07-03
Save active tab in settings upon every tab switch

rojkov:


tworaz:
LGTM

src/persistenttabmodel.cpp
```diff
@@ -70,6 +68,8 @@ void PersistentTabModel::tabsAvailable(QList<Tab> tabs)
         m_loaded = true;
         emit loadedChanged();
     }
+
+    connect(this, SIGNAL(activeTabIndexChanged()), this, SLOT(saveActiveTab()));
```
rainemak: This should be unique connection as tabsAvailable will be emitted also when all tabs are closed.

rainemak:
Commit sha1 a829460 has a broken change log entry and the activeTabIndexChanged connection should be unique. Please fix those before merging. Other than that LGTM

rojkov:
Updated the activeTabIndexChanged connection to be unique and shortened the commit message.

rainemak:
LGTM

PR #366, opened 2015-07-06
[sailfish-browser] Disable decoder limit workaround. Contributes to JB#30517

tworaz:
Since the workaround breaks loading of pages like jolla.com/tablet and
the original bug it was supposed to fix (JB#11038) is no longer
reproducible let's disable it.


rojkov:
LGTM

rainemak:
LGTM

PR #369, opened 2015-07-06
[sailfish-browser] Insert new tab next to the parent tab. Fixes JB#30333

rainemak:
When a tab is closed and it had a parent tab, the parent tab activated
otherwise the index next to the previous active tab is activated.

### DIFF
--- a/src/declarativetabmodel.cpp
+++ b/src/declarativetabmodel.cpp
@@ -13,6 +13,7 @@
 #include "declarativetabmodel.h"
 #include "linkvalidator.h"
 #include "declarativewebutils.h"
+#include "declarativewebcontainer.h"
 #include "declarativewebpage.h"
 
 #include <QFile>
@@ -24,12 +25,13 @@
 #define DEBUG_LOGS 0
 #endif
 
-DeclarativeTabModel::DeclarativeTabModel(int nextTabId, QObject *parent)
-    : QAbstractListModel(parent)
+DeclarativeTabModel::DeclarativeTabModel(int nextTabId, DeclarativeWebContainer *webContainer)
+    : QAbstractListModel(webContainer)
     , m_activeTabId(0)
     , m_loaded(false)
     , m_waitingForNewTab(false)
     , m_nextTabId(nextTabId)
+    , m_webContainer(webContainer)
 {
 }
 
@@ -48,10 +50,13 @@ QHash<int, QByteArray> DeclarativeTabModel::roleNames() const
     return roles;
 }
 
-void DeclarativeTabModel::addTab(const QString& url, const QString &title) {
+void DeclarativeTabModel::addTab(const QString& url, const QString &title, int index) {
     if (!LinkValidator::navigable(QUrl(url))) {
         return;
     }
+
+    Q_ASSERT(index >= 0 && index <= m_tabs.count());
+
     int tabId = createTab();
     int linkId = createLink(tabId, url, title);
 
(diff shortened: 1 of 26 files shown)


src/declarativewebcontainer.cpp
```diff
@@ -411,6 +411,14 @@ bool DeclarativeWebContainer::activatePage(const Tab& tab, bool force, int paren
     return false;
 }
 
+int DeclarativeWebContainer::findParentTabId(int tabId)
```
rojkov: This method can be marked as `const`.

src/declarativewebcontainer.h
```diff
@@ -114,6 +114,7 @@ class DeclarativeWebContainer : public QWindow, public QQmlParserStatus, protect
 
     bool isActiveTab(int tabId);
     bool activatePage(const Tab& tab, bool force = false, int parentId = 0);
+    int findParentTabId(int tabId);
```
rojkov: I guess this also deprecates 
```
int parentTabId(int tabId) const;
```

src/declarativetabmodel.cpp
```diff
@@ -378,6 +378,22 @@ void DeclarativeTabModel::updateActiveTab(const Tab &activeTab, bool loadActiveT
     }
 }
 
+void DeclarativeTabModel::setWebContainer(DeclarativeWebContainer *webContainer)
```
rojkov: Just a note: I think there is no need to register tabModels as QML types nowadays thus this method can be dropped with QML type registration altogether.

rainemak: Those must be still registered but not as creatable types. Will add a commit on top. Above method is only used from the unit tests at the moment.

rainemak:
Fixed and amended review feedback.

In addition, added a commit on top that removes means to change tab models from qml.

src/declarativetabmodel.cpp
```diff
@@ -161,13 +165,9 @@ void DeclarativeTabModel::closeActiveTab()
 {
     if (!m_tabs.isEmpty()) {
         int index = activeTabIndex();
+        int newActiveIndex = nextActiveTabIndex(index);
         removeTab(m_activeTab.tabId(), m_activeTab.thumbnailPath(), index);
-
```
rainemak: Removing last active tab fails. Will fix.

src/declarativetabmodel.cpp
rojkov: Should this "uncreatableness" be reflected in unit tests as well?

rainemak: In tst_webview yes, others are not having webcontainer. Will update.

src/declarativetabmodel.cpp
```diff
@@ -122,31 +128,33 @@ bool DeclarativeTabModel::activateTab(const QString& url)
...
     }
     return false;
 }
 
-bool DeclarativeTabModel::activateTab(int index, bool loadActiveTab)
+void DeclarativeTabModel::activateTab(int index, bool loadActiveTab)
 {
-    if (index >= 0 && index < m_tabs.count()) {
-        const Tab &newActiveTab = m_tabs.at(index);
+    if (m_tabs.isEmpty()) {
```
rainemak: Don't do anything if all tabs have been closed.

src/declarativetabmodel.cpp
```diff
@@ -48,10 +50,14 @@ QHash<int, QByteArray> DeclarativeTabModel::roleNames() const
...
 }
 
-void DeclarativeTabModel::addTab(const QString& url, const QString &title) {
+void DeclarativeTabModel::addTab(const QString& url, const QString &title, int index) {
     if (!LinkValidator::navigable(QUrl(url))) {
         return;
     }
+
+    Q_ASSERT(index >= 0);
+    Q_ASSERT(index <= m_tabs.count());
```
rainemak: This could be changed to be one-liner

rojkov:
LGTM

PR #370, opened 2015-07-07
[sailfish-browser] Polish orientation fading animation. Contributes to JB#30405

rainemak:
Remove unnecessary parallel animation and change fade in
animation to be a sequental animation so that total duration increases
by the pause animations.


rojkov:
LGTM

PR #371, opened 2015-07-07
Use persistent connection to update tab titles. JB#30433

rojkov:


src/declarativetabmodel.h
```diff
@@ -107,7 +106,7 @@ public slots:
     virtual void updateThumbPath(int tabId, QString path) = 0;
 
     // This should be replaced by m_activeTabIndex
```
rainemak: Remove the above comment :)

```m_activeTabIndex``` is as good as ```m_activeTabId``` both would occasionally trigger some searching by index or by tabId.

File: src/declarativetabmodel.cpp [hunk 2]
```diff
@@ -249,7 +250,8 @@ const QList<Tab> &DeclarativeTabModel::tabs() const
 
 const Tab &DeclarativeTabModel::activeTab() const
 {
-    return m_activeTab;
+    Q_ASSERT(contains(m_activeTabId));
```
rainemak: What if all tabs are closed?

tests/auto/tst_declarativetabmodel/tst_declarativetabmodel.cpp
```diff
@@ -283,9 +281,6 @@ void tst_declarativetabmodel::multipleTabsWithSameUrls()
     QCOMPARE(tabModel->data(modelIndex, DeclarativeTabModel::UrlRole).toString(), page2Tab1Url);
     QVERIFY(tabModel->data(modelIndex, DeclarativeTabModel::TitleRole).toString().isEmpty());
 
-    tabModel->updateTitle(tab2, true, page2Tab2Url, page2Tab2Title);
-    QCOMPARE(tabModel->activeTab().title(), page2Tab2Title);
-
```
rainemak: Would be good to test that both url and title are updated correctly. One approach that we could take is that we utilize the mock webpage, add apis to the mock for setting url and title, and connect the mock the model. This way test case could control input and verify that model updates correctly.

src/declarativetabmodel.cpp
rojkov: Good point!
The assertion gets triggered when switching to the private mode at the function updating `DeclarativeWebContainer::m_tabId`. Now I wonder why we need to maintain `m_tabId` in `DeclarativeWebContainer`...

tests/auto/tst_declarativetabmodel/tst_declarativetabmodel.cpp
rojkov: Yes. I'd like to rewrite tst_declarativetabmodel where each test case would run in the same environment set up in setUp() and tearDown() functions.

src/declarativetabmodel.cpp
rainemak: Probably with a connection to the active model we could remove ```m_tabId``` from DeclarativeWebContainer as well and just leave tabId() method to return activate tab id from the current active model. It is used in ```DeclarativeWebContainer::reload``` and in WebView.qml within the WebPage {} to check if this is the active tab.

Maybe at some point we could add an attached property like ```isCurrentItem``` to the webpage.

tests/auto/tst_declarativetabmodel/tst_declarativetabmodel.cpp
rainemak: Sounds like a plan.

rojkov:
I've dropped `DeclarativeWebContainer::m_tabId` and simplified the signature of `activeTabChanged()` signal. The assertion in `DeclarativeTabModel::activeTab()` should be safe now.

tests/auto/tst_declarativetabmodel/tst_declarativetabmodel.cpp
```diff
@@ -58,8 +58,6 @@ private slots:
     void invalidTabs_data();
```
rainemak: At least activateTabs and closeActiveTab are failing due to signal prototype change.

QSignalSpy::QSignalSpy:88 - QSignalSpy: No such signal: 'activeTabChanged(int,int)'

rainemak:
All in all, looking good. Minor fix needed for tst_declarativetabmodel and one commit message had a typo.

tests/auto/tst_declarativetabmodel/tst_declarativetabmodel.cpp
rojkov: Updated the signal's signature.

rojkov:
fixed the commit message

rainemak:
LGTM

PR #376, opened 2015-07-09
Reimplement unit tests for PersistentTabModel with isolated cases.

rojkov:
Depends on https://github.com/tmeshkova/qtmozembed/pull/142

### DIFF
--- a/tests/auto/tst_persistenttabmodel/tst_persistenttabmodel.cpp
+++ b/tests/auto/tst_persistenttabmodel/tst_persistenttabmodel.cpp
@@ -1,0 +1,424 @@
+/****************************************************************************
+**
+** Copyright (C) 2015 Jolla Ltd.
+** Contact: Dmitry Rozhkov <dmitry.rozhkov@jolla.com>
+**
+****************************************************************************/
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include <QtTest/QtTest>
+#include <QQmlEngine>
+#include <QQmlComponent>
+#include <QQuickView>
+
+#include "persistenttabmodel.h"
+#include "dbmanager.h"
+#include "declarativewebpage.h"
+
+static const QByteArray QML_SNIPPET = \
+        "import QtQuick 2.0\n" \
+        "import Sailfish.Browser 1.0\n" \
+        "Item {\n" \
+        "   width: 100; height: 100\n" \
+        "   property alias tabModel: model\n" \
+        "   PersistentTabModel { id: model }\n" \
+        "}\n";
+
+class tst_persistenttabmodel : public QObject
+{
+    Q_OBJECT
+
+private slots:
+    void initTestCase();
+    void cleanupTestCase();
+    void init();
+    void cleanup();
+
+    void addTab();
+    void remove();
+    void removeTabById_data();
+    void removeTabById();
+    void clear();
+    void activateTab_by_url_data();
+    void activateTab_by_url();
+    void activateTabById_data();
+    void activateTabById();
+    void closeActiveTab();
+    void updateUrl_data();
+    void updateUrl();
+    void onTitleChanged();
+
+private:
+    void addThreeTabs();
+
+    PersistentTabModel* tabModel;
+    QQuickView mView;
+    QPointer<QQmlComponent> mComponent;
+    QObject* mRootObject;
+    QString mDbFile;
+};
+
+void tst_persistenttabmodel::initTestCase()
+{
+    qGuiApp->setAttribute(Qt::AA_Use96Dpi, true);
+    qmlRegisterUncreatableType<DeclarativeTabModel>("Sailfish.Browser", 1, 0, "TabModel",
+                                                    "TabModel is abstract!");
+    qmlRegisterType<PersistentTabModel>("Sailfish.Browser", 1, 0, "PersistentTabModel");
+    mComponent = new QQmlComponent(mView.engine());
+    mComponent->setData(QML_SNIPPET, QUrl());
+    mDbFile = QString("%1/%2")
+            .arg(QStandardPaths::writableLocation(QStandardPaths::DataLocation))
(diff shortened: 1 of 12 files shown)


tests/auto/tst_persistenttabmodel/tst_persistenttabmodel.cpp
```diff
@@ -0,0 +1,424 @@
...
+    QSignalSpy countChangeSpy(tabModel, SIGNAL(countChanged()));
+    QSignalSpy tabAddedSpy(tabModel, SIGNAL(tabAdded(int)));
+    QSignalSpy nextTabIdChangedSpy(tabModel, SIGNAL(nextTabIdChanged()));
+    QSignalSpy dataChangedSpy(tabModel, SIGNAL(dataChanged(QModelIndex, QModelIndex, QVector<int>)));
+    QSignalSpy activeTabIndexChangedSpy(tabModel, SIGNAL(activeTabIndexChanged()));
+    QSignalSpy activeTabChangedSpy(tabModel, SIGNAL(activeTabChanged(int,bool)));
+
+    QList<QString> urls, titles;
+    urls << "http://example.com" << "file:///opt/tests/testpage.html" << "https://example.com";
+    titles << "Test title1" << "Test title2" << "Test title3";
```
rainemak: I'd prefer using ```_data``` function instead of using fixed array.

See: http://doc.qt.io/qt-5/qttestlib-tutorial2-example.html

rojkov: This function is not idempotent. That's why it's important to call it multiple times in a single test case run. Hence the dataset is required to be defined inside the case.

rainemak:
Seeing plenty of ```QSqlDatabasePrivate::addDatabase:261 - QSqlDatabasePrivate::addDatabase: duplicate connection ``` warnings.

TitleChanged case is failing

I don't see anything that would require: tmeshkova/qtmozembed#142

Approach in general looks good. I'd integrate init / cleanup enablers to the TestObject. The ```mComponent``` is created only once, and instances multiple times => TestObject could have a ```create``` that would create the ```mRootObject``` and maybe ```reset``` or ```clear``` or similar. Virtual ```init``` and ```cleanup``` should work as well but would likely require small changes to other test cases.

Then the ```void tst_persistenttabmodel::init()``` would look like:

```
void tst_persistenttabmodel::init()
{
    TestObject::create()
    tabModel =  TestObject::qmlObject<PersistentTabModel>("tabModel");
    QVERIFY(tabModel);

    if (!tabModel->loaded()) {
        QSignalSpy loadedSpy(tabModel, SIGNAL(loadedChanged()));
        // Tabs must be loaded with in 500ms
        QVERIFY(loadedSpy.wait());
        QCOMPARE(loadedSpy.count(), 1);
    }
}
```
And similarly ```persistenttabmodel::cleanup``` would first reset the TestObject.

tests/auto/tst_persistenttabmodel/tst_persistenttabmodel.cpp
rainemak: My point was that this could be changed so...

```
    QTest::addColumn<int>("insertToIndex");
    QTest::addColumn<QString>("newUrl");
    QTest::addColumn<QString>("newTitle");
    QTest::addColumn<QList<TestTab>("expectedTabs");
    // Each tab data and expected count would be covered by the above. Also model order could be checked.
```
TestTab could be moved from the testobject.h to own header and cpp and changed to be a QObject if needed.

In addition, it could be extended to include also tabId.

rojkov:
I'll try to clear the DB connection too in cleanup().

The failing TitleChanged case is exactly what requires tmeshkova/qtmozembed#142. Without the PR the code calls methods of real `DeclarativeWebPage` class, not the mock. By the way, I see this as a drawback of the approach that uses QML to create the objects under test. I'd prefer to create C++ objects in tests with just `new`. Because this is the way they are created in real code.

I see TestObject as an utility used to increase code reuse. Its content should become clear when other test suites are rewritten. Before that I don't see big value in coupling the tests with TestObject.

rainemak:
It just cannot call the real instance as you're creating a dummy mock only. For instance, try to clear part of the mock interface that is required by the tabmodel (e.g. title()) and see how the build starts failing. The mock is not inherited from the real webpage and both of the are not in the build of tst_persistenttabmodel. Dummy mock is just a QObject.

rainemak:
Aah, you're right and I'm wrong. But the solution is still wrong as this doesn't require libqt5embedwidget but only browser classes.

rojkov:
Agree, it shouldn't depend on libqt5embedwidget. The dep is introduced by the test_common.pri and `NO_COMMON_INCLUDE` defaulted to False. With NO_COMMON_INCLUDE=1 builds start to fail on linking.

tests/auto/tst_persistenttabmodel/tst_persistenttabmodel.cpp
rojkov: I guess you've missed the point the by the time the test case is called for every QTest::row `tabModel` doesn't contain entries at all thus the column `insertToIndex` doesn't make sense.

_data functions are used to execute a test multiple times with different test data. But the input table like
```
QTest::addColumn<QList<TestTab>("inputTabs");
QTest::addColumn<int>("insertToIndex");
```
would be useful indeed. I'll update the case.

rojkov:
I got rid of the warning, avoided inclusion of real headers when building objects with mocks and reworked ::addTab() unit test.

rainemak:
Works nicely now! LGTM.

We should rework a bit the source tree structure so that history.pri and bookmarks.pri could be included independently of the other sources so that include paths would not collide.

- src/history/history.pri (all headers and sources of the history.pri)

- src/bookmarks/bookmarks.pri (all headers and sources of the bookmarks.pri)

- src/core/core.pri (webpage, webcontainer, webpagecreator, webpages, webpagequeue, webutils, iconfetcher, downloadmanager, inputregion, fileuploadmode, fileupdatefilter)

- src/ could have then dbusadaptor and sailfishbrowser.cpp

rojkov:
I agree. I'd like to tackle tests and project layout refactoring.

PR #377, opened 2015-07-10
Revoke geolocation permissions upon app restart. 

rojkov:
Keep accepted/reject hosts in a hash instead of single values for accepted and rejected urls.

### DIFF
--- a/src/pages/components/LocationDialog.qml
+++ b/src/pages/components/LocationDialog.qml
@@ -16,6 +16,8 @@ import Sailfish.Silica 1.0
 UserPrompt {
     id: dialog
 
+    property string host
+
     //: Allow the server to use location
     //% "Allow"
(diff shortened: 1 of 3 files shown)


tworaz:
The patch LGTM. I just wonder if we plan to include it in 1.1.9 release. If so I don't know if we can change any translations.

rainemak:
Translation deadline has passed so as is this cannot be picked to 1.1.9. I don't think that this is a 1.1.9 blocker per se but would be good addition. Will check if this would be accepted still without the translation change.

rojkov:
The spec requires to show host name in UI. Without the UI change it's safer to eagerly revoke the permission thus I wouldn't take it to 1.1.9

rainemak:
@rojkov I agree, eager revoke is safe for 1.1.9.

rainemak:
LGTM

PR #378, opened 2015-07-13
Don't request Geolocation permission if gps is disabled.

rojkov:


rainemak:
Works as expected. LGTM.

PR #380, opened 2015-07-13
[sailfish-browser] WIP: Improve tab thumbnailing logic. Fixes JB#23670

rainemak:
Do not drop the thumbnail when switching between tabs, only drop the thumbnail when the url has really changed. As we clear with white, we allow capturing from 4th frame onwards (number is out of my head) to increase likely hood that some sane content has been rendered.

This also adds grayish background for the tab item and the actual thumbnail is faded in. Light gray background for the tab item needs some design feedback but this looks imo better than e.g. highlight dimmer color ( @roppola ).

### DIFF
--- a/src/declarativetabmodel.cpp
+++ b/src/declarativetabmodel.cpp
@@ -179,6 +179,15 @@ void DeclarativeTabModel::newTab(const QString &url, const QString &title, int p
     emit newTabRequested(url, title, parentId);
 }
 
+QString DeclarativeTabModel::url(int tabId) const
+{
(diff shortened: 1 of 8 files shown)


rojkov:
Couldn't crash browser with this PR. The unit tests for declarativetabmodel fail though.

rainemak:
Will check the unit tests.

rainemak:
Rebased and updated.

rainemak:
Closing in favour of https://github.com/sailfishos/sailfish-browser/pull/411

PR #384, opened 2015-07-17
[sailfish-browser] Fix the browser cover not showing when there are no open tabs. Fixes JB#30818

adenexter:
…o open tabs. Fixes JB#30818

Inform ApplicationWindow rather than the cover window that the web
view is the main application window.


rainemak:
LGTM

PR #385, opened 2015-07-17
[sailfish-browser] Reset glScissor state before clering the surface. Fixes JB#30882

tworaz:
Gecko's CompositorOGL can use glScissor to limit the rendering area when
compositing pages containing inline surfaces (example: iframes). In case
we've closed a view which has set some scissor box the clear operation
will not actually clear the whole surface, but only part of it. The
remaining area will have undefined content. This patch ensures that we
always reset the scissor box before clearing the surface. We currently
clear the surface manually only when no webview is opened so there is no
risk we'll affect operation of some existing CompositorOGL instance.
There shouldn't be any when the reset occurs.

### DIFF
--- a/src/declarativewebcontainer.cpp
+++ b/src/declarativewebcontainer.cpp
@@ -473,6 +473,8 @@ void DeclarativeWebContainer::clearWindowSurface()
     m_context->makeCurrent(this);
     QOpenGLFunctions_ES2* funcs = m_context->versionFunctions<QOpenGLFunctions_ES2>();
     Q_ASSERT(funcs);
+    QSize s = m_context->surface()->size();
+    funcs->glScissor(0, 0, s.width(), s.height());
     funcs->glClearColor(1.0, 1.0, 1.0, 0.0);
     funcs->glClear(GL_COLOR_BUFFER_BIT);
     m_context->swapBuffers(this);


src/declarativewebcontainer.cpp
```diff
@@ -474,6 +474,7 @@ void DeclarativeWebContainer::clearWindowSurface()
     QOpenGLFunctions_ES2* funcs = m_context->versionFunctions<QOpenGLFunctions_ES2>();
     Q_ASSERT(funcs);
     funcs->glClearColor(1.0, 1.0, 1.0, 0.0);
+    funcs->glScissor(0, 0, width(), height());
```
rainemak: Or like:
```   
Q_ASSERT(m_context->surface());
QSize s = m_context->surface()->size();
funcs->glScissor(0, 0, s.width(), s.height());
funcs->glClearColor(1.0, 1.0, 1.0, 0.0);
```

tworaz: Yep, technically we're clearing the surface, so we should use it's size, not the window. It just happens that in this case the surface is the same size as the window.. Fixed in next version.

rainemak:
LGTM. Cherry-pick also to upgrade-1.1.9 branch.

PR #386, opened 2015-07-17
Reorganize project structure

rojkov:
The code is split to 5 subcomponents now:

1. core
2. history
3. bookmarks
4. qtmozembed
5. storage

The names of the last three more or less reflect their purposes. The first two will need changes. E.g. the history subcomponent is just a collection of list models including the history model coincidently.

The added unit tests for `DeclarativeWebContainer` is an attempt to test its methods in isolation from other code. But without proper mocks it's nearly impossible since the class uses almost all methods of underlaying `QOpenglWebPage` class.


tests/auto/runtests.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+BASEDIR=$(dirname $0)
+PWD=`(cd $BASEDIR && pwd)`
+
+export LD_LIBRARY_PATH=$PWD/../../../qtmozembed/objdir-mer-dbg/src
```
rainemak: It should be possible to run this without own qtmozembed build.

rojkov: This script is supposed to run in a development environment where we always have a qtmozembed build and tst_* apps are not installed to their respective targets. Its purpose is to produce code coverage statistics in build directories.

rojkov: Ah.. you probably meant the script should be able to run in SDK. I'll need to install it first.

rainemak: Yeap, exactly.

rojkov:
Tweaked spec to make the project build with SailfishSDK.

Now `runtests.sh` is basically two scripts in one file: it detects where the tests are resided and runs them either from the same directory where the script was put or from respective subdirectories. It is not a replacement for the test runner that uses `tests.xml`, just a convenience tool.

rojkov:
Dropped tst_declarativetabmodel test case completely. The replacement is tst_persistenttabmodel: it tests methods in isolated and controlled environment and provides >90% code coverage.

rainemak:
Please, no test changes into "reorg" PR. Already difficult to see what's just moving and what is new.

tests/auto/auto.pro
```diff
@@ -5,6 +5,7 @@ SUBDIRS += tst_dbmanager \
     tst_declarativebookmarkmodel \
     tst_declarativehistorymodel \
     tst_declarativetabmodel \
+    tst_declarativewebcontainer \
```
rainemak: Add this one to the tests.xml

src/qtmozembed/qtmozembed.pri
```diff
@@ -0,0 +1,18 @@
+INCLUDEPATH += $$PWD
```
rainemak: Maybe this should be called something like "engine" so that it would not be tightly coupled to the qtmozembed. We clearly need to isolate this one from the "core".

tests/auto/test_common.pri
```diff
@@ -1,17 +1,14 @@
...
 
-INCLUDEPATH += ../../../src
-
-isEmpty(NO_COMMON_INCLUDES) {
-    include(../../src/common.pri)
-    include(../../src/history.pri)
-    include(common/downloadmanager_mock.pri)
-    include(common/declarativewebutils_mock.pri)
+# Put a file named gcov_enabled to this directory in order to enable code coverage
+exists(gcov_enabled) {
```
rainemak: Just ```CONFIG(gcov) {``` so that CONFIG+=gcov can be used to enable this.

rojkov:
Added tst_declarativewebcontainer to tests.xml and check for gcov in CONFIG.

rainemak:
LGTM

PR #389, opened 2015-07-20
Set of small fixes for the overlay

rainemak:
- Use animator for animating favorite grid and history list
- Don't unnecessary open virtual keyboard
- Enable clipping for the overlay only when absolutely mandatory


tworaz:
LGTM

rojkov:
LGTM

PR #391, opened 2015-07-23
[sailfish-browser] Add "like Andoird" token to user agent string. Contributes to JB#30687

tworaz:
The "like Android" keyword goal is to tell the servers that we're not
really Android, but would like to get the same content. Such token is
for example used in Windows mobile 8.1 browser.


rojkov:
LGTM

rainemak:
I think that this "breaks" more than fixes. Noticed that some sites start providing mobile friendly web pages just my removing "Linux" keyword from the user agent.

tworaz:
Dropping for now. Besides breaking youtube.com it also exposes some bugs in gecko v31 viewport support.

PR #393, opened 2015-07-24
Make mock objects configurable to check intended SUT's behaviour

rojkov:


tests/auto/tst_persistenttabmodel/tst_persistenttabmodel.cpp
```diff
@@ -60,7 +63,11 @@ private slots:
     void closeActiveTab();
     void updateUrl_data();
     void updateUrl();
+    void updateThumbnailPath();
+    void onUrlChanged();
     void onTitleChanged();
+    void nextActiveTabIndex();
+    void misc();
```
rainemak: Just based on the name this feel suspicious.

tests/auto/tst_persistenttabmodel/tst_persistenttabmodel.cpp
```diff
@@ -428,11 +494,45 @@ void tst_persistenttabmodel::onTitleChanged()
...
+
+    EXPECT_CALL(container, webPage()).WillRepeatedly(Return(&page));
+    EXPECT_CALL(page, parentId()).WillOnce(Return(1));
+    EXPECT_CALL(page, tabId()).WillOnce(Return(2));
+    tabModel->nextActiveTabIndex(0);
+}
+
+void tst_persistenttabmodel::misc()
+{
+    // Call simple methods to increase coverage
```
rainemak: Increasing coverage by calling just random functions doesn't make much sense if there's nothing really tested.

For instance roleNames() coverage improvement could be done with a ListView + PersistentTabModel QML test and using valid and invalid roles in the delegate. ListView::currentItem in turn could be used to check that data is read to the correctly by using role names.

tests/auto/tst_persistenttabmodel/tst_persistenttabmodel.cpp
```diff
@@ -428,11 +494,45 @@ void tst_persistenttabmodel::onTitleChanged()
...
+
+    tabModel->roleNames();
+    QCOMPARE(tabModel->waitingForNewTab(), false);
+
+    QSignalSpy loadedChangedSpy(tabModel, SIGNAL(loadedChanged()));
+    tabModel->m_loaded = true;
+    tabModel->setUnloaded();
+    QCOMPARE(loadedChangedSpy.count(), 1);
```
rainemak: This could be part of loaded() test.

tests/auto/tst_persistenttabmodel/tst_persistenttabmodel.cpp
```diff
@@ -428,11 +494,45 @@ void tst_persistenttabmodel::onTitleChanged()
...
+
+    QSignalSpy newTabRequestedSpy(tabModel, SIGNAL(newTabRequested(QString, QString, int)));
+    tabModel->newTab(QString("http://example.com"), QString("Test"), 0);
+    QCOMPARE(newTabRequestedSpy.count(), 1);
+    QCOMPARE(tabModel->waitingForNewTab(), true);
```
rainemak: And this requestNewTab() test.

tests/auto/tst_declarativetabmodel/tst_declarativetabmodel.cpp
```diff
@@ -1,493 +0,0 @@
...
-{
-    QFETCH(QString, url);
-
-    int tabId = currentTabId();
-    tabModel->updateUrl(tabId, url, false);
-    QCOMPARE(tabModel->activeTab().url(), url);
-    QTest::qWait(1000);
-}
-
-void tst_declarativetabmodel::invalidTabs_data()
```
rainemak: This could be moved as is to the tst_persistenttabmodel.cpp to test adding tabs with invalid urls. Later that should be extended to take into account also malformed urls like without protocol.

rainemak:
Couple of notes that are not related to this PR but could be fixed as a renaming commit on top:

```invalide_``` => ```invalid_``` (in tst_persistenttabmodel.cpp)

```activateTab_by_url``` => ```activateTabByUrl``` as other cases are camel cases as well (in tst_persistenttabmodel.cpp)

```igmore_``` => ```ignore_``` (in tst_persistenttabmodel.cpp) when closing the active tab.

File: tests/auto/tst_declarativetabmodel/tst_declarativetabmodel.cpp [hunk 6]
```diff
@@ -1,493 +0,0 @@
...
-
-    tabModel->activateTab(1);
-    QCOMPARE(currentTabId(), 4);
-
-    QCOMPARE(tabModel->activeTab().currentLink(), 12);
-    QString url = "http://www.foobar.com/something";
-    tabModel->updateUrl(currentTabId(), url, false);
-    QTest::qWait(1000);
-
-    QCOMPARE(tabModel->activeTab().tabId(), 4);
```
rainemak: ```tst_persistenttabmodel::activateTabById()``` could be extended to check also url and title of the active tab when activation is expected to succeed and fail.

rpm/sailfish-browser.spec
```diff
@@ -22,6 +22,8 @@ BuildRequires:  qt5-qttools
 BuildRequires:  qt5-qttools-linguist
 BuildRequires:  gdb
 BuildRequires:  oneshot
+BuildRequires:  gtest-devel
+BuildRequires:  libgmock-devel
```
rainemak: Should be under ```%package tests``` section.

tests/auto/tst_declarativetabmodel/tst_declarativetabmodel.cpp
```diff
@@ -1,493 +0,0 @@
-/****************************************************************************
```
rainemak: Remove also from tests.xml.

rainemak:
Generally speaking looking good! Please go through my comments.

tests/auto/tst_persistenttabmodel/tst_persistenttabmodel.cpp
rojkov: Agree, we need to add also QML tests.

For now I've split this case to 3 different cases dedicated to their respective methods. The case that simply calls a method is still useful since it makes sure the public method doesn't explode.

tests/auto/tst_persistenttabmodel/tst_persistenttabmodel.cpp
rojkov: I'd prefer to have test cases corresponding to tested methods and not to mix them.

Any method is just a function which produces some output for some given input. In C++ the input also includes object state. Any usage scenario can be described in the terms of input and output, hence it should be enough to provide sufficient input in the *_data() methods in order to test any possible scenario. This approach also implies that overloaded methods should be tested separately because they accept different input (activateTab(int), activateTab(QString)).

There might be situations when a tested method calls some other object (usually mocked one) which calls back synchronously and modifies the state of the tested object, but this smells like a bad architecture.

rpm/sailfish-browser.spec
rojkov: Why?
BuildRequires describes source package, not binaries.
Also rpm handles shared lib deps automatically.

rojkov:
Updated PR:

- tst_declarativetabmodel removed from tests.xml;
- added a separate test case for the scenario when addTab() is called with invalid input;
- got rid of misc() test case;
- typo fixes;
- additional checks for url and title in activateTab() test cases;
- added test case for activateTab(int).

rpm/sailfish-browser.spec
rainemak: IMO having ```BuildRequires``` at the top level is not good habit when the build requirement is not required by all packages that are going to be build. Thus, I'd specific this for test package only even thou "mock" and "test" are explaining this quite well :). More like a documentation thingie.

Ah right, ```Autoreq``` takes care of installation time dependency by checking compilation results.

rojkov: The main consumer of this info (apart from the build robots) is not a developer, but a package maintainer. In her perverted logic she builds the source package `sailfish-browser` that results in some build artifacts including coincidentally the binary package `sailfish-browser`. :)  This binary package can't be built unless all build deps are met. And it may save significant amount of time when hunting down build fails in the build machinery (e.g. when a package that we depend on gets renamed) if all the build deps are not scattered around the spec which is difficult to scan quickly due its size. Debian, for example, forces build deps to be included in the Source section of debian/control and for a good reason.
Probably I've spent too much time on package maintenance :)

rainemak: Ok, nevermind then. :)

rainemak:
LGTM

PR #397, opened 2015-07-27
[history] Use actual current URL when modifying session history.

rojkov:
By the time session history is restored WebPage already knows
its actual URL since this is one of two conditions required to
be met when DeclarativeWebPage::restoreHistory() is called.
This actual URL needs to be provided to Gecko's DocShell
when loading session history instead of the tab's initialURL,
because Gecko may alter the URL when loading a page.

For example initialURL instead of a valid URI may contain a text
input which will be recognized as a request to search engine.
Another example would be initialURL containing just a file path
without proto. This input will be recognized as an URL of
file:// protocol. And so on...


tworaz:
LGTM

rainemak:
LGTM

PR #402, opened 2015-08-11
Refactor storage subsystem

rojkov:
* Don't maintain `maxTabId` in `DBManager`. Initialize tab models' `nextTabId` once upon their creation instead.
* Encapsulate `Link` creation into `DBWorker` (`TabModel` doesn't have to know about this implementation detail). Also this makes possible  dropping `TabModel`'s `nextLinkId` members and their maintenance code.
* `TabModel`'s `nextTabId` property is used nowhere - drop it from public API.
* Reduce data exchange between SQLite and DBWorker by using SQL inner joins.
* Don't include `Link` to `Tab` as a member to avoid maintaining linkId in `Tab`.
* Fully cover the storage subsystem with unit tests.

### DIFF
--- a/src/history/persistenttabmodel.cpp
+++ b/src/history/persistenttabmodel.cpp
@@ -82,18 +82,14 @@ Tab PersistentTabModel::createTab(int tabId, QString url, QString title) {
     return DBManager::instance()->createTab(tabId, url, title);
 }
 
-void PersistentTabModel::updateTitle(int tabId, int linkId, QString url, QString title)
+void PersistentTabModel::updateTitle(int tabId, QString url, QString title)
 {
-    DBManager::instance()->updateTitle(tabId, linkId, url, title);
+    DBManager::instance()->updateTitle(tabId, url, title);
 }
 
 void PersistentTabModel::removeTab(int tabId)
 {
     DBManager::instance()->removeTab(tabId);
-}
-
-int PersistentTabModel::nextLinkId() {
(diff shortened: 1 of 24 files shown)


tworaz:
LGTM

src/history/persistenttabmodel.cpp
```diff
@@ -57,10 +57,15 @@ void PersistentTabModel::tabsAvailable(QList<Tab> tabs)
         emit countChanged();
     }
 
-    int maxTabId = DBManager::instance()->getMaxTabId();
+    int maxTabId(0);
+    foreach (Tab tab, tabs) {
```
rainemak: ```foreach (const Tab &tab, tabs) {```

rojkov:
Good catch! Also I've maid the ::tabsAvailable() slot less hungry for copying tabs.

tests/auto/tst_dbmanager/tst_dbmanager.cpp
```diff
@@ -11,125 +11,482 @@
```
rainemak: Cleanup tests.xml a bit. Remove ```-platform wayland-egl -iterations 10``` part from the tst_dbmanager case.

rainemak:
Added one comment about the tests.xml. Other than that LGTM now.

rojkov:
Updated tests.xml

rojkov:
Amended commit messages.

PR #403, opened 2015-08-13
[sailfish-browser] Allow Homescreen to show browser cover when launch…

blammit:
…ing URLs and change favorite desktop configuration. Fixes JB#30997

Use the org.sailfishos.browser.ui service rather than
org.sailfishos.browser when opening URLs via libcontentaction. This
menas the Home screen will show the browser app cover while the app is
loading if triggered via an app-open action e.g. Qt.openUrlExternally().

However, browser bookmarks in the app grid are created using the "Link"
desktop entry, which already causes a separate desktop app to be
launched. To prevent two app covers from being displayed in in this
case, the .desktop configuration for grid bookmarks now uses the
"Application" entry type rather than "Link", with the URL passed via
the "Exec" entry. So, instead of launching a Link type that is
interpreted as a content action that should open the browser, the
favourite will now be launched directly with the browser app.

### DIFF
--- a/src/desktopbookmarkwriter.cpp
+++ b/src/desktopbookmarkwriter.cpp
@@ -90,10 +90,10 @@ QString DesktopBookmarkWriter::write(QString url, QString title, QString icon)
 {
     QString fileName = uniqueDesktopFileName(title);
     QString desktopFileData = QString("[Desktop Entry]\n" \
-                                      "Type=Link\n" \
+                                      "Type=Application\n" \
                                       "Name=%1\n" \
                                       "Icon=%2\n" \
-                                      "URL=%3\n" \
+                                      "Exec=invoker -s --type=browser /usr/bin/sailfish-browser %3\n" \
                                       "Comment=%4\n").arg(title.trimmed(), icon,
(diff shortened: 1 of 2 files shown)


src/desktopbookmarkwriter.cpp
```diff
@@ -90,10 +90,10 @@ QString DesktopBookmarkWriter::write(QString url, QString title, QString icon)
 {
     QString fileName = uniqueDesktopFileName(title);
     QString desktopFileData = QString("[Desktop Entry]\n" \
-                                      "Type=Link\n" \
+                                      "Type=Application\n" \
                                       "Name=%1\n" \
                                       "Icon=%2\n" \
-                                      "URL=%3\n" \
+                                      "Exec=invoker -s --type=browser /usr/bin/sailfish-browser %3\n" \
```
rainemak: Please rebase and add ```-G``` option.

Other than that LGTM

blammit:
Sorry, I forgot to close this PR. The issue with the duplicate covers has been fixed in lipstick instead.

PR #405, opened 2015-08-19
[sailfish-browser] Use .ui service when launching browser for the purpose of opening a URL. Fixes JB#30997

blammit:
…pose of opening a URL. Fixes JB#30997


tworaz:
LGTM

PR #408, opened 2015-09-01
[sailfish-browser] Add environment variable for enabling qml debugging. Contributes to JB#6054

rainemak:
This way there is no need to patch browser separately if you'd like to e.g. do some QML profiling. Just start the browser from command line like

> QML_DEBUGGING_ENABLED=1 sailfish-browser -qmljsdebugger=port:6666

or when profiling startup just add block.

> QML_DEBUGGING_ENABLED=1 sailfish-browser -qmljsdebugger=port:6666,block

And connect remotely to the device from the QtCreator.

### DIFF
--- a/src/sailfishbrowser.cpp
+++ b/src/sailfishbrowser.cpp
@@ -11,6 +11,7 @@
 
 #include <QGuiApplication>
 #include <QQuickView>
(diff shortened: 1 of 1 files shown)


tworaz:
LGTM

PR #411, opened 2015-09-01
[sailfish-browser] Improve tab thumbnailing logic. Fixes JB#23670

rainemak:
It's a bit of guess work but now we capture the third frame and
when the page is fully loaded. In addition, we do not clear the thumbnails
that aggresively as we are capturing only the active web page. Down
side is more or less same as it is now. For instance, if you close the
browser and leave it closed for a while, restart it, go to tab view and
you find that there out-dated thumbnails not matching the reality.

To get updates working properly, we refresh model data so (name is always
the same, like tab-1-thumb.jpg) that the new updated capture gets
rendered to the tab view.

### DIFF
--- a/src/history/declarativetabmodel.cpp
+++ b/src/history/declarativetabmodel.cpp
@@ -177,6 +177,15 @@ void DeclarativeTabModel::newTab(const QString &url, const QString &title, int p
     emit newTabRequested(url, title, parentId);
 }
 
+QString DeclarativeTabModel::url(int tabId) const
+{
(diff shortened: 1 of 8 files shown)


rainemak:
Do not merge yet.

rojkov:
LGTM
I've tried the patch and it seems to work ok. A tab's thumbnail flickers two times now though, but I find it a reasonable tradeoff against obsolete thumbnails.

rojkov:
LGTM

PR #414, opened 2015-09-08
[sailfish-browser] Prevent gecko from rendering when screen is off. Fixes JB#31866.

tworaz:
When the screen goes off we'd like to prevent gecko from doing any
rendering. We don't want to completely pause the compositor, but we also
don't want it to execute any GL/EGL commands, especially eglSwapBuffers.
The reason for this is because it seems that on sailfish eglSwapBuffers
can currently block indefinitely when the display is turned off.

Blocked compositor thread was the main reason for huge memory leak
observed when the video was being played in the browser and the screen was
turned off. Gecko media thread produced new frame data which was never
consumed by the compositor thread. The data kept piling up until the
system run out of memory.

### DIFF
--- a/src/pages/components/ResourceController.qml
+++ b/src/pages/components/ResourceController.qml
@@ -20,6 +20,7 @@ Item {
     property QtObject webView
     property bool videoActive
     property bool audioActive
+    readonly property alias displayOff: screenBlanked.value
     property bool background
 
     property string _mediaState: "pause"
--- a/src/pages/components/WebView.qml
+++ b/src/pages/components/WebView.qml
@@ -97,7 +97,8 @@ WebContainer {
             width: container.rotationHandler && container.rotationHandler.width || 0
             fullscreenHeight: container.fullscreenHeight
             toolbarHeight: container.toolbarHeight
-            throttlePainting: !foreground && !resourceController.videoActive
+            throttlePainting: !foreground && !resourceController.videoActive || resourceController.displayOff
+            readyToPaint: !resourceController.displayOff
             enabled: webView.enabled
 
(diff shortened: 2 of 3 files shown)


src/pages/components/ResourceController.qml
```diff
@@ -20,6 +20,7 @@ Item {
     property QtObject webView
     property bool videoActive
     property bool audioActive
+    property alias displayOff: screenBlanked.value
```
rainemak: Looks like a read only property ```readonly property alias displayOff: screenBlanked.value```

src/pages/components/WebView.qml
```diff
@@ -97,7 +97,8 @@ WebContainer {
             width: container.rotationHandler && container.rotationHandler.width || 0
             fullscreenHeight: container.fullscreenHeight
             toolbarHeight: container.toolbarHeight
-            throttlePainting: !foreground && !resourceController.videoActive
+            throttlePainting: !foreground && !resourceController.videoActive || resourceController.displayOff
+            readyToPaint: !resourceController.displayOff
```
rainemak: This property does not exist.

rainemak:
Requires: https://github.com/nemomobile-packages/qtmozembed/pull/12

src/pages/components/ResourceController.qml
tworaz: Right, fixed in next version.

tworaz:
This is a final patch in 3 part series. It's dependencies can be found at:
https://github.com/nemomobile-packages/gecko-dev/pull/49
https://github.com/nemomobile-packages/qtmozembed/pull/12

PR #418, opened 2015-09-15
Move WebPage creation to a factory object.

rojkov:
* Decouple `BookmarkManager` from `SettingsManager` (this is the reason for simplification of private data settings).
* Factor out `DeclarativeWebPage` creation to a dedicated factory object. This way it's easier to test the `WebPages` class and different scenarios of web page creation.
* Add unit tests for `WebPages`.

### DIFF
--- a/src/core/declarativewebcontainer.h
+++ b/src/core/declarativewebcontainer.h
@@ -70,7 +70,7 @@ class DeclarativeWebContainer : public QWindow, public QQmlParserStatus, protect
     Q_PROPERTY(bool privateMode READ privateMode WRITE setPrivateMode NOTIFY privateModeChanged FINAL)
     Q_PROPERTY(bool activeTabRendered READ activeTabRendered NOTIFY activeTabRenderedChanged FINAL)
 
-    Q_PROPERTY(QQmlComponent* webPageComponent MEMBER m_webPageComponent WRITE setWebPageComponent NOTIFY webPageComponentChanged FINAL)
+    Q_PROPERTY(QQmlComponent* webPageComponent READ webPageComponent WRITE setWebPageComponent NOTIFY webPageComponentChanged FINAL)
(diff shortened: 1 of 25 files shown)


src/core/declarativewebcontainer.h
```diff
@@ -70,7 +70,7 @@ class DeclarativeWebContainer : public QWindow, public QQmlParserStatus, protect
     Q_PROPERTY(bool privateMode READ privateMode WRITE setPrivateMode NOTIFY privateModeChanged FINAL)
     Q_PROPERTY(bool activeTabRendered READ activeTabRendered NOTIFY activeTabRenderedChanged FINAL)
 
-    Q_PROPERTY(QQmlComponent* webPageComponent MEMBER m_webPageComponent NOTIFY webPageComponentChanged FINAL)
+    Q_PROPERTY(QQmlComponent* webPageComponent MEMBER m_webPageComponent WRITE setWebPageComponent NOTIFY webPageComponentChanged FINAL)
```
rainemak: Even thou above is a valid property declaration, I'd change ```MEMBER m_webPageComponent``` to ```READ webPageComponent``` and add getter.

rojkov:
The new syntax looks ok to me. We don't use overloaded slots usually, but I remember there are some places where we rely on default values for slot arguments. I'd try to move to the new syntax in the scope of a separate task.

src/core/declarativewebcontainer.h
rojkov: Added getter for the webPageComponent prop.

tests/auto/tests.xml
```diff
@@ -30,6 +30,11 @@
            </case>
            <case manual="false" name="webutils">
                <step>cd /opt/tests/sailfish-browser/auto/ &amp;&amp; ./tst_webutils</step>
```
rainemak: Missing ```</case>```

rojkov:
Oops, added missing </case> and rebased to next.

rainemak:
LGTM

PR #420, opened 2015-09-24
Add stricter build flags

rainemak:
Treat warnings as errors, nit pick about parentheses, and
abort build upon error.

### DIFF
--- a/src/downloadmanager.cpp
+++ b/src/downloadmanager.cpp
@@ -140,7 +140,8 @@ bool DownloadManager::moveMyAppPackage(QString path)
         uid_t uid = getuid();
         // assumes that correct groupname is same as username (e.g. nemo:nemo)
         int gid = getgrnam(getpwuid(uid)->pw_name)->gr_gid;
-        chown(aptoideDownloadPath.toLatin1().data(), uid, gid);
+        int success = chown(aptoideDownloadPath.toLatin1().data(), uid, gid);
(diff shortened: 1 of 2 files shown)


src/downloadmanager.cpp
```diff
@@ -140,7 +140,8 @@ bool DownloadManager::moveMyAppPackage(QString path)
         uid_t uid = getuid();
         // assumes that correct groupname is same as username (e.g. nemo:nemo)
         int gid = getgrnam(getpwuid(uid)->pw_name)->gr_gid;
-        chown(aptoideDownloadPath.toLatin1().data(), uid, gid);
+        int success = chown(aptoideDownloadPath.toLatin1().data(), uid, gid);
+        Q_UNUSED(success);
```
rainemak: Using ```Q_UNUSED(chown(aptoideDownloadPath.toLatin1().data(), uid, gid));``` would expand to ```(void)chown(aptoideDownloadPath.toLatin1().data(), uid, gid);``` which does not work.

tworaz: How is that possible? Casting function return value to void is one of the most common methods of silencing unused return value compiler warnings in C/C++.

rainemak: I don't know. Imo this is not a big deal.

tworaz:
LGTM

PR #422, opened 2015-09-25
Isolate declarativebookmarkmodel test cases

rojkov:


tests/auto/tst_declarativebookmarkmodel/tst_declarativebookmarkmodel.cpp
```diff
@@ -9,211 +9,227 @@
...
+    QString m_bookmarksFile;
 };
 
-
-tst_declarativebookmarkmodel::tst_declarativebookmarkmodel()
-    : TestObject(QML_SNIPPET)
+void tst_declarativebookmarkmodel::initTestCase()
 {
-    bookmarkModel = TestObject::qmlObject<DeclarativeBookmarkModel>("bookmarkModel");
+    QString settingsLocation = QStandardPaths::writableLocation(QStandardPaths::DataLocation);
```
rainemak: Maybe worth adding in another PR something like ```BrowserPaths``` class that would have static methods like.

```
static QString downloadLocation()
static QString picturesLocation()
static QString dataLocation()
...
```
This way we could clean up a bit the webutils thingie and unit test cases would not contain assumptions about paths.

rojkov: Yep, I agree.

tests/auto/tst_declarativebookmarkmodel/tst_declarativebookmarkmodel.pro
```diff
@@ -1,15 +1,13 @@
 TARGET = tst_declarativebookmarkmodel
 
-QT += quick concurrent
+QT += network concurrent
```
rainemak: What component requires this one?

rainemak:
Added a small question? Works as specified, LGTM. If question don't trigger actions, go ahead and merge.

tests/auto/tst_declarativebookmarkmodel/tst_declarativebookmarkmodel.pro
rojkov: `src/bookmarks/iconfetcher.cpp` (included by `src/bookmarks/bookmarks.pri`) needs it. I think it makes sense to refactor the tests not to depend on iconfetcher, but probably in a separate PR.

rainemak: Yeap. Let's have a separate PR about that. So looks good.

PR #425, opened 2015-09-28
[sailfish-browser] Position tabview when opened in landscape. Fixes JB#31994

pvuorela:


src/pages/components/Overlay.qml
```diff
@@ -417,6 +417,11 @@ Background {
 
             onStatusChanged: browserPage.tabPageActive = (status == PageStatus.Active)
 
+            Component.onCompleted: {
```
rainemak: Place of ```Component.onCompleted``` (attached properties) has not been clearly defined, my personal preference for ```Component.onCompleted``` would be at the end of the ```tabPage```.

The ```positionViewAtIndex``` looks like obsolete now as this completes after tabView. Please remove if not really needed.

pvuorela: I don't understand the 'looks like obsolete' part. If I remove it here I have no changes. If I remove it from TabView I'm relying on undefined execution order being defined as I see it happening.

pvuorela:
Moved signal handler position.

rainemak:
LGTM

PR #426, opened 2015-09-29
[browserpaths] Move directory functions to dedicated module.

rojkov:


src/declarativewebutils.cpp
```diff
@@ -342,14 +342,9 @@ DeclarativeWebUtils *DeclarativeWebUtils::instance()
     return gSingleton;
 }
 
-QString DeclarativeWebUtils::downloadDir() const
-{
-    return QStandardPaths::writableLocation(QStandardPaths::DownloadLocation);
-}
-
 QString DeclarativeWebUtils::picturesDir() const
```
rainemak: We should drop this as well. Only used in ```BrowserContextMenu``` in its ```getUniqueFileName```.

The ```getUniqueFileName``` is also one of the places that uses ```WebUtils.fileExists```, the other place is in webutils.cpp itself. That makes me wondering if we should have some kind of FileUtils for creating ```uniqueFileName(fileName)```. I do not recall by heart if/where we are doing file checks etc on QML side.

To get this removed, you could add ```BrowserPaths``` to the root context. Better approach, that would not add extra stuff to root context, would be if we'd have a QObject wrapper and register this as singleton. ```BrowserPaths``` cannot be directly registered as singleton as qmlengine will take the ownership of it.

rojkov:
Dropped WebUtils.picturesDir.
I decided not to add anything to root context, but to move the JS code that needs the paths to C++ side.

src/declarativewebutils.cpp
```diff
@@ -75,10 +81,35 @@ int DeclarativeWebUtils::getLightness(QColor color) const
     return color.lightness();
 }
 
-bool DeclarativeWebUtils::fileExists(QString fileName) const
+QString DeclarativeWebUtils::getUniquePictureName(const QString& fileName) const
```
rainemak: Drop the ```get```.

common/browserpaths.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include <QDir>
+#include <QStandardPaths>
+#include "browserpaths.h"
+
+static QString getLocation(QStandardPaths::StandardLocation locationType) {
+    QString location(QStandardPaths::writableLocation(locationType));
+    QDir dir(location);
+    if (!dir.exists()) {
+        if (!dir.mkpath(location)) {
+            qWarning(QString("Can't create directory ").append(location).toLatin1().data()) ;
```
rainemak:
```
qWarning() << QString("Can't create directory %1").arg(location);
```

common/browserpaths.cpp
```diff
@@ -0,0 +1,53 @@
...
+** Copyright (C) 2015 Jolla Ltd.
+** Contact: Dmitry Rozhkov <dmitry.rozhkov@jolla.com>
+**
+****************************************************************************/
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include <QString>
```
rainemak: ```#include <QDebug>``` for qWarning

src/declarativewebutils.cpp
```diff
@@ -75,10 +81,35 @@ int DeclarativeWebUtils::getLightness(QColor color) const
...
 {
-    QFile file(fileName);
-    return file.exists();
+    int collisionCount(0);
+    const QString picturesDir(BrowserPaths::picturesLocation());
```
rainemak: Wrote this out of my head but it should depict the idea. Rather than mangling the file manually utilize QFileInfo instead.

```
...
QString result = picturesDir + "/" + fileName;
QFileInfo fileInfo(fileName);
// Path, base name, count, and suffix
QString newFile("%1/%2(%3).%4");
QString baseName = fileInfo.baseName();
QString suffix = fileInfo.completeSuffix();
while (QFileInfo::exists(result)) {
    collisionCount++;
    result = newFile.arg(picturesDir).arg(baseName).arg(collisionCount).arg(suffix);
}
...
```

rojkov:
Good point. Looks much simpler now.

src/declarativewebutils.cpp
```diff
@@ -75,10 +81,23 @@ int DeclarativeWebUtils::getLightness(QColor color) const
...
+    const QString newFile("%1/%2(%3)%4%5");
+    const QString baseName = fileInfo.baseName();
+    const QString suffix = fileInfo.completeSuffix();
+
+    QString result(picturesDir + "/" + fileName);
+    int collisionCount(1);
+
+    while (QFileInfo::exists(result)) {
+        collisionCount++;
+        result = newFile.arg(picturesDir).arg(baseName).arg(collisionCount).arg(suffix.isEmpty() ? "": ".").arg(suffix);
```
rainemak: Small nit... space missing in ```suffix().isEmpty() ? "":``` between last quotation mark and colon.

rainemak:
LGTM

src/declarativewebutils.cpp
rojkov: done

rojkov:
Rebased and tested with unit tests.

PR #428, opened 2015-10-01
Open overlay on browser startup if no tabs exists

rojkov:


src/core/declarativewebcontainer.cpp
```diff
@@ -738,15 +738,14 @@ void DeclarativeWebContainer::initialize()
...
-    // 2) model has tabs, load initial url or active tab.
-    bool firstUseDone = DeclarativeWebUtils::instance()->firstUseDone();
-    if (m_model->count() == 0 && (firstUseDone || !m_initialUrl.isEmpty())) {
-        QString url = m_initialUrl.isEmpty() ? DeclarativeWebUtils::instance()->homePage() : m_initialUrl;
-        QString title = "";
-        m_model->newTab(url, title);
-    } else if (m_model->count() > 0) {
+    if (m_model->count() == 0) {
+        if (m_initialUrl.isEmpty()) {
+            emit overlayRequested();
```
rainemak: This is already too late moment as you can see the toolbar visible at the bottom before overlay is opened and there is no need to wait for gecko context initialization. In theory tabModel.loaded == true plus m_initialUrl initialized is enough information to figure if we need to show overlay or normal toolbar. So, I'd try to avoid this signal all together.

In practice this means that when tabModel is loaded and openUrlRequested signal handler has been executed around BrowserPage.qml 311 we are ready to decide if we want to show chrome to overlay.

In OverlayAnimator.qml we should not do extra work when starting browser. Meaning that initial state should be ``` _noOverlay```  and also initial position of the overlay should be hidden.

rojkov:
Updated the PR not to show toolbox too early. Personally I don't like the fact that additional state is preserved to track that both openUrlRequested() and tabModel.loadedChanged() signals were fired, but this is the best I could come up with atm.

rainemak:
Review comments:

- Cover state is not updated when starting browser without tabs

- Overlay flickers on the atTop state when browser started from external link so that we don't have tabs from previous browsing session (not happening always)

rojkov:
clean up open PRs

PR #433, opened 2015-10-14
[sailfish-browser] Fix entering and exiting fullscreen mode. Fixes JB#32814

rainemak:
The chrome state of web page should not affect when updating margins. Use shorter page only when contentHeight is such that gesture cannot be used to go to fullscreen mode.


rojkov:
LGTM

PR #434, opened 2015-10-16
Remove fixed press and hold delay

rainemak:
Includes two commits. First one sorts Qt related includes in webutils (should be done everywhere).

### DIFF
--- a/src/declarativewebutils.cpp
+++ b/src/declarativewebutils.cpp
@@ -15,6 +15,7 @@
 #include <QtCore/QFile>
 #include <QtCore/QFileInfo>
 #include <QtCore/QLocale>
+#include <QtCore/QSettings>
 #include <QtCore/QStandardPaths>
 #include <QtCore/QStringList>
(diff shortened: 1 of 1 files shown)


jpetrell:
Works like a charm. :) :+1:

rojkov:
LGTM

src/declarativewebutils.cpp
```diff
@@ -43,6 +47,19 @@ bool testScreenDimensions(qreal pixelRatio) {
     return fmod(w, 1.0) == 0 && fmod(h, 1.0) == 0;
 }
 
+const QSettings &quickSettings()
+{
+    static const QSettings settings(QSettings::SystemScope, QStringLiteral("QtProject"), QStringLiteral("QtQuick2"));
```
matthewvogt: No need for 'static' here; you only access the settings once, so it won't be helpful to keep them in memory.

rainemak: That's right.

PR #439, opened 2015-10-22
[sailfish-browser] Update margins and send vkb metrics when vkb is open. Fixes JB#32818

rainemak:
Metrics are updated to contain bottomMargin and window (screen) size.
With this data all calculations can be done in embedlite-components.


rojkov:
LGTM

PR #444, opened 2015-11-02
[sailfish-browser] Do not update view margins upon suspend/resume. Fixes JB#33119

rainemak:
Updating bottom margin upon suspend/resume may trigger infinite
rendering loop between zero bottom margin and toolbar height margin.

In addition to above, require at least 1.5x times fullscreen height
in order to set the bottom margin. Setting bottom margin normally
triggers another content height change as the available height
changed. Normal case when bottom margin is set to zero is when
a web page has plenty of content.

Bottom margin with virtual keyboard is updated separately.


rojkov:
LGTM.
With the patch applied the infinite loop is not reproducible.

PR #445, opened 2015-11-02
Improve resetting margins

rainemak:
Reset margins when a page is loaded and when fullscreen height
changes.

### DIFF
--- a/src/qtmozembed/declarativewebpage.cpp
+++ b/src/qtmozembed/declarativewebpage.cpp
@@ -95,6 +95,12 @@ DeclarativeWebPage::DeclarativeWebPage(QObject *parent)
     connect(&m_grabWritter, SIGNAL(finished()), this, SLOT(grabWritten()));
     connect(this, SIGNAL(urlChanged()), this, SLOT(onUrlChanged()));
     connect(this, &QOpenGLWebPage::contentHeightChanged, this, &DeclarativeWebPage::updateViewMargins);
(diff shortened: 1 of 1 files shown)


src/qtmozembed/declarativewebpage.cpp
```diff
@@ -95,6 +95,12 @@ DeclarativeWebPage::DeclarativeWebPage(QObject *parent)
     connect(&m_grabWritter, SIGNAL(finished()), this, SLOT(grabWritten()));
     connect(this, SIGNAL(urlChanged()), this, SLOT(onUrlChanged()));
     connect(this, &QOpenGLWebPage::contentHeightChanged, this, &DeclarativeWebPage::updateViewMargins);
+    connect(this, &QOpenGLWebPage::loadedChanged, this, [=]() {
```
rojkov: I guess this form of `connect` should accept only 3 arguments: the second `this` is not needed. Also it's recommended to avoid capturing the whole context: `this` instead of `=` should be enough.

rainemak: Good point. Dropped 3rd argument (2nd this) and used ```this``` instead of ```=```.  In real life kind of simple lambda case there would not be any diff as we're not copying anything inside the lambda.

rainemak:
Good point. Dropped 3rd argument (2nd this) and used this instead of =. This kind of simple lambda case there would not be any diff as we're not copying anything inside the lambda.

rojkov:
LGTM

PR #446, opened 2015-11-03
[prefs] Increase async scroll event throttling and timeout intervals. Contributes to JB#32858

rainemak:
We're not relying on async scroll events. So lets fire them less often than the default value (100) or B2G (15). Currently only toolbar is interested about async scroll events and it's not urgent to be updated immediately.


rojkov:
LGTM

PR #447, opened 2015-11-05
[sailfish-browser] Update rendering related preferences

rainemak:
Clear only once per frame. Without gfx.compositor.clear-context
both drawUnderlay and CompositorOGL would clear.

The embedlite.compositor.request_external_gl_context_early is not
needed EmbedLiteWindowBaseChild::CreateWidget instantiates EmbedLitePuppetWidget
and updates size for it. Due to size update the normal compositor
creation code path will get executed.

EmbedLitePuppetWidget::UpdateSize =>
EmbedLitePuppetWidget::Resize =>
EmbedLitePuppetWidget::Invalidate =>
EmbedLitePuppetWidget::GetLayerManager =>
EmbedLitePuppetWidget::CreateCompositor


rojkov:
LGTM

PR #448, opened 2015-11-05
[sailfish-browser] Adapt to new download manager. 

rojkov:
Adaptations basically means that now
* we don't need to wait until nsDownloadManager is initialized in order start URL download upon browser launch;
* there is no need to set URL scheme `file://` when downloading an URL to a local file target.


rainemak:
LGTM

PR #450, opened 2015-11-12
[qml] Remove PageStackAction.Replace for about:config dialog. Fixes JB#32873

rojkov:
…B32873

### DIFF
--- a/src/pages/components/ConfigWarning.qml
+++ b/src/pages/components/ConfigWarning.qml
@@ -19,7 +19,6 @@ ConfirmDialog {
     //: Warning of changing browser configurations.
     //% "Changing these advanced settings can cause issues with stability, security and performance of Sailfish Browser. Continue ?"
     text: qsTrId("sailfish_browser-la-config-warning");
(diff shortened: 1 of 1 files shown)


src/pages/components/ConfigWarning.qml
```diff
@@ -19,7 +19,6 @@ ConfirmDialog {
     //: Warning of changing browser configurations.
     //% "Changing these advanced settings can cause issues with stability, security and performance of Sailfish Browser. Continue ?"
     text: qsTrId("sailfish_browser-la-config-warning");
-    acceptDestinationAction: PageStackAction.Replace
```
siteshwar: Sounds like more of a workaround than actual fix.

rojkov: Yes, it is. There's a bug in Silica.

siteshwar:
LGTM

src/pages/components/ConfigWarning.qml
rainemak: As this fixes bigger issue LGTM.

PR #456, opened 2016-01-13
[sailfish-browser] Open external urls in new tab. Fixes JB#19746

### LINKED ISSUE #454: Open external urls to new tab
Implement external link opening to a new tab.

I'd believe that best would be if all external urls would be opened to new tabs. However, if community likes a setting more, we could have a settings for this. Something like:

"/apps/sailfish-browser/settings/open_external_url_to_new_tab"

siteshwar:
Fixes #454

### DIFF
--- a/src/pages/BrowserPage.qml
+++ b/src/pages/BrowserPage.qml
@@ -319,8 +319,8 @@ Page {
 
             webView.grabActivePage()
             if (!webView.tabModel.activateTab(url)) {
-                // Not found in tabs list, load it. A new tab will be created if no tabs exist.
-                webView.load(url)
+                // Open new tab with empty title
+                webView.tabModel.newTab(url, "")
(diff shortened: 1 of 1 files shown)


rainemak:
Discussed with design and this behaviour is good. No need to show notifications or such as we have that rotating tabs icon showing that something is happening for tabs (rotating clockwise).

This should be going to next branch as this is not Gecko 38 specific but I can cherry-pick it there. No need to open new PRs.

Regarding commit message, please remove the Fixes #454 from the commit message body. I'm not 100% sure if our CI infra handles that correctly. Instead of having that you could amend the internal bug id 19746 to the commit message subject:

```[sailfish-browser] Open external urls in new tab. Fixes JB#19746```

rainemak:
Thanks @siteshwar !

SfietKonstantin:
Wow thanks !

PR #457, opened 2016-01-18
[sailfish-browser] Fix external link load when starting browser. Fixes JB#33892

rainemak:
Contains two small commits. Probably still easier to review per commit.

Main problem was / is that the previous active tab got loaded when browser was launched so that there were tabs from previous browsing session. The first commit sha1 78efdd9 fixes that. The second commit sha1 41d4b76 fixes overlay flickering when browser launched via external link so that there were no tabs from previous browsing session -- overlay appears quickly on top position from where changed to chrome mode.

### DIFF
--- a/org.sailfishos.browser.service
+++ b/org.sailfishos.browser.service
@@ -1,3 +1,3 @@
 [D-BUS Service]
 Name=org.sailfishos.browser
-Exec=/usr/bin/invoker -s --type=browser -G /usr/bin/sailfish-browser
+Exec=/usr/bin/invoker -s --type=browser -G /usr/bin/sailfish-browser -prestart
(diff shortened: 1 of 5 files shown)


siteshwar:
I cherry picked these changes to nemo_embedlite_38 branch and I see no change in behavior with respect to starting browser when external urls are clicked. Since 3fe19d4a7d7b38da4f1d2c181d9efaa61b5cb838 browser seems to freeze when it's started by tapping external url and this PR does not seem to fix this issue.

rainemak:
Embedlite 38 is a different topic. I'm 99% sure (without further checking) that there's a bug also with Gecko 38. It's not about timing, the initialization is just wrong if/when browser start by opening a new tab.

jpetrell:
Fixed the issue, no flickering seen. Code changes look sane.

LGTM

PR #458, opened 2016-02-10
[sailfish-browser] Add "Save link" context menu action.

rainemak:


src/pages/components/WebPopupHandler.js
```diff
@@ -136,9 +136,16 @@ function openContextMenu(data) {
 
         if (_contextMenu) {
             _contextMenu.linkHref = linkHref
-            _contextMenu.linkTitle = linkTitle.trim()
             _contextMenu.imageSrc = imageSrc
+            _contextMenu.linkTitle = linkTitle.trim()
+            _contextMenu.linkProtocol = data.linkProtocol || ""
```
rainemak: Amended a fallback for linkProtocol. Image urls do not specify protocol.

Long press on top of an image, select open to new tab, on the tab new long press on top of the opened image.

jpetrell:
On some files (like rpms) "Open link" and "Save link" do the same thing, would be nice to hide the open option if browser cannot handle the content type

For photos and media I would expect "Copy to clipboard" to copy the content, maybe rename to "Copy link"?

I would maybe simplify naming "Open link | Open link in new  tab | Share | Copy to clipboard | Save link" -> "Open | Open in new tab | Share | Copy | Save"

src/pages/components/BrowserContextMenu.qml
```diff
@@ -28,6 +29,11 @@ Rectangle {
     readonly property bool active: visible
     readonly property bool landscape: width > height
 
+    readonly property string url: !isJavascriptFunction ? linkHref : ""
+    readonly property bool isJavascriptFunction: linkProtocol === "javascript"
+    readonly property bool isLink: !isJavascriptFunction && linkHref.length > 0 && imageSrc.length === 0
```
siteshwar: Should we ignore other protocols for e.g. tel, mailto etc. here ?

siteshwar: May be reuse ```isImage``` variable here :
```
    readonly property bool isLink: !isJavascriptFunction && !isImage && linkHref.length > 0
```

rainemak: @jpetrell 
Having either "Open link" or "Save link" would be definitely nice. However, that would require much more changes and some new apis. Something to be fixed in future.

Long pressing on top of a photo you'll get "Open image in new tab" and "Save to gallery". Identifying what's behind a link requires some sort of partial prefetching of a href content. Engines have been handling prefetching while engine is idling for some time (link element) but it really cannot/should not be arbitrary prefetching. 

I'm fine with dropping "link" suffix. 

Google Chrome doesn't have "Open link" options in context menu but they save "Open in new tab" and "Save link". Both Android Firefox and Google Chrome have more context information about the data. We added "Open link" to the menu as somebody was asking about it => personally I could even think of dropping it. One use case that I can think of is that rather clicking a suspicious link directly, you'd like to check that link is somewhat sane.

Furthermore, Google Chrome logic hides "Save link" when long pressing a article on  m.yle.fi/uutiset.  However, they still have "Copy link address" which doesn't make any sense as that is a javascript function. Also "Open in new tab" works => I just hide it as it is trying to open "void(0)" and it just doesn't work in our context => improvements needed on the embedlite-components side.

What comes to context menu handling in general there's plenty of room to improve when comparing to competitors browsers. The Metro code (Windows Firefox) didn't really help here: https://github.com/tmeshkova/embedlite-components/pull/94 => they have some what similar approach as we.

As a summary about this quick fix, this handles basic stuff :)

@siteshwar :
tel, mailto, geo etc are valid links in this context. Thus, we should show them normally. However, handling of those could be improved a lot. I'll amend some basic cases.

rainemak: My proposal:

mailto:

1. Write an email (anything better?)

2. Share

3. Copy to clipboard

sms:

1. Send a message

2. Share

3. Copy to clipboard

tel:

1. Call

2. Share

3. Copy to clipboard

geo:

1. Open

2. Share

3. Copy to clipboard

=> Conceptually all these four would have same context menus but open text would be context sensitive. I'd not start customizing "Copy to clipboard" behaviour as there can be also a body in query arguments. This would already be better than Google Chrome has they are showing for tel/sms/geo/mailto

- "Open in new tab"

- "Open in incognito tab"

- "Copy link address"

- "Copy link text"

The mailto protocol does an exception => that has "Copy email address" that tries to extract email address but fails on CC and BCC only cases (TO is not given).

rainemak:
Added two commits on top.

rainemak:
Updated

jpetrell:
Works like a charm. Code looks good. LGTM.

PR #461, opened 2016-02-19
[sailfish-browser] Do not add myapp to the transfer ui. Fixes JB#34283

rainemak:
myapp is application metadata file that has custom handling. Thus, it should not be added to the transfer ui.

### DIFF
--- a/src/downloadmanager.cpp
+++ b/src/downloadmanager.cpp
@@ -41,6 +41,11 @@ DownloadManager::DownloadManager()
             this, SLOT(setPreferences()));
     connect(QMozContext::GetInstance(), SIGNAL(recvObserve(const QString, const QVariant)),
             this, SLOT(recvObserve(const QString, const QVariant)));
+
+    // Ignore the download info argument of the downloadStatusChanged signal.
+    connect(this, &DownloadManager::downloadStatusChanged, [=](int downloadId, int status) {
+        m_statusCache.insert(downloadId, static_cast<DownloadStatus::Status>(status));
+    });
 }
 
 DownloadManager::~DownloadManager()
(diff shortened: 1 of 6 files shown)


src/downloadmanager.cpp
```diff
@@ -164,22 +163,55 @@ bool DownloadManager::moveMyAppPackage(QString path)
     return true;
 }
 
+bool DownloadManager::isMyApp(const QString &targetPath) const
+{
+    return QFileInfo(targetPath).completeSuffix() == QLatin1Literal("myapp");
+}
+
+bool DownloadManager::updatePlatformTransfers(const QString &targetPath) const
```
pvuorela: Guess fine, but thinking if e.g. needPlatformTransfersUpdate() would be clearer. This name sounds a bit like performing an action.

src/downloadmanager.cpp
```diff
@@ -164,22 +163,55 @@ bool DownloadManager::moveMyAppPackage(QString path)
...
+    return !isMyApp(targetPath);
+}
+
+void DownloadManager::finishMyAppDownload(const QString &targetPath) const
+{
+    Q_ASSERT(isMyApp(targetPath));
+
+    QString rootNameSpace("com.aptoide.partners%1");
+    QString aptoideSupport = rootNameSpace.arg(QLatin1Literal(".AptoideJollaSupport"));
+#if CPU_ARM
```
pvuorela: Code just moved around, but assumptions like this might bite back later.

pvuorela:
Couple cosmetic comments, but in general LGTM.

src/downloadmanager.cpp
rainemak: Most likely this will bite back. Regrdless of that Store tablet is not actually a store for tablet rather a store that provides x86 apks.

src/downloadmanager.cpp
rainemak: Quite funny, had exactly that at some point while doing the change. Will update that back :)

rainemak:
Updated... and merging.

PR #466, opened 2016-04-20
[sailfish-browser] Add site specific overrides. Contributes to JB#34904

rainemak:
Let's override youtube.com here so that site specific override gets triggered. Noticed that navigating from other site that was overridden to youtube.com there was just a network request. Thus, previous mapped user-agent override got picked. Will cleanup Embedlite Components side more once we get this one in. 

Furthermore, I'll add means to configure browser specific overrides in easier manner.


rainemak:
LGTMing ( @pvuorela )

PR #468, opened 2016-04-27
Move overrides to json file

rainemak:


oneshot.d/browser-update-default-data
```diff
@@ -0,0 +1,28 @@
...
+    mkdir -p $MOZEMBED_DIR
+fi
+
+if [ ! -f "$MOZEMBED_DIR/ua-update.json" ]; then
+    while IFS='' read -r line || [[ -n "$line" ]]; do
+        # Skip comment lines
+        if [[ $line !=  *\/\/* ]] ; then
+            echo "$line" >> "$MOZEMBED_DIR/ua-update.json"
+        fi
+    done < "$BROWSER_DATA_DIR/ua-update.json.in"
```
pvuorela: I'd suspect comparing with pattern is bash specific feature. Also don't think this allows comments at end of lines.

If above functionality is enough, wouldn't it be just simpler with
 grep -v // $BROWSER_DATA_DIR/ua-update.json.in > $MOZEMBED_DIR/ua-update.json

pvuorela:
One comment, but generally LGTM.

oneshot.d/browser-update-default-data
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+MOZEMBED_DIR=/home/$USER/.mozilla/mozembed
+BROWSER_DATA_DIR=/usr/share/sailfish-browser/data
+
+if [ ! -d "$MOZEMBED_DIR" ]; then
+    mkdir -p $MOZEMBED_DIR
+fi
+
+if [ ! -f "$MOZEMBED_DIR/ua-update.json" ]; then
```
siteshwar: How will you update ```$MOZEMBED_DIR/ua-update.json``` file if ```ua-update.json.in``` file is updated with a new release ?

rainemak: That's simple. I'll enable ```general.useragent.updates.url``` preference  (+interval & retry timeouts) [1][2]. We haven't had yet time to deployment json to the cdn. I've verified the approach with one of server. The UserAgentUpdates.jsm will cache changes to the /home/nemo/.mozilla/mozembed/ua-update.json => that's why we only need to take care of the initial flashing of the device.

Sorry that I cut some corners in the PR description.

[1] Similar to:
https://git.merproject.org/mer-core/gecko-dev/blob/nemo_embedlite/b2g/app/b2g.js#L871

[2]
https://git.merproject.org/mer-core/gecko-dev/blob/nemo_embedlite/netwerk/protocol/http/UserAgentUpdates.jsm#L45

siteshwar: Sounds good.

rainemak: It also means that it's easier for community to test different sites and for us easier way to apply fixes.

Personally I'd like to see the next level where you guys (community) would start providing better UAs for you favourite web sites. Enablers starts to be in place :). A topic that we comment for the community a meeting? Do it together still holds. I just need to describe how to test user-agents and tell where we are going to. My thinking is that we should flip to super simple UA like B2G has. I feel that this would be beneficial to both sides.

oneshot.d/browser-update-default-data
rainemak: Comments at the end of the lines are not allowed.

Yes, grep works just fine. As it is a bit simpler will change to that.

siteshwar:
LGTM

PR #469, opened 2016-04-29
[sailfish-browser] Set application.process.binary variable for pulse audio. Contributes to JB#34695

rainemak:
As we are not Gtk+ application and not using directly GLib
set PULSE_PROP_application.process.binary environment variable.

There must not be garbage in the application.process.binary value.


jusa:
LGTM

PR #471, opened 2016-05-02
Move browser related code from main to Browser class

rainemak:
Moving code around. I was doing this in different context some days back. Didn't want this one to grow bigger than this.

Should be rather easy to review with gitk or similar that shows file move diff correctly.


src/browser/browser_p.h
```diff
@@ -0,0 +1,27 @@
...
+****************************************************************************/
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef BROWSER_PRIVATE_H
+#define BROWSER_PRIVATE_H
+
+#include "closeeventfilter.h"
```
siteshwar: Just declaring ```class CloseEventFilter``` should be enough here.

siteshwar:
LGTM

src/browser/browser_p.h
rainemak: Yes, on the other hand this is private class. So, doesn't matter that much.

rainemak: Updated.

PR #473, opened 2016-05-06
[sailfish-browser] Reset MozWindow upon close event if all tabs have been closed. Fixes JB#33457

rainemak:
Exit was not correctly handle on Gecko 38 if all tabs were closed (last view)
before closing the browser from the switcher.

@siteshwar review ?

### DIFF
--- a/src/core/declarativewebcontainer.cpp
+++ b/src/core/declarativewebcontainer.cpp
@@ -519,6 +519,11 @@ bool DeclarativeWebContainer::eventFilter(QObject *obj, QEvent *event)
     if (event->type() == QEvent::Close && m_mozWindow) {
         m_mozWindow->suspendRendering();
         m_closing = true;
+
+        if (m_webPages->count() == 0) {
+            m_mozWindow.reset();
+        }
+
         m_webPages->clear();
     }
 


src/core/declarativewebcontainer.cpp
```diff
@@ -519,6 +519,11 @@ bool DeclarativeWebContainer::eventFilter(QObject *obj, QEvent *event)
     if (event->type() == QEvent::Close && m_mozWindow) {
         m_mozWindow->suspendRendering();
         m_closing = true;
+
+        if (m_webPages->count() == 0) {
+            m_mozWindow.reset();
```
siteshwar: Is ```DeclarativeWebContainer::onLastViewDestroyed``` slot still used or we can remove it ?

rainemak: Yes it is in use. Called when the last tab is closed.

I think that closeeventfilter should be merged to the webcobtainer/browser but didn't want to do that in this MR. All filtering is done now in webcontainer (only connections based op are left in closeeventfilter). Thus, "exit browser when downloading is finished" could be done differently. Case being when browser is "closed" before downloading is finished and "restarted" again when the same downloading is still ongoing. So, we also need to alter how the browser content window surface is recreated incase browser is closed while downloading is ongoing and restarted again.

siteshwar:
LGTM

PR #474, opened 2016-05-11
[user-agent] Add mobile friendlier UA for Yandex. Contributes to JB#25096

rainemak:
Search result page doesn't return optimal web page. There could
something wrong elsewhere that affects layout maximum width. As
Android Firefox UA agent also reproduced wrong search result page.

Image search works better.


jpetrell:
LGTM

Issue #124, opened 2014-05-05
JavaScript-heavy (asm.js) WebGL page crashes browser


thp:
Steps to reproduce:

* Open http://thp.io/2014/petals/play/ in the browser
* Wait for the page to load

Expected result:

* Page loads and works, content is rendered via WebGL and has acceptable performance

Actual result:

* Browser locks up, then crashes

Other comments:

* The page works on Nexus 4 in Android (tested using Chrome 34 and Firefox 29)
* The page doesn't work on Firefox 29 running in Alien on Jolla (also crashes)
* The same code, running natively, works on SFOS on Jolla, so all the GL features it uses (FBOs with depth buffer, ...) should be supported by the GL driver

siteshwar:
These are top frames of backtrace of the crash :

```
(gdb) bt
#0  0xb07fc3d4 in nsXPCWrappedJS::AddRef (this=0xa69be8c0)
    at /usr/src/debug/xulrunner-qt5-38.0.5.3+nemo_embedlite_38.20151109122853.9214377/js/xpconnect/src/XPCWrappedJS.cpp:228

#1  0xb05a61ce in nsXPTCStubBase::QueryInterface (this=0xa6999b70, aIID=..., aInstancePtr=0xa817c544)
    at /usr/src/debug/xulrunner-qt5-38.0.5.3+nemo_embedlite_38.20151109122853.9214377/xpcom/reflect/xptcall/xptcall.cpp:20

#2  0xb059dfea in nsComponentManagerImpl::GetServiceByContractID (this=0xabec4080, aContractID=0xb1866a18 "@mozilla.org/xre/app-info;1", aIID=..., aResult=0xa817c544)
    at /usr/src/debug/xulrunner-qt5-38.0.5.3+nemo_embedlite_38.20151109122853.9214377/xpcom/components/nsComponentManager.cpp:1544

#3  0xb05ac8d2 in nsGetServiceByContractID::operator() (this=this@entry=0xa817c53c, aIID=..., aInstancePtr=aInstancePtr@entry=0xa817c544)
    at /usr/src/debug/xulrunner-qt5-38.0.5.3+nemo_embedlite_38.20151109122853.9214377/xpcom/glue/nsComponentManagerUtils.cpp:280

#4  0xb05ac8ee in nsCOMPtr_base::assign_from_gs_contractid (this=0xa817c550, aGS=..., aIID=...)
[...]

PR #477, opened 2016-05-12
[sailfish-browser] Enable asmjs. Fixes JB#22349

rainemak:
See also https://github.com/sailfishos/sailfish-browser/issues/124 and
https://git.merproject.org/mer-core/embedlite-components/merge_requests/13

Using the correct AppInfo also allows use to enable asmjs. Likely
fixed by processId, processType, or/and XPCOMABI property fixes.


chriadam:
LGTM

PR #479, opened 2016-05-20
Jb35163

rainemak:
PR contains two small cleanup commits.

Throttle always when browser content window is invisible. Throttling
is disabled on switcher if there is video playing. Browser
content window is invisible when display is turned off.

This also resets suspend intention after timer has triggered.


rainemak:
See also https://git.merproject.org/mer-core/qtmozembed/merge_requests/7

chriadam:
LGTM

PR #481, opened 2016-06-01
[sailfish-browser] Activate tabs only after browser initialized. Fixes JB#35387

rainemak:
As external urls are loaded always to new tab, we need to
wait that browser is initialized (context, mozwindow, and model are ready).
After browser initialized, we can proceed initialization steps normally.

This also fixes overwriting the existing active tab with the external
url when browser is started.

### DIFF
--- a/src/core/declarativewebcontainer.cpp
+++ b/src/core/declarativewebcontainer.cpp
@@ -196,8 +196,10 @@ void DeclarativeWebContainer::setTabModel(DeclarativeTabModel *model)
             emit m_model->countChanged();
         }
 
-        // Set waiting for a tab.
-        if (m_model) {
+        // Set waiting for a tab. This can only happen when
+        // browser is already started. Initialization steps take
+        // care creating new tabs as per need.
+        if (m_model && m_initialized) {
             m_model->setWaitingForNewTab(true);
         }
     }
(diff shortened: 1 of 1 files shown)


rainemak:
@siteshwar @chriadam would you like to review?

src/core/declarativewebcontainer.cpp
```diff
@@ -743,11 +746,14 @@ void DeclarativeWebContainer::initialize()
         return;
     }
 
+    // From this point onwards, we're ready to initialize.
```
siteshwar: nitpick: Are we ready to initialize or we are initialized ? :)

rainemak: We're not yet initialized as init is finalized within if-else if block right after this. As we might end up calling ```loadTab``` and it is guarded by ```m_initialized``` we need to set it before.

chriadam: Maybe add that as a second-sentence to the comment, something like: "We set m_initialized to true prior to the block below since we may need to call loadTab() within it, and that function is guarded by the value of m_initialized."

rainemak: Done

chriadam:
LGTM

PR #483, opened 2016-06-08
[sailfish-browser] Mark tab as rendered after one frame is rendered. Fixes JB#35310

rainemak:
The firstPaint is is not emitted correctly on web pages like ruutu.fi or dailymotion.com when those are opened to a new tab. The content dimmer does not open as first paint doesn't reach this level. What is in common with these web sites is that both contains video elements. Not sure if that play a role here but it could that e.g. that the sub tree handles the first paint (https://git.merproject.org/mer-core/gecko-dev/blob/nemo_embedlite_38/layout/base/nsPresContext.cpp#L2451). What I can see is that first paint does not reach WebBrowserChrome (https://git.merproject.org/mer-core/gecko-dev/blob/nemo_embedlite_38/embedding/embedlite/utils/WebBrowserChrome.cpp#L405).

Regardless of above, let's use QMozWindow::compositingFinished signal instead as that is coming directly from LayerManagerComposite::Render().

Piggybacked a small signaling fix when last tab is closed.

### DIFF
--- a/src/core/declarativewebcontainer.cpp
+++ b/src/core/declarativewebcontainer.cpp
@@ -125,9 +125,9 @@ QMozWindow *DeclarativeWebContainer::mozWindow() const
     return m_mozWindow.data();
 }
 
-void DeclarativeWebContainer::setWebPage(DeclarativeWebPage *webPage)
+void DeclarativeWebContainer::setWebPage(DeclarativeWebPage *webPage, bool triggerSignals)
 {
-    if (m_webPage != webPage) {
+    if (m_webPage != webPage || triggerSignals) {
         // Disconnect previous page.
         if (m_webPage) {
             m_webPage->disconnect(this);
@@ -819,7 +819,7 @@ void DeclarativeWebContainer::releasePage(int tabId)
         m_webPages->release(tabId);
         // Successfully destroyed. Emit relevant property changes.
         if (m_model->count() == 0) {
-            setWebPage(NULL);
+            setWebPage(NULL, true);
             postClearWindowSurfaceTask();
         }
     }
--- a/src/core/declarativewebcontainer.cpp
+++ b/src/core/declarativewebcontainer.cpp
@@ -134,6 +134,8 @@ void DeclarativeWebContainer::setWebPage(DeclarativeWebPage *webPage)
         }
 
         m_webPage = webPage;
+        // Mark as not rendered when ever tab is changed.
+        setActiveTabRendered(false);
 
         if (m_webPage) {
             connect(m_webPage, SIGNAL(canGoBackChanged()), this, SIGNAL(canGoBackChanged()), Qt::UniqueConnection);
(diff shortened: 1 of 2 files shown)


rainemak:
@siteshwar @chriadam @pvuorela would you like to review? thanks in advance.

src/core/declarativewebcontainer.cpp
```diff
@@ -851,17 +852,17 @@ void DeclarativeWebContainer::updateLoading()
...
 }
 
-void DeclarativeWebContainer::onFirstPaint(int, int)
-{
-    updateActiveTabRendered();
-}
-
 void DeclarativeWebContainer::updateActiveTabRendered()
 {
+    if (m_webPage && !m_webPage->completed()) {
```
chriadam: If we get the compositingFinished() signal before m_webPage->completed() is true, does this mean that we will never setActiveTabRendered(true), or does the compositingFinished() signal get emitted once per frame?  I assume the latter, but wanted to double check.

src/core/declarativewebcontainer.cpp
```diff
@@ -851,17 +852,17 @@ void DeclarativeWebContainer::updateLoading()
...
+        return;
+    }
+
     setActiveTabRendered(true);
-    // One frame rendered so let's disconnect.
-    disconnect(m_mozWindow.data(), &QMozWindow::compositingFinished,
-               this, &DeclarativeWebContainer::updateActiveTabRendered);
+    // One frame rendered so let's disconnect. Either of these reported.
+    // We're still suffering from bug
```
chriadam: this comment is a bit unclear

rainemak: The last sentence is a left over. Will remove, thanks.

src/core/declarativewebcontainer.cpp
rainemak: Gecko cannot render the active web page before we set the page as active [1]. In other words, during time period between creation start and finish the newly created web page cannot render. Browser takes care of tracking active web page.

Rendering sequence [2] goes as follows, in case you're interested:

When widget is valid (our EmbedLitePuppetWidget that is hosting window)
- First let window to draw underlay (e.i. we're doing custom clearing)
- Do rendering
- Draw custom overlay
- Do post rendering steps (ends up emitting compositingFinished)

The EmbedLitePupperWidget (nsIWidget) implements above methods [3] and forwards calls to EmbedLiteWindowListener which in turn is implemented by QMozWindow private. All of this is done in gecko compositor thread. Due to that in this case we don't use direct connection but unique connection (takes case switching tabs) so that all is done on the main thread.

[1]
https://git.merproject.org/mer-core/qtmozembed/blob/nemo_embedlite_38/src/qopenglwebpage.cpp#L114

[2]
https://git.merproject.org/mer-core/gecko-dev/blob/nemo_embedlite_38/gfx/layers/composite/LayerManagerComposite.cpp#L724

[3]
https://git.merproject.org/mer-core/gecko-dev/blob/nemo_embedlite_38/embedding/embedlite/embedshared/EmbedLitePuppetWidget.cpp#L755

pvuorela:
Not really familiar with code here either, but LGTM from what I gather.

PR #485, opened 2016-06-15
[sailfish-browser] Mark tab as renderer also when dom content is loaded. Fixes JB#35310

rainemak:
Embedlite-components reports domContentLoaded, telling also
if it was root frame. In practice this means that from this
point onwards there is nothing blocking browser to render.

Some references:
https://developer.mozilla.org/en-US/docs/Web/Events/DOMContentLoaded

### DIFF
--- a/src/qtmozembed/declarativewebpage.h
+++ b/src/qtmozembed/declarativewebpage.h
@@ -31,6 +31,7 @@ class DeclarativeWebPage : public QOpenGLWebPage {
     Q_PROPERTY(bool userHasDraggedWhileLoading MEMBER m_userHasDraggedWhileLoading NOTIFY userHasDraggedWhileLoadingChanged FINAL)
     Q_PROPERTY(bool fullscreen READ fullscreen NOTIFY fullscreenChanged FINAL)
     Q_PROPERTY(bool forcedChrome READ forcedChrome NOTIFY forcedChromeChanged FINAL)
+    Q_PROPERTY(bool domContentLoaded READ domContentLoaded NOTIFY domContentLoadedChanged FINAL)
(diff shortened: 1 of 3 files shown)


src/qtmozembed/declarativewebpage.h
```diff
@@ -31,6 +31,8 @@ class DeclarativeWebPage : public QOpenGLWebPage {
     Q_PROPERTY(bool userHasDraggedWhileLoading MEMBER m_userHasDraggedWhileLoading NOTIFY userHasDraggedWhileLoadingChanged FINAL)
     Q_PROPERTY(bool fullscreen READ fullscreen NOTIFY fullscreenChanged FINAL)
     Q_PROPERTY(bool forcedChrome READ forcedChrome NOTIFY forcedChromeChanged FINAL)
+    Q_PROPERTY(bool domContentLoaded READ domContentLoaded NOTIFY domContentLoadedChanged FINAL)
```
rainemak: This is a bit unrelated but handy addition when you need debug things. You can do debugging quicker from QML side.

rainemak:
@siteshwar , @pvuorela , @chriadam would you like to review?

src/qtmozembed/declarativewebpage.h
```diff
@@ -31,6 +31,8 @@ class DeclarativeWebPage : public QOpenGLWebPage {
...
+
```
pvuorela: Empty line intentional?

pvuorela:
LGTM

siteshwar:
LGTM

src/qtmozembed/declarativewebpage.h
rainemak: Nope. Removed.

PR #487, opened 2016-06-21
Jb32815

rainemak:
When browser is launched for the first time, the browser engine does not
know about the search engines. Once the browser engine is ready,
we request the engine to load the default configured search engines.

This commit changes browser initialization to listen "search-engine-added"
message. Once all initially configured engines are added, the default
search engine is set.

Browser does not handle yet a case where all search engines would
be changed on the fly but that is another issue.

See also:
https://git.merproject.org/mer-core/embedlite-components/merge_requests/16

As I touched unit tests, fixed them to pass as well.

### DIFF
--- a/src/core/settingmanager.cpp
+++ b/src/core/settingmanager.cpp
@@ -11,6 +11,7 @@
 
 #include "settingmanager.h"
 #include "dbmanager.h"
+#include "opensearchconfigs.h"
 
 #include <MGConfItem>
 #include <qmozcontext.h>
@@ -21,6 +22,8 @@ static SettingManager *gSingleton = 0;
 SettingManager::SettingManager(QObject *parent)
     : QObject(parent)
     , m_initialized(false)
+    , m_searchEnginesInitialized(false)
+    , m_addedSearchEngines(0)
 {
     m_clearHistoryConfItem = new MGConfItem("/apps/sailfish-browser/actions/clear_history", this);
     m_clearCookiesConfItem = new MGConfItem("/apps/sailfish-browser/actions/clear_cookies", this);
@@ -35,6 +38,8 @@ SettingManager::SettingManager(QObject *parent)
     m_toolbarLarge = new MGConfItem("/apps/sailfish-browser/settings/toolbar_large", this);
     connect(m_toolbarSmall, SIGNAL(valueChanged()), this, SIGNAL(toolbarSmallChanged()));
     connect(m_toolbarLarge, SIGNAL(valueChanged()), this, SIGNAL(toolbarLargeChanged()));
+    connect(QMozContext::GetInstance(), SIGNAL(recvObserve(QString, QVariant)),
+            this, SLOT(handleObserve(QString, QVariant)));
 }
 
 bool SettingManager::clearHistoryRequested() const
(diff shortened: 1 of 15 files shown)


rainemak:
@jpetrell review?

src/core/settingmanager.cpp
```diff
@@ -143,6 +157,74 @@ void SettingManager::doNotTrack()
...
+            const QStringList configuredEngines = configs.keys();
+            QStringList registeredSearches(dataMap.value(QLatin1String("engines")).toStringList());
+            QString defaultSearchEngine = dataMap.value(QLatin1String("defaultEngine")).toString();
+
+            m_searchEnginesInitialized = !registeredSearches.isEmpty();
+
+            // Upon first start, engine doesn't know about the search engines.
+            // Engine load requests are send within the for loop below.
+            if (!m_searchEnginesInitialized) {
+                m_addedSearchEngines = new QStringList(configuredEngines);
```
siteshwar: Could there be a memory leak if search engine initialization fails while handling ```search-engine-added``` message ?  ```m_addedSearchEngines``` is deleted only if search engine initialization is succesful.

rainemak: Yes there could but imo this is still better than having QStringList allocated to the stack.

As we're interested only about the first time when search engines are added (registeredSearches is empty) => ERROR_DUPLICATE_ENGINE (0x2) should not occur. 

As ```confirm``` argument is always false when adding search engines that rules out one ERROR_UNKNOWN_FAILURE. Didn't manage to create failure with local files. However, with something like ```http://invalid/data/engine.xml``` an error will be triggered.

rainemak: Error will be shown to user as well. So, that kind of error should be caught while developing.

src/core/settingmanager.cpp
```diff
@@ -133,8 +138,17 @@ bool SettingManager::clearCache()
...
 void SettingManager::setSearchEngine()
 {
-    QVariant searchEngine = m_searchEngineConfItem->value(QVariant(QString("Google")));
-    QMozContext::GetInstance()->setPref(QString("browser.search.defaultenginename"), searchEngine);
+    if (m_searchEnginesInitialized) {
+        QVariant searchEngine = m_searchEngineConfItem->value(QVariant(QString("Google")));
+        QMozContext *context = QMozContext::GetInstance();
+        context->setPref(QString("browser.search.defaultenginename"), searchEngine);
+
+        // Let's the nsSearchService to update the search engine (through EmbedLiteSearchEngine).
```
jpetrell: Let's the nsSearchService to update -> Let nsSearchService update

rainemak:
Fixed review findings and added a warning when adding of a search engine fails.

jpetrell:
LGTM

src/core/settingmanager.cpp
siteshwar: You can delete ```m_addedSearchEngines``` if it's not NULL before this line.

rainemak: There will be only one ```init``` message per browser startup

siteshwar:
LGTM

PR #488, opened 2016-06-23
[sailfish-browser] Remove default search engine guard. Fixes JB#32815

rainemak:
It's enough to check that search engines has been initialized.

@pvuorela / @jpetrell quick one to review ?

### DIFF
--- a/src/core/settingmanager.cpp
+++ b/src/core/settingmanager.cpp
@@ -199,11 +199,9 @@ void SettingManager::handleObserve(const QString &message, const QVariant &data)
                 mozContext->sendObserve(QLatin1String("embedui:search"), QVariant(removeMsg));
             }
 
-            // After first start we can update the default search engine immediately
-            // without
-            if (m_searchEnginesInitialized && defaultSearchEngine.isEmpty()) {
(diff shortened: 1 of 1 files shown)


src/core/settingmanager.cpp
```diff
@@ -199,9 +199,8 @@ void SettingManager::handleObserve(const QString &message, const QVariant &data)
                 mozContext->sendObserve(QLatin1String("embedui:search"), QVariant(removeMsg));
             }
 
-            // After first start we can update the default search engine immediately
-            // without
-            if (m_searchEnginesInitialized && defaultSearchEngine.isEmpty()) {
+            // After first start we can update the default search engine immediately.
+            if (m_searchEnginesInitialized) {
                 setSearchEngine();
```
siteshwar: ```setSearchEngine()``` method checks if ```m_searchEnginesInitialized``` is set. Probably this check can be removed altogether.

rainemak: As, true.

rainemak:
@siteshwar updated.

siteshwar:
LGTM

PR #489, opened 2016-06-29
[sailfish-browser] Open orientation fader aften pending orientation is effective orientation. Fixes JB#35447

rainemak:
See also:
https://git.merproject.org/mer-core/embedlite-components/merge_requests/17
https://git.merproject.org/mer-core/qtmozembed/merge_requests/9

### DIFF
--- a/src/core/declarativewebcontainer.cpp
+++ b/src/core/declarativewebcontainer.cpp
@@ -146,7 +146,7 @@ void DeclarativeWebContainer::setWebPage(DeclarativeWebPage *webPage, bool trigg
             connect(m_webPage, SIGNAL(loadingChanged()), this, SLOT(updateLoading()), Qt::UniqueConnection);
             connect(m_webPage, SIGNAL(loadProgressChanged()), this, SLOT(updateLoadProgress()), Qt::UniqueConnection);
             connect(m_webPage, SIGNAL(contentOrientationChanged(Qt::ScreenOrientation)),
-                    this, SIGNAL(webContentOrientationChanged(Qt::ScreenOrientation)), Qt::UniqueConnection);
+                    this, SLOT(handleContentOrientationChanged(Qt::ScreenOrientation)), Qt::UniqueConnection);
 
             // NB: these signals are not disconnected upon setting current m_webPage.
             connect(m_webPage, SIGNAL(urlChanged()), m_model, SLOT(onUrlChanged()), Qt::UniqueConnection);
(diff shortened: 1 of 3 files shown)


rainemak:
Review @jpetrell @chriadam @siteshwar ?

src/core/declarativewebcontainer.cpp
```diff
@@ -962,3 +973,10 @@ void DeclarativeWebContainer::drawUnderlay()
         funcs->glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
     }
 }
+
+void DeclarativeWebContainer::handleContentOrientationChanged(Qt::ScreenOrientation orientation)
+{
+    if (orientation == pendingWebContentOrientation()) {
```
siteshwar: I am not sure what is the motivation for this method. I see below code in ``` void QMozWindowPrivate::timerEvent(QTimerEvent *event)``` method:

```
            if (mOrientation != mPendingOrientation) {
                 mWindow->SetContentOrientation(QtToMozillaRotation(mPendingOrientation));
                 mOrientation = mPendingOrientation;
             } else {
                 q.orientationChangeFiltered(mOrientation);
             }
```

It means this event should be ignored when pending orientation is same as current orientation.

rainemak: Pending orientation is orientation that we're waiting for. The handleContentOrientationChanged is connected to two different signals: 

* DeclarativeWebPage::contentOrientationChanged(Qt::ScreenOrientation) => comes from OrientationChangeHandler.jsm of embedlite-components

* QMozWindow::orientationChangeFiltered(Qt::ScreenOrientation)

Former comes after engine has applied new orientation for the content (content also resized). 

With the latter one (filtered) there is a catch. Think of sequence "portrait" -> "landscape" -> "portrait" that is executed quickly, nothing should change, right. We do try to apply "portrait" orientation on top of "portrait" orientation. As the landscape orientation was filter out nsScreen [1] of Gecko considers that we never did orientation change -- logically this is correct. Thus, OrientationChangeHandler [2] won't get notified that orientation actually changed as it in reality it did not change.

Now comes the problem, we have a fader on top of content when are performing orientation change so that user won't see the reflow etc flickering (BrowserPage.qml) which needs to be opened after orientation has changed. So now back to "portrait" -> "landscape" -> "portrait" sequence. In the slot ```handleContentOrientationChanged``` we're always waiting that reported orientation matches to the pending one => so when the orientation filtering timer triggers for the last "portrait" orientation nothing happens and we inform that it was filtered but in practice it is the current orientation.

So maybe ```orientationChangeFiltered``` is the misleading one here. Maybe it should be called ```forceContentOrientation``` or similar to emphasize that this is actually the current content orientation.

Does my explanation make sense to you?

[1]
https://git.merproject.org/mer-core/gecko-dev/blob/nemo_embedlite_38/dom/base/nsScreen.cpp#L149

[2]
https://git.merproject.org/mer-core/embedlite-components/blob/nemo_embedlite_38/jsscripts/OrientationChangeHandler.jsm

siteshwar: Yes, thanks for the explanation. Feel free to adjust the signal name.

siteshwar:
LGTM

PR #490, opened 2016-06-29
Jb35379

rainemak:
Contains three commits:

- Switch to use listen "chrome:contentloaded" messages (same as domcontentloaded)

- Clean surface when we're about to initialize a new page

- Do not allow compitor to render before content is loaded.

### DIFF
--- a/src/core/declarativewebcontainer.cpp
+++ b/src/core/declarativewebcontainer.cpp
@@ -707,6 +707,11 @@ void DeclarativeWebContainer::updateContentOrientation(Qt::ScreenOrientation ori
     reportContentOrientationChange(orientation);
 }
 
+void DeclarativeWebContainer::clearSurface()
+{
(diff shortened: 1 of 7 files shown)


rainemak:
Requires: https://git.merproject.org/mer-core/embedlite-components/merge_requests/17

Review @jpetrell @chriadam @siteshwar ?

Prerender calls are called from the compositor thread:

1) https://git.merproject.org/mer-core/gecko-dev/tree/nemo_embedlite_38/embedding/embedlite/embedshared/EmbedLitePuppetWidget.cpp (EmbedLitePuppetWidget::PreRender)
2) https://git.merproject.org/mer-core/qtmozembed/tree/nemo_embedlite_38/src/qmozwindow_p.cpp (QMozWindowPrivate::PreRender())

Furthermore, surface is cleared just before we create a new web page.

siteshwar:
LGTM

PR #492, opened 2016-06-30
[sailfish-browser] Mark contet loaded when page is loaded. Contributes to JB#35681

rainemak:
For instance when loading images directly to a new tab, chrome does
not get domContentLoaded message from the engine.


siteshwar:
LGTM

src/qtmozembed/declarativewebpage.cpp
```diff
@@ -100,6 +100,9 @@ DeclarativeWebPage::DeclarativeWebPage(QObject *parent)
     connect(this, &QOpenGLWebPage::loadedChanged, [this]() {
         if (loaded()) {
             updateViewMargins();
+            // E.g. when loading images directly we don't necessarily get domContentLoaded message from engine.
+            // So mark content loaded when webpage is loaded.
```
rainemak: Amended this comment a bit => previous one didn't make much sense.

PR #497, opened 2016-08-09
[sailfish-browser] Do not wait for new tab when active tab is changed. Fixes JB#35786

rainemak:
When we are about to create a new tab we wait for new tab. However,
when tab is switched or new tab is created it is not needed to
wait for new tab.

This caused a glitch that occasionally that loaded website
was still on the waiting mode (blank view) even thou it was
already loaded.


rainemak:
@chriadam @siteshwar review ?

siteshwar:
LGTM

PR #504, opened 2016-08-29
[sailfish-browser] Fix opacity of url text field. Fixes JB#36093

rainemak:
Since Qt5.4 QQuickItem::opacity has been clamped between 0.0 - 1.0.
Thus, use an intermediate property for cross fading.

### DIFF
--- a/src/pages/components/Overlay.qml
+++ b/src/pages/components/Overlay.qml
@@ -225,10 +225,13 @@ Background {
             Browser.ToolBar {
                 id: toolBar
 
+                property real crossfadeRatio: (overlay.y - webView.fullscreenHeight/2)  / (webView.fullscreenHeight/2 - toolBar.height)
+
                 url: webView.contentItem && webView.contentItem.url || ""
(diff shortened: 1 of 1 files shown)


rainemak:
@martinjones @jpetrell review ?

jpetrell:
LGTM. I would perhaps rename the intermediate value to something else than scaledOpacity, it is not opacity, more like a relative distance to some vertical border.

rainemak:
Ah true, now that I think ```scaledOpacity``` is bad. Maybe something like ```relativeDistanceToOverlayCenter``` but even that might be difficult to understand. The overlayCenter part should indicate that it's referring to the center position when overlay is open => both overlay and toolbar are moving when dragging overlay.

Hmm... It is actually more like ```relativeDistanceToCrossFadePosition```. Nevertheless, will amend.

chriadam:
crossfadeRatio perhaps?

rainemak:
Updated to crossfadeRatio... that sounds good.

Issue #506, opened 2016-08-31
White cover for the tab view


fsausset:
When displayng the tabs inside the browser and going back to home screen, the cover displayed is the last opened tab. After a while (hard to isolate causes) it becomes white...
On jolla C with sailfish 2.0.2.51

fsausset:
I managed to reproduce it: it happens if I close the last opened tab before going back to home screen.

rainemak:
@fsausset thanks for reporting. Managed to reproduce.

rainemak:
Fix merged and scheduled for 2.0.4. @fsausset again thanks for reporting.

fsausset:
Thanks for the quick fix!

PR #507, opened 2016-09-01
Soft load new active tab when a tab is removed

rainemak:
@pvuorela could you review.

There are two commits.

1) Contains actual fix
2) Cleans up code that became deadcode after the fix

See also: https://github.com/sailfishos/sailfish-browser/issues/506


pvuorela:
LGTM

PR #519, opened 2016-10-05
[sailfish-browser] Fix browser profile path. Fixes JB#36500

rainemak:
We accidentally changed profile path when updating browser
to use common WebEngine. Thus, previously saved passwords,
caches, preferences got lost.


rainemak:
@chriadam / @pvuorela / @siteshwar review ?

See also : https://git.merproject.org/mer-core/gecko-dev/blob/nemo_embedlite_38/embedding/embedlite/EmbedLiteApp.cpp#L61

chriadam:
LGTM

PR #520, opened 2016-10-05
[sailfish-browser] Cleanup deprecated mozcontext calls. Fixes JB#36499

rainemak:
More or less find & replacing deprecated method calls. This also fixes one signal name for QMozContext mock (2nd commit), fixes WebUtils unit tests (3rd commit), and fixes WebPage unit tests  (4th commit).

There are still references to QMozContext but at least logs are no longer full of deprecated API calls.

### DIFF
--- a/src/browser/declarativewebutils.cpp
+++ b/src/browser/declarativewebutils.cpp
@@ -32,7 +32,6 @@
 
 #include <math.h>
 #include "declarativewebutils.h"
-#include "qmozcontext.h"
 #include "browserpaths.h"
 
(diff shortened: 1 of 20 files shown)


chriadam:
One minor comment, but LGTM.

src/browser/declarativewebutils.cpp
```diff
@@ -32,7 +32,6 @@
 
 #include <math.h>
 #include "declarativewebutils.h"
-#include "qmozcontext.h"
```
chriadam: gah, github removes angle brackets.  #include webenginesettings.h or #include webengine.h, is what i meant.

PR #524, opened 2016-10-20
[sailfish-browser] Suspend / resume rendering when ready to paint changes. Contributes to JB#35379

rainemak:
This pauses compositor when we are not ready to render making
LayerManagerComposite to return early from
LayerManagerComposite::BeginTransaction.

Occasionally we can see "chessboard" like garbage textures when you start loading web pages. Not sure if all error are caused due to this. Nevertheless, we should pause the gecko compositor.

### DIFF
--- a/src/core/declarativewebcontainer.cpp
+++ b/src/core/declarativewebcontainer.cpp
@@ -351,6 +351,12 @@ bool DeclarativeWebContainer::readyToPaint() const
 void DeclarativeWebContainer::setReadyToPaint(bool ready)
 {
     if (m_mozWindow && m_mozWindow->setReadyToPaint(ready)) {
+        if (ready) {
+            m_mozWindow->resumeRendering();
+        } else {
+            m_mozWindow->suspendRendering();
+        }
+
         emit readyToPaintChanged();
     }
 }


rainemak:
@pvuorela @chriadam review ?

chriadam:
From what I can see, the readyToPaint property is only set in a single binding:

     readyToPaint: resourceController.videoActive ? webView.visible && !resourceController.displayOff : webView.visible && webView.contentItem && webView.contentItem.domContentLoaded

I'm just wondering if it could ever be the case that this binding could be true but we still want rendering to be suspended (e.g., when minimised and another app is in foreground, or...?)

rainemak:
In case browser is minimized and another app is in foregroung ```readyToPaint``` evaluates to false. This is because binding contains in both branches ```webView.visible```. As webView == QWindow it is visible when ever switcher is visible.

To elaborate a bit more this ```readyToPaint``` guards https://git.merproject.org/mer-core/gecko-dev/blob/nemo_embedlite_38/gfx/layers/composite/LayerManagerComposite.cpp#L688 (```!mCompositor->GetWidget()->PreRender(this)```). The widget is EmbedLitePuppetWidget which in turn asks permission to render from the registered EmbedLiteWindow (QMozWindowPrivate implements window listener). 

Call sequence is like
LayerManagerComposite::EndTransaction -> LayerManagerComposite::Render -> EmbedLitePuppetWidget::PreRender

So readyToPaint guard is at the end of the transaction and by using the same state (readyToPaint) here we bailout earlier in case content is not ready to be rendered (domContentLoaded).

The ```Render``` method

chriadam:
Ok, great, just wanted to check that case.  LGTM.

PR #525, opened 2016-10-20
[sailfish-browser] Do not allow margins updates when browser is in background. Fixes JB#35305

rainemak:
When browser is cover by another application and user peeks
the top most application a bit so that switcher becomes visible also
browser activates and continues rendering. However, this peek
should not update content margins which can happen if
contentHeight changes.

This also moves toolbar / chrome handling to C++ side. Chrome gesture
is now enabled / disabled based forcedChrome property. Chrome is
reset when nagivating from one web page to another.

See also : https://git.merproject.org/mer-core/qtmozembed/merge_requests/27

### DIFF
--- a/src/qtmozembed/declarativewebpage.cpp
+++ b/src/qtmozembed/declarativewebpage.cpp
@@ -105,6 +105,15 @@ DeclarativeWebPage::DeclarativeWebPage(QObject *parent)
             setContentLoaded();
         }
     });
+
+    // When loading start reset state of chrome.
+    connect(this, &QOpenGLWebPage::loadingChanged, [this]() {
+        if (loading()) {
+            forceChrome(false);
+            setChrome(true);
+        }
+    });
+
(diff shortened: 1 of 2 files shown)


rainemak:
@pvuorela  @chriadam review ?

chriadam:
One comment but LGTM

src/qtmozembed/declarativewebpage.cpp
```diff
@@ -370,9 +384,11 @@ void DeclarativeWebPage::updateViewMargins()
         qreal threshold = qMax(m_fullScreenHeight * 1.5f, (m_fullScreenHeight + (m_toolbarHeight*2)));
         if (contentHeight() < threshold) {
             margins.setBottom(m_toolbarHeight);
+            chromeVisible = true;
         }
     }
 
+    forceChrome(chromeVisible);
```
rainemak: forceChrome disables the gesture to hide/show toolbar when panning content. When content height is big enough, we can let the gesture to work other wise it's disabled.

PR #526, opened 2016-10-21
[sailfish-browser] Install "Search" events-view quick-action. Contributes to JB#35172

chriadam:
Previously, the events view quick actions were installed by the
lipstick-jolla-home package.  This commit ensures that the quick
action to search the web is installed by the sailfish-browser,
so that it is only available when sailfish-browser is installed.

Contributes to JB#35172

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -102,6 +102,9 @@ Q_DECL_EXPORT int main(int argc, char *argv[])
     translator.load(QLocale(), "sailfish-browser", "-", translationPath);
     qApp->installTranslator(&translator);
 
+    //% "Search the web"
+    QT_TRID_NOOP("sailfish-browser-quickaction-la-search_the_web");
+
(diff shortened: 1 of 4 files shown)


src/main.cpp
```diff
@@ -102,6 +102,9 @@ Q_DECL_EXPORT int main(int argc, char *argv[])
     translator.load(QLocale(), "sailfish-browser", "-", translationPath);
     qApp->installTranslator(&translator);
 
+    //% "Search the web"
+    QT_TRID_NOOP("sailfish-browser-quickaction-la-search_the_web");
```
rainemak: Wrap with Q_UNUSED so that build won't fail due to unused variable.

rainemak:
Added one comment other than that LGTM

chriadam:
As per discussion in JB#35172, declining this one.

PR #530, opened 2016-10-25
[sailfish-browser] Do not change chrome state before dom content is loaded. Contributes to JB#14628

rainemak:
When we are loading a web page, atYBeginning and/or atYEnd might change
so that content reaches beginning/end while in reality content loading is
about to start. Thus, guard statements with domContentLoaded property.


rainemak:
@chriadam review ?

chriadam:
LGTM

PR #532, opened 2016-11-03
[sailfish-browser] Close vkb when opening context menu. Contributes to JB#36163

rainemak:
Give focus to the other window and hide virtual keyboard.

See also: https://github.com/sailfishos/sailfish-components-webview/pull/18


chriadam:
LGTM

PR #534, opened 2016-11-10
Basic text selection and copy

rainemak:
Selected text is copied to clipboard when selection starts and ends (selection
marker is released).

Selection toolbar has four actions: share, search (with default search engine),
clear, and call. Call icon is only visible when a phone number is selected.

Currently we are still missing correct "call" icon and selection markers need to be updated.


rainemak:
@chriadam / @pvuorela / @jpetrell review ?

Needs also https://git.merproject.org/mer-core/embedlite-components/merge_requests/29

Also new QtMozEmbed is needed. That fixed contentResolution error.  So, if you see that selection markers reposition upon orientation change to wrong place, then you don't have the fix yet. But that should be already built.

jpetrell:
Wow, didn't realise you were this far :)

Some comments:
- Sharing over email includes the copied text twice, once on header and once on body, maybe only add to the body
- Do the selection handles follow some design? The don't look bad, but they could maybe align better with the platform style
- If you drag on the left or right edge of selection area slightly above the handles I would expect the selection to update
- If long-press selects a word could longer press select the whole paragraph?  
- Would be nice if the selection region and selection handles color followed the Theme object, now region seems to be always blue and handles orange
- Couple of times the handles jumped to a surprising place
- "No sharing options" label flickers when sharing page slides in, not caused by your changes but should be trivial to fix so pretty please?
- The performance is not good and not being able to scroll the page are clumsy, but having an option to copy is already miles better than before so I am not complaining :)
- I wouldn't mind seeing clipboard icon on the notification banner, making a variant of icon-s-clipboard that works on banner layout would probably be quite easy

rainemak:
Comments:

*    Sharing over email includes the copied text twice, once on header and once on body, maybe only add to the body
    * @rainemak to @jpetrell : will check
*    Do the selection handles follow some design? The don't look bad, but they could maybe align better with the platform style
    * @rainemak to @jpetrell : those will be updated
*    If you drag on the left or right edge of selection area slightly above the handles I would expect the selection to update
    * @rainemak to @jpetrell : we could implement the swap method that I already added to the TextSelectionController. embedlite-components SelectionPrototype.js sends a swap message when selection markers collide. I didn't see that so big deal at this phase. Will file a separate task about it, ok?
*    If long-press selects a word could longer press select the whole paragraph?
    * @rainemak to @jpetrell : not gonna happen at this phase. Might be relatively easy to do. Will file a task.
*    Would be nice if the selection region and selection handles color followed the Theme object, now region seems to be always blue and handles orange
    * @rainemak to @jpetrell : Handlers will be updated. Selection color should not be big deal to change. Need to what nsTextFrame does with colors but it looks that GetSelectionTextColors returns both foreground and background colors what ever those are. 
*    Couple of times the handles jumped to a surprising place
    * @rainemak to @jpetrell :  that can happen and happens also with Android FF. Noticed that we have now similar jumps as you can see with Android FF.
*    "No sharing options" label flickers when sharing page slides in, not caused by your changes but should be trivial to fix so pretty please?
    * @rainemak to @jpetrell : Will check from where this is coming from
*   The performance is not good and not being able to scroll the page are clumsy, but having an option to copy is already miles better than before so I am not complaining :)
    * @rainemak to @jpetrell : Yeap, performance is not good but not bad either. This is what you get with our architecture where we don't have xul. This means that chrome related operations are implemented with qml maybe we could have a hybrid solution where parts would be implemented in CSS/JS or we could render markers with opengl from gecko's compositor thread on top of the content. I was pondering this again yesterday a bit and it actually should be so that selection works better with QuickItem based WebView which is kind of funny.
*    I wouldn't mind seeing clipboard icon on the notification banner, making a variant of icon-s-clipboard that works on banner layout would probably be quite easy
   * @rainemak to @jpetrell :I have there now browser icon but that's not appearing either. Maybe I'm misusing it or support is just missing from the lipstick-jolla-home. Will give a look.

src/pages/ShareTextPage.qml
```diff
@@ -0,0 +1,48 @@
+/****************************************************************************
```
rainemak: This actually should go to sailfish-components-webview as will be eventually shared component.

rainemak:
Rebased and updated. 

- https://github.com/sailfishos/sailfish-browser/pull/534/commits/1dc7a210e918ed87811340a2821bc3f3521d8a96 fixes issues #https://github.com/sailfishos/sailfish-browser/issues/536 but also fixes selection marker handling in the same case.
- Uses now correct text selection icon but size needs to be rechecked
- Made selection source to be state of the TextSelectionController and copy on state change

rainemak:
Add notification fix commit on top. Will require https://git.merproject.org/mer-core/nemo-qml-plugin-notifications/merge_requests/3

src/pages/components/Overlay.qml
```diff
@@ -212,6 +212,58 @@ Background {
                 opacity: toolBar.visible && webView.privateMode ? toolBar.opacity : 0.0
             }
 
+            Browser.TextSelectionToolbar {
```
jpetrell: Only load just before showing to speed up initial Browser startup?

src/pages/components/Overlay.qml
```diff
@@ -212,6 +212,58 @@ Background {
...
+                onClear: {
+                    if (controller) {
+                        controller.clearSelection()
+                    }
+                }
+
+                Component {
+                    id: shareText
+
+                    Popups.ShareTextPage {}
```
jpetrell: Move outside to avoid unnecessary compilation.

jpetrell:
Added few comments about lazy loading/compilation. Otherwise code looks good (and the implementation does already).

src/pages/components/Overlay.qml
rainemak: Yeap. I should have recalled that we have in pagestack that import resolver.

rainemak:
Pushed two commit on top.

jpetrell:
LGTM

PR #540, opened 2016-11-14
[sailfish-browser] Use common link sharing page. Fixes JB#36934

rainemak:
Switch browser to use fixed ShareLinkPage that does not flicker
"No sharing options" labels when sharing page slides in.

@jpetrell review ?


chriadam:
LGTM

PR #542, opened 2016-11-16
[sailfish-browser] Disable call action from selection toolbar if sims are disabled. Fixes JB#36981

rainemak:


rainemak:
@jpetrell @chriadam review ?

jpetrell:
LGTM. Matches phone action hiding logic elsewhere.

PR #544, opened 2016-12-06
[sailfish-browser] Show frequently visited web pages at top of history search. Fixes JB#36478

rainemak:
This commit adds visited count to the history search. It should have been there from day one.

@jpetrell @chriadam review ?

### DIFF
--- a/src/storage/dbworker.cpp
+++ b/src/storage/dbworker.cpp
@@ -550,7 +550,7 @@ void DBWorker::getHistory(const QString &filter)
 
     if (!filter.isEmpty()) {
         filterQuery = filterQuery.arg(QString("(url LIKE :search OR title LIKE :search)"));
-        order = QString("LENGTH(url), title, date ASC");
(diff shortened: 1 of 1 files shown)


src/storage/dbworker.cpp
```diff
@@ -550,7 +550,7 @@ void DBWorker::getHistory(const QString &filter)
 
     if (!filter.isEmpty()) {
         filterQuery = filterQuery.arg(QString("(url LIKE :search OR title LIKE :search)"));
-        order = QString("LENGTH(url), title, date ASC");
+        order = QString("visited_count desc, LENGTH(url), title, date ASC");
```
abranson: It's important that sites rankings decay if they haven't been visited for some time. Maybe bring 'date' up to the second place?

rainemak: good point, will rearrange the order

rainemak:
Updated

PR #550, opened 2017-01-09
[sailfish-browser] Clear restored tab history after session history is restored. Contributes to JB#36736

rainemak:


rainemak:
@pvuorela / @jpetrell review ?

This is just a cleanup and does not affect to the functionality. We were just holding unnecessarily simple data structures in memory through out page's life-cycle (used only once when a tab is read back from persistent storage) .

jpetrell:
LGTM

PR #553, opened 2017-01-31
[sailfish-browser] Update selected text status when cellular is enabled. Fixes JB#37557

rainemak:
This improves phone number selection. The error that users saw were such that
selected phone number didn't show "Call" icon in the toolbar. The "Call" icon
was shown only after either of the selection handles were moved a bit.

### DIFF
--- a/src/pages/components/TextSelectionController.qml
+++ b/src/pages/components/TextSelectionController.qml
@@ -30,10 +30,14 @@ MouseArea {
     property string searchUri
     property bool isPhoneNumber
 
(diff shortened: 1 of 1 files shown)


rainemak:
@jpetrell @pvuorela quick review ?

pvuorela:
Would be logical if isPhoneNumber would be true if phone number is selected, ability checked elsewhere?

But check was already done on that level, so LGTM.

rainemak:
@pvuorela true that. Thanks!

PR #560, opened 2017-05-11
[sailfish-browser] Improve tab matching when activating a tab. Contributes to JB#38635

rainemak:
Always chop trailing slash if fragment or query do not exist. The
QUrl::StripTrailingSlash doesn't remove trailing slash if path is "/".

So this fixes "https://sailfishos.org/about" or "https://sailfishos.org"
kind of urls not to create new tab if the tab exists already.

Can be tested e.g. by using:
xdg-open https://sailfishos.org

This effects only external link handling.


rainemak:
@chriadam review ?

chriadam:
LGTM

PR #561, opened 2017-05-11
Add setting to hide close all tabs action.

jusa:


src/pages/components/TabView.qml
```diff
@@ -37,6 +38,12 @@ SilicaListView {
 
     spacing: Theme.paddingMedium
 
+    ConfigurationValue {
+        id: showCloseAllAction
+        key: "/apps/sailfish-browser/settings/show_close_all"
+        defaultValue: true
+    }
+
```
rainemak: Drop after object properties e.g. before PullDownMenu declaration.

Order should be like:
```
id
property declarations
signal declarations
JavaScript functions
object properties
child objects
states
transitions
```

jusa: Moved before PullDownMenu

rainemak:
:+1: LGTM, thanks for contribution :)

Issue #563, opened 2017-06-14
On SFOS browser logging into GitHub flicker and browser crash.


CaptainKhan:
How to replicate:
1. clear online/offline cache
2. clear cookies, history
3. navigate to github and login using SFOS

The browser starts widely flickering before it eventually shuts down. After browser app restart login is complete and there is no flicker.

PR #564, opened 2017-06-20
Throttle setting of bottom margin

rainemak:


rainemak:
This should fix also https://github.com/sailfishos/sailfish-browser/issues/563

@pvuorela maybe you'd like to take this to test ride? See the other commit that this fixes.

rainemak:
Not there yet. Needs something more.

pvuorela:
Already better, but...

- Facebook chat open, click on message field -> gets focus, keyboard appears on top of the field (no change here).
- Scroll page to have field visible and then swipe keyboard away -> keyboard sized margin stays on the screen. Sometimes it works, though, and if keyboard swipe is changed to clicking the page outside the message field it works ok too.

rainemak:
Scrolling input to visible region needs some fixes, especially textarea handling.

That's good finding that keyboard sized margins stays on the screen. Will check, thanks!

rainemak:
Updated. Split updateViewMargins into two. updateViewMargins is guarded and called similarly as it used to be, resetViewMargins resets view margins as per content height. When vkb is closed, resetViewMargins is called directly.

pvuorela:
Swipe issue now fixed, yay. LGTM with i-don't-really-now-browser-well disclaimer. But changes seem sane.

PR #565, opened 2017-06-21
[sailfish-browser] Fix user-agent override pattern matching. Contributes to JB#30687

rainemak:
Got broken by:
https://github.com/sailfishos/sailfish-components-webview/blob/master/lib/webengine.h#L23

@chriadam review ?


chriadam:
LGTM

PR #572, opened 2017-09-05
Do not start margin change throttle without content

rainemak:


chriadam:
What happens if contentHeight bounces between some value and zero, repeatedly?

rainemak:
Zeros would be ignored. However, that's not a normal case as when frame metrics starts appearing we do have content <= something to render as well. Normal bouncing case is between less than screenful of content and multiple screenfuls of content.

chriadam:
And in that case, the throttle would still apply.  Cool.  LGTM

rainemak:
Yes, it would still apply. Will merge.

PR #573, opened 2017-10-02
Sync posting clear window task and GL context creation

rainemak:


rainemak:
@pvuorela @mkosola review ?

PR #575, opened 2017-10-05
[sailfish-browser] Shorten UA update interval. Contributes to JB#39970

rainemak:
Shorten UA update interval to two every second day.


xfade:
LGTM

PR #576, opened 2017-10-05
Fix changing to fullscreen with HTML5 Fullscreen API

rainemak:


saukko:
👍

pvuorela:
Normal browser review disclaimer plus wasn't able to trigger initial problem, but LGTM.

Noticed there is fullscreenHeightChanged too which is connected updateViewMargins, should that be changed?

rainemak:
fullscreenHeight is just current height of the chrome window (the other window aka qquickwindow) and that is not affected.

Only affected place is the fullscreen mode change through HTML5 fullscreen API which must not be guarded similarly as we do normally when hiding toolbar with a pan/flick gesture.

PR #586, opened 2018-01-02
Add BROWSER environment variable

rainemak:


rainemak:
@saukko  review ?

rpm/sailfish-browser.spec
```diff
@@ -101,17 +101,20 @@ Unit tests and additional data needed for functional tests
 %install
 rm -rf %{buildroot}
 %qmake5_install
-chmod +x %{buildroot}/%{_oneshotdir}/*
+chmod +x %{buildroot}/%{_oneshotdir}/* || :
```
pvuorela: Don't think we need || : protection on compile time, i.e. on %install. Better fail if packaging doesn't succeed.

data/70-browser.conf
```diff
@@ -0,0 +1 @@
+BROWSER=/usr/bin/sailfish-browser
```
pvuorela: Should have invoker etc? Colon separated list of commands so I'd suppose it's ok to use: https://manpages.debian.org/stretch/man-db/man.1.en.html

rpm/sailfish-browser.spec
saukko: Yes, that is correct. Shouldn't be used in %install.

data/70-browser.conf
saukko: Was also pondering if this should be with or without /usr/bin/ see reference from https://cgit.freedesktop.org/xdg/xdg-utils/tree/scripts/xdg-open.in#n414

Not sure if it makes any difference and otherwise looks good to me, just pondering :)

rainemak:
Updated

pvuorela:
LGTM

PR #593, opened 2018-03-05
[browser] Improve local file handling when opening a local externally. Contributes to JB#40671

rainemak:
Unknown mime type triggers still misleading download notification even though
the file is already in the file system.

### DIFF
--- a/src/browser/declarativewebutils.cpp
+++ b/src/browser/declarativewebutils.cpp
@@ -92,6 +92,30 @@ void DeclarativeWebUtils::handleDumpMemoryInfoRequest(QString fileName)
     }
 }
 
+void DeclarativeWebUtils::openUrl(QString url)
+{
+
+    QFileInfo fileInfo(url);
+    QUrl targetUrl(url);
+
+    if (targetUrl.scheme().isEmpty()) {
+        QUrl tmpUrl;
+        if (fileInfo.isAbsolute()) {
+            tmpUrl = QUrl::fromLocalFile(url);
+        } else {
+            QUrl baseUrl = QUrl::fromLocalFile(QDir::currentPath() + QDir::separator());
+            tmpUrl = baseUrl.resolved(url);
+        }
+
+        if (QFileInfo::exists(tmpUrl.path())) {
+            targetUrl = tmpUrl;
+        }
+    }
+
+    url = targetUrl.toEncoded();
+    emit openUrlRequested(url);
+}
+
 void DeclarativeWebUtils::updateWebEngineSettings()
 {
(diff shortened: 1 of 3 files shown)


src/browser/declarativewebutils.cpp
```diff
@@ -92,6 +92,30 @@ void DeclarativeWebUtils::handleDumpMemoryInfoRequest(QString fileName)
     }
 }
 
+void DeclarativeWebUtils::openUrl(QString url)
+{
+
+    QFileInfo fileInfo(url);
```
saukko: You could create this later only after checking the targetUrl.

src/browser/declarativewebutils.cpp
```diff
@@ -92,6 +92,30 @@ void DeclarativeWebUtils::handleDumpMemoryInfoRequest(QString fileName)
...
+    QUrl targetUrl(url);
+
+    if (targetUrl.scheme().isEmpty()) {
```
saukko: QUrl::isRelative() does the same thing than .scheme().isEmpty(), right?

src/browser/declarativewebutils.h
```diff
@@ -42,6 +42,7 @@ public slots:
     QString homePage() const;
     void clearStartupCacheIfNeeded();
     void handleDumpMemoryInfoRequest(QString fileName);
+    void openUrl(QString url);
```
saukko: Could this be const QString &url and then use tmpUrl inside openUrl by taking it one step higher?

rainemak: Not big and did want to amend while cnanging this. Qt has copy-on-write copy constructor.

src/browser/declarativewebutils.cpp
rainemak: Nope, doesn't work here as url scheme could be empty. E.g. ../../foobar/mytextfile.txt is relative file path but from url point of view is false.

src/browser/declarativewebutils.cpp
rainemak: Not a big deal either as allocated in stack on variables will be free'ed after the call. Nevertheless, apparently left one extra empty line there.

rainemak: Main thing is that google.fi keeps on working which also looks like relative.

src/browser/declarativewebutils.h
rainemak: Personally prefer always const refs but as all were non const didn't start changing them.

saukko:
:+1:

PR #595, opened 2018-03-07
Add remorse popup for closing all tabs

rainemak:
This contains two commits.

### DIFF
--- a/src/pages/components/TabView.qml
+++ b/src/pages/components/TabView.qml
@@ -77,6 +77,11 @@ SilicaListView {
         defaultValue: true
     }
 
+    // Remorse popup for closing all tabs.
+    RemorsePopup {
+        id: closeAllTabsRemorse
+    }
+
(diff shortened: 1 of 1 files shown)


monich:
That's a bit of an overkill :) With the remorse timer, the setting is unnecessary, IMO.

rainemak:
@monich I know that some people likes to remove "Close all tabs" completely. This would satisfy that need as well.

rainemak:
Updated PR and cleaned settings side of it

pvuorela:
LGTM

src/pages/components/TabView.qml
```diff
@@ -77,6 +77,11 @@ SilicaListView {
         defaultValue: true
     }
 
+    // Remorse popup for closing all tabs.
```
pvuorela: Element and id say pretty much the same thing already :)

PR #602, opened 2018-08-07
[browser] Fix Qt 5.9 compatibility issues. Contributes to JB#42592

adenexter:
This might be take it or leave it stuff in the end. I was seeing a crash in the DeclarativeWebContainer code but I can't see anything actually wrong with the original version so its possible that despite having been rebuilt in OBS I was pulling a stale version and the fix was that I had a guaranteed build against Qt 5.9.

Likewise, the MOUSE_REGION window property would ideally be deprecated since its almost functionally identical to the window mask. The problem as it turns out is its not possible to create a QRegion from a rectangle with a 0 width or height and an empty QRegion is interpreted as an infinite region and all touch events to the web view are blocked when the crome is hidden. Ideally we'd hide the chrome surface too in that circumstance but that's likely to have knock on affects.

### DIFF
--- a/src/core/declarativewebcontainer.cpp
+++ b/src/core/declarativewebcontainer.cpp
@@ -598,10 +598,10 @@ bool DeclarativeWebContainer::eventFilter(QObject *obj, QEvent *event)
 
 bool DeclarativeWebContainer::event(QEvent *event)
 {
-    QPlatformWindow *windowHandle;
-    if (event->type() == QEvent::PlatformSurface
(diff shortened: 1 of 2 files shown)


pvuorela:
I think you want to target the 'next' branch for any changes. Got confused myself on these just a while ago :)

adenexter:
Yeah, I figured that out for myself when setting up a webhook and moved my change on top of next but apparently didn't follow through with retargeting the PR. I'll retarget it now.

PR #603, opened 2018-08-15
[browser] Use the new PageStack animatorPush(). Contributes to JB#21786

jpetrell:


jpetrell:
Removed animator-based push for now from the app, only used in settings. Need to fix page background implementation to work with animators before enabling the new transition path for the browser.

jpetrell:
Adapted Browser background to work with animator-based push, all page pushes inside Browser should now use the new animator.

PR #607, opened 2018-09-27
Create background for each page if doesn't background exists

rainemak:
In theory we could also use pageStack.currentPage.asyncObject.pageCompleted
internals but this takes care of synchronous page pushes as well.

Having pagePush method on the browser side is not enough as there
can be arbitrary page push / pushAnimator calls from the components
that browser is using.


jpetrell:
Nice and simple. :+1: In testing couldn't find issues.

PR #608, opened 2018-10-02
Check foreground from browser window visibility instead of Qt.application.active

rainemak:


src/browser.qml
```diff
@@ -21,6 +21,7 @@ ApplicationWindow {
     id: window
 
     readonly property bool largeScreen: Screen.sizeCategory > Screen.Medium
+    readonly property int browserVisibility: _mainWindow && _mainWindow.visibility || QuickWindow.Window.Hidden
```
jpetrell: _mainWindow.visibility || QuickWindow.Window.Hidden looks weird?

src/pages/components/TabView.qml
```diff
@@ -9,7 +9,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import QtQuick 2.1
+import QtQuick 2.2
+import QtQuick.Window 2.2 as QuickWindow
```
pvuorela: Not used here?

src/browser.qml
pvuorela: "_mainWindow ? _mainWindow.visibility : QuickWindow.Window.Hidden" perhaps?

rainemak: I tend to use that notation. Will check to ternary operator.

src/pages/BrowserPage.qml
```diff
@@ -40,7 +40,7 @@ Page {
     }
 
     function bringToForeground(window) {
-        if (!Qt.application.active && window) {
+        if ((webView.visibility == QtWindow.Window.Minimized) && window) {
```
pvuorela: How would webView.visible work generally? Shorter and perhaps takes care of Window.Hidden?

rainemak: It's still visible when on cover but updated a bit other way round.

src/pages/BrowserPage.qml
```diff
@@ -40,7 +40,7 @@ Page {
     }
 
     function bringToForeground(window) {
-        if (!Qt.application.active && window) {
+        if ((webView.visibility < QuickWindow.Window.Maximized) && window) {
```
rainemak: I don't think that QWindow::AutomaticVisibility could cause issue here.

pvuorela: As small note conceptually raise() should be valid operation with everything else than fullscreen. Or maybe even that if considering app being fullscreen and window app on top of it. Method here doesn't look like getting called too much so could even remove window state checks, though doesn't matter much.

src/pages/components/Overlay.qml
```diff
@@ -523,6 +524,7 @@ Background {
                 model: webView.tabModel
                 portrait: tabPage.isPortrait
                 privateMode: webView.privateMode
+                pullDownVisible: webView.visibility >= QuickWindow.Window.Maximized
```
pvuorela: A little curious why this is done. Tab page per se not included in cover? Introduced 2014 with commit "Hide pulley menu, close tab button, title, and the opacity ramp in cover"

rainemak: Ah, so true :+1: Will cleanup.

PR #610, opened 2018-10-10
[browser] Color tab item controls always light on dark. Fixes JB#43267

pvuorela:
Web pages are more commonly light than dark so always dark controls
suit these better. Alternative whitening gradient on top would
also act quite differently.

@rainemak  @jpetrell


rainemak (approved):
LGTM even though not optimal.

PR #611, opened 2018-10-15
Cleanup old overlay active handling

rainemak:
Respect keyboard focus for the overlay.


src/pages/components/OverlayAnimator.qml
```diff
@@ -68,27 +68,6 @@ Item {
     }
 
     state: _chromeVisible
-
-    // TODO: Fix real cover. Once that is fixed, we should remove this block.
-    onActiveChanged: {
```
pvuorela: Remove also the property?

rainemak: Sure thanks

PR #613, opened 2018-10-23
[browser] Use shared icon grid component in favorites grid. Contributes to JB#43423

jpetrell:


rainemak:
Two rows of favorites are collapsing almost on top of each other on landscape.

PR #614, opened 2018-10-31
[browser] Add baseXOffset for FavoriteGrid's context menu. Fixes JB#43519

rainemak:
On full hd portrait difference is 1048 vs 1080 and now the context
menu on favorite grid covers the full screen width.


rainemak:
Declining. Let's handle this inside the context menu.

PR #616, opened 2018-11-22
[oneshot] Do not use perl regex. Contributes to JB#40843

Tomin1:
Busybox does not support perl regex on grep. We want to be compatible
with busybox as well as GNU grep.

### DIFF
--- a/oneshot.d/browser-update-default-data
+++ b/oneshot.d/browser-update-default-data
@@ -8,8 +8,7 @@ fi
 
 if [ ! -f "$MOZEMBED_DIR/ua-update.json" -o ! -s "$MOZEMBED_DIR/ua-update.json" ]; then
     # Remove lines starting with a comment.
-    # Using Perl regexp so that this works on device.
-    grep -vP  ^\\s*// $BROWSER_DATA_DIR/ua-update.json.in > "$MOZEMBED_DIR/ua-update.json"
+    sed  '\|^\s*//|d' $BROWSER_DATA_DIR/ua-update.json.in > "$MOZEMBED_DIR/ua-update.json"
 fi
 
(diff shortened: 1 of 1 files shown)


oneshot.d/browser-update-default-data
```diff
@@ -8,8 +8,7 @@ fi
 
 if [ ! -f "$MOZEMBED_DIR/ua-update.json" -o ! -s "$MOZEMBED_DIR/ua-update.json" ]; then
     # Remove lines starting with a comment.
-    # Using Perl regexp so that this works on device.
-    grep -vP  ^\\s*// $BROWSER_DATA_DIR/ua-update.json.in > "$MOZEMBED_DIR/ua-update.json"
+    grep -vE  '^\s*//' $BROWSER_DATA_DIR/ua-update.json.in > "$MOZEMBED_DIR/ua-update.json"
```
rainemak: Apparently requires updated grep if done like proposed. Already with existing setup could done like this to get the same result:

`sed  '/^\s*\/\// d' $BROWSER_DATA_DIR/ua-update.json.in > "$MOZEMBED_DIR/ua-update.json"`

Tomin1: This sed solution looks good. I didn't happen to test this grep -E on a device that has GNU grep, although that would have been good, and basically just assumed that if it works on busybox it works with GNU as well. Have to be more careful next time.

Btw, is the title fine? I'm wondering if it should be [sailfish-browser] instead of [oneshot] to look good on change log. It might be that it doesn't really matter.

rainemak: It's just name of the component that is in the square brackets. Technically doesn't matter.
Title is good.

rainemak: Approved also this way.

PR #617, opened 2018-12-10
[browser] Remove permanent highlight on tab close icon. Fixes JB#44094

adenexter:


rainemak:
Target branch should be "next" other than that LGTM

PR #623, opened 2019-02-26
Merge esr45

rainemak:


lkraav:
While this was a significantly helpful update, are ESR52 or dare I say ESR60 updates also on the horizon?

Native Web Components would be very useful, but this only arrived in Firefox 63.

PR #630, opened 2019-05-30
[sailfish-browser] Allow privileged apps to close tabs. Contributes to JB#45881

chriadam:


chriadam:
@rainemak I'm not sure whether this is the way it should be done, but it seems to work?

src/browser/browserservice.cpp
```diff
@@ -174,6 +174,20 @@ void BrowserUIService::requestTab(int tabId, const QString &url)
...
+{
+    if (!isPrivileged()) {
+        return;
+    }
+
+    Q_D(BrowserUIService);
+
+    DeclarativeWebContainer::instance()->closeTab(tabId);
+    const QDBusMessage &msg = message();
+    QDBusMessage reply = msg.createReply();
```
rainemak: Should we pass here true/false if closing succeeds? Or can you safely assume that closing succeeds.

Pondering a case like:
- 1) Tab requested
- 2) Something loaded in the tab
- 3) User closes the tab
- 4) close tab called

=> We it be helpful to know that user has interrupted the sequence? I.e. the tab you created 1st step must have been closed by user if close in 4th step fails. I'm just thinking loud.

chriadam: From the calling application's perspective, I don't think there's much they can do.  They want the tab to be closed, and they ask for it to be closed - but if cannot be closed (either because the user already closed it, or some other error) that application can't "do anything" about it, I think.  So not much use for the return value IMO.

rainemak:
LGTM

PR #640, opened 2019-09-24
[sailfish-browser] Add backup unit. Contributes to JB#29667

salmelamike:
Moved backup unit here from jolla-vault. Reduced the usage of qtaround

### DIFF
--- a/backup-unit/browserunit.cpp
+++ b/backup-unit/browserunit.cpp
@@ -1,0 +1,100 @@
+#include "logging.h"
+#include <vault/unit.h>
+#include <QProcess>
+#include <QCoreApplication>
+#include <QVariantList>
+#include <QVariantMap>
+#include <QDebug>
+#include <QLoggingCategory>
+#include <sys/types.h>
+#include <signal.h>
+#include <set>
+#include <unistd.h>
+#include <stdexcept>
+
+void stop_browser()
+{
+    qCDebug(lcBackupLog) << "Terminating browser";
+    QProcess ps;
+    auto get_browser_pids = [&ps]() { 
+        std::set<int> res;
+        ps.execute("pgrep", {"-f", "sailfish-browser"});
+        if (ps.exitStatus() == QProcess::CrashExit) {
+            qCDebug(lcBackupLog) << "pgrep failed";
+            return res;
+        }
+
+        auto data = QString(ps.readAllStandardOutput()).split("\n");
+        for (auto const &line : data) {
+            if (!line.length())
+                continue;
+
+            bool ok = false;
+            auto pid = line.toInt(&ok);
+            if (ok)
+                res.insert(pid);
+        }
+        return res;
+    };
+
+    auto const sec0_1 = 100000;
+
+    auto sig = SIGINT;
+    auto counter = 31;
+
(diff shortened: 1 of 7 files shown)


backup-unit/browserunit.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include <QVariantList>
+#include <QVariantMap>
+#include <QDebug>
+#include <sys/types.h>
+#include <signal.h>
+#include <set>
+#include <unistd.h>
+#include <stdexcept>
+
+
```
rainemak: Extra empty line.

backup-unit/browserunit.cpp
```diff
@@ -0,0 +1,105 @@
...
+void stop_browser()
+{
+    qDebug() << "Terminating browser";
```
rainemak: You could introduce:

```
Q_LOGGING_CATEGORY(lcCoreLog, "org.sailfishos.browser.backup", QtWarningMsg)
```
to the logging.cpp and declare it in logging.h. As this is standalone binary we could compile those into this project directly.

When above is added then qCDebug(lcBackupLog) for qDebug and qCWarning for qWarnings.

backup-unit/browserunit.cpp
```diff
@@ -0,0 +1,105 @@
...
+    QProcess ps;
+    auto get_browser_pids = [&ps]() {
+        std::set<int> res;
+        ps.start("pgrep", {"-f", "sailfish-browser"});
+        ps.waitForFinished(-1);
+        if (ps.exitStatus() == QProcess::CrashExit) {
+            throw std::runtime_error("pgrep failed");
+            return res;
```
rainemak: Which one is more important? I see that there's try-catch in the main. The get_browser_pids of set<int> won't have anything meaningful as this return won't be handled.

I think qCWarning + return res as empty should sufficient.

backup-unit/browserunit.cpp
```diff
@@ -0,0 +1,105 @@
...
+    };
+}
+
+int main(int argc, char *argv[])
+{
+    try {
+        QCoreApplication app(argc, argv);
+        stop_browser();
+        vault::unit::execute(vault::unit::getopt(), info);
+
```
rainemak: An extra empty line.

backup-unit/browserunit.cpp
```diff
@@ -0,0 +1,105 @@
+#include <vault/unit.h>
+#include <qtaround/os.hpp>
```
rainemak: Is this really needed?

backup-unit/browserunit.cpp
```diff
@@ -0,0 +1,105 @@
...
+        , {"options", QVariantMap({{"overwrite", true}})}
+    };
+}
+
+int main(int argc, char *argv[])
+{
+    try {
+        QCoreApplication app(argc, argv);
+        stop_browser();
+        vault::unit::execute(vault::unit::getopt(), info);
```
rainemak: This getopt looks to be something that we could drop from the interface.

rainemak: Though, not a big deal

PR #644, opened 2019-10-21
[sailfish-browser] Filter out accidental clicks near the display edge. Fixes JB#47839

jpetrell:
…. Fixes JB#47839

### DIFF
--- a/src/pages/components/IconButton.qml
+++ b/src/pages/components/IconButton.qml
@@ -11,6 +11,7 @@
 
 import QtQuick 2.0
 import Sailfish.Silica 1.0
+import Sailfish.Silica.private 1.0
 
 IconButton {
     // Keep icon enabled so that drag area can filter mouse events.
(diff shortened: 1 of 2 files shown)


jpetrell:
Normally we use Theme.paddingLarge, but since the main browser controls are located right at the edge don't want to reduce the reactive area too much.

src/pages/components/IconButton.qml
```diff
@@ -32,5 +33,9 @@ IconButton {
         }
     }
 
+    onPressed: DragFilter.begin(mouse.x, mouse.y)
+    onCanceled: DragFilter.end()
+    DragFilter.screenMargin: Theme.paddingMedium
```
rainemak: Not related to this but noticed while testing that tab icon and menu icon do not highlight when tapping near the bottom corner.

src/pages/components/ToolBar.qml
```diff
@@ -264,6 +265,10 @@ Column {
                 }
             }
 
+            DragFilter.screenMargin: Theme.paddingMedium
+            onPressed: DragFilter.begin(mouse.x, mouse.y)
+            onCanceled: DragFilter.end()
```
rainemak: Nit (of nits): Different order than IconButton :)

rainemak:
Noticed a highlight issue but that's not related to this one PR se. LGTM :+1:

src/pages/components/ToolBar.qml
jpetrell: Fixed.

PR #646, opened 2019-11-07
Ensure empty browser opens with URL entry

llewelld:
When the browser opens for the first time there is no page to display,
and the user should be presented with the URL entry field open.

Previous changes which introduced the TLS panel broke this behaviour.
These changes fix the regression.

The problem seems to be twofold. First, the atTop/atBottom state
behaviour changed (because panels with multiple sizes were introduced,
meaning it was possible to change from a large panel size to a smaller
panel size and still be atTop). The initial URL entry field relies on a
switch from atTop = false to atTop = true in order to trigger a
searchField.requestingFocus change in Overlay.qml. Second, the
initial _showUrlEntry entry also has to be true in order to trigger this
requestingFocus state change. This will revert to false as soon as
atBottom becomes true (i.e. as soon as the URL entry field is closed).

### DIFF
--- a/src/core/core.pri
+++ b/src/core/core.pri
@@ -1,5 +1,8 @@
 INCLUDEPATH += $$PWD
 
+PKGCONFIG += \
+        sailfishsilica
+
 # C++ sources
 SOURCES += \
     $$PWD/declarativewebcontainer.cpp \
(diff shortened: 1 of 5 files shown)


llewelld:
> Then starting for the first time the overlay is not atTop initially. One can see it at the bottom first from where it switching to the top position without animation (which is ok). Would be better to have it at the top position directly.

As we discussed, Given it's not possible (or at least, not straightfoward) to make the overlay appear immediately over the empty web container surface, I've changed the web container background to use the Theme background colour rather than white, which definitely makes the opening flow less visually abrupt.

rainemak:
LGTM :+1:

PR #648, opened 2019-11-13
[mdm] Block browser use when disabled by MDM. Fixes JB#47995

abranson:


settings/browser.qml
```diff
@@ -38,8 +40,12 @@ Page {
                 title: qsTrId("settings_browser-ph-browser")
             }
 
+            DisabledByMdmBanner {
+                active: !AccessPolicy.browserEnabled
+            }
```
pvuorela: Are there any needs to protect the settings? If not, I'd just skip these parts.

src/pages/BrowserPage.qml
```diff
@@ -334,7 +337,11 @@ Page {
             bringToForeground(webView.chromeWindow)
             window.activate()
         }
-        onActivateNewTabViewRequested: activateNewTabView()
+        onActivateNewTabViewRequested: {
+            // Refuse if blocked by MDM
+            if (!AccessPolicy.browserEnabled) return
+            activateNewTabView()
+            }
```
pvuorela: Wrong indentation.

settings/browser.qml
abranson: Just consistency with the camera blocking really.

src/pages/BrowserPage.qml
abranson: I moved this to the activateNewTabView() function anyway, to catch the cover action.

src/pages/components/DisabledByMdmView.qml
```diff
@@ -0,0 +1,48 @@
+import QtQuick 2.0
```
rainemak: Move Camera's DisabledByMdmvView to the settings (same place as DisabledByMdmBanne) and use that here.

Just leave "text" property to the interface.

settings/browser.qml
rainemak: But does it make sense really?

abranson: I can imagine someone filing a bug if it's still active.

rainemak: Let's keep it there.

src/browser.qml
```diff
@@ -96,4 +97,8 @@ ApplicationWindow {
             z: -1
         }
     }
+    
+    Browser.DisabledByMdmView {
+        active: !AccessPolicy.browserEnabled
```
rainemak: onActiveChaged: if (active) webView.tabModel.clear()

src/pages/BrowserPage.qml
```diff
@@ -46,6 +46,8 @@ Page {
     }
 
     function activateNewTabView() {
+        // Refuse if blocked by MDM
+        if (!AccessPolicy.browserEnabled) return
```
rainemak: Having a dead cover action is bad. I'd bring the app to foreground at min,

src/pages/BrowserPage.qml
```diff
@@ -297,7 +299,7 @@ Page {
     }
 
     CoverActionList {
-        enabled: (browserPage.status === PageStatus.Active || browserPage.tabPageActive || !webView.tabModel || webView.tabModel.count === 0)
+        enabled: AccessPolicy.browserEnabled && (browserPage.status === PageStatus.Active || browserPage.tabPageActive || !webView.tabModel || webView.tabModel.count === 0)
```
rainemak: I don't think this is good. Having a dead cover action is bad.

src/pages/BrowserPage.qml
```diff
@@ -310,6 +312,9 @@ Page {
     Connections {
         target: WebUtils
         onOpenUrlRequested: {
+            // Refuse if blocked by MDM
+            if (!AccessPolicy.browserEnabled) return
```
rainemak: I'd still bring browser to the foreground.

```
if (!AccessPolicy.browserEnabled) {
    bringToForeground(webView.chromeWindow)
    window.activate()
    return
}
```

abranson: Done

src/browser.qml
abranson: Done

src/pages/components/DisabledByMdmView.qml
abranson: Done. Used 'activity' as the text, as it's a parameter to the larger standard text.

rainemak:
Update also copyrights for the affected files

src/browser.qml
```diff
@@ -96,4 +99,12 @@ ApplicationWindow {
             z: -1
         }
     }
+    
+    DisabledByMdmView {
+        enabled: !AccessPolicy.browserEnabled
+        activity: "Web browsing"
```
rainemak: qsTrId for this

abranson: Done

src/core/declarativewebcontainer.cpp
```diff
@@ -864,11 +873,18 @@ void DeclarativeWebContainer::initialize()
     // 2) model has tabs, load initial url or active tab.
     bool firstUseDone = DeclarativeWebUtils::instance()->firstUseDone();
     if ((m_model->waitingForNewTab() || m_model->count() == 0) && (firstUseDone || !m_initialUrl.isEmpty())) {
-        QString url = m_initialUrl.isEmpty() ? DeclarativeWebUtils::instance()->homePage() : m_initialUrl;
+        QString url = m_initialUrl;
+        if (m_initialUrl.isEmpty()) {
+            if (!browserEnabled()) url = ABOUT_BLANK;
```
rainemak:
```
if (!browserEnabled())
    url = ABOUT_BLANK;
else
    url = DeclarativeWebUtils::instance()->homePage();
```

src/browser.qml
```diff
@@ -96,4 +99,19 @@ ApplicationWindow {
             z: -1
         }
     }
+    
+    DisabledByMdmView {
+        enabled: !AccessPolicy.browserEnabled
+        //% "Web browsing"
+        activity: qsTrId("sailfish_browser-la-web_browsing");
+        onEnabledChanged: {
+            if (enabled) {
```
rainemak: In following scenario you'd still have content running... 
1) Browser is running on private mode
2) Browser is disabled via MDM

src/core/declarativewebcontainer.cpp
abranson: Done. Though I think this should happen in all privateMode, rather than just when the browser is disabled.

src/browser.qml
abranson: Also if private mode was previously used, you'd get those tabs restored. Inserted a tab model clear after the change.

rainemak:
Still LGTM

PR #651, opened 2019-11-26
Revert background colour match to ambience

llewelld:
This reverts some of the changes made in commit
3b56be95b265854cc77d96022bf825b084bac22d

Originally the browser background colour was always cleared to white. On
first opening there's a slight delay between the browser opening with a
cleared canvas and the URL panel opening, resulting in an abrupt colour
switch. The background was therefore changed to match the panel colour
of the current ambience.

As a result of this change, in in the more common case of opening a new
tab, the canvas opens dark and then later renders the web page. On the
basis that most web pages are a light colour, this PR reverts the canvas
colour to always be white to avoid an abrupt switch in this case.


rainemak (approved):
LGTM :+1:

Siltaar:
Please revert this revert.

Most websites nowadays implements a dark background theme, automatically sync with system preferences. This is true for Windows, Mac OS and Linux.

Opening a new tab now result in a white flash covered by dark background websites.

And with WebExtensions enabled we could even install Dark Reader to ensure that every websites are displayed with dark backgrounds.

PR #653, opened 2019-12-09
[sailfish-browser] Use remorse delete for favorites. JB#48357

rainemak:
This commit also removes obsoleted animated removal call.


rainemak:
@jpetrell review ?

jpetrell:
I don't see "Deleted" label on so small remorse item. Remorse within grid item works with large grid items like we have in Notes and Gallery, but for smaller grid items like what you see in Media playlist grid, Browser and People favorites, we normally show the remorse below the row as a separate, wider item. I will try to extend GridItem to cover that case implicitly in Silica side, if that works the change here should then work too.

rainemak:
Rebased this one. @jpetrell do you have comments?

jpetrell:
No looks good. 👍 Sorry tested this earlier, but forgot to write LGTM.

PR #654, opened 2019-12-11
[sailfish-browser] Removing hardcoded nemo home. Fixes JB#47612

mkenttala:


rainemak:
Could you cleanup a bit this line from browserpaths.cpp:

> // assumes that correct groupname is same as username (e.g. nemo:nemo)
to
> // assumes that correct groupname is same as username

rainemak:
Clean up browserpaths.cpp. Other than that LGTM.

:+1:

PR #657, opened 2019-12-19
Use selection handles from the Sailfish.WebView.Controls module

adenexter:
These are the exact same thing in a module where they can be used in the
WebView as well.


rainemak:
Rebase (as per need) and remove WIP from the merge commit.

rainemak:
There's also a typo in the merge commit => TextSelection to Controls.

PR #658, opened 2019-12-19
[browser] Avoid toolbar button offset while loading. Fixes JB#48356

pvuorela:
Regression on commit 31edc522b560, switching hiding to active
property makes it fade in/out too.

Also adjusted some minor details while at it.

@rainemak @mkenttala


mkenttala (approved):

rainemak:
I don't like much that one mixes style related changes to actual function changes but meh.

LGTM. Thanks @pvuorela from the fix.

pvuorela:
Yea, could do different commits, but doesn't always feel big gain with only small amount of changed lines.

PR #668, opened 2020-03-10
Download logging and cleanup browser.helperApps.deleteTempFileOnExit

rainemak:


src/browser/downloadmanager.cpp
```diff
@@ -56,18 +57,23 @@ DownloadManager::~DownloadManager()
 
 void DownloadManager::recvObserve(const QString message, const QVariant data)
 {
-
     if (message != "embed:download") {
         // here we are interested in download messages only
         return;
     }
 
+
```
rainemak: Will remove this before merging.

rainemak: done

PR #669, opened 2020-03-23
Move search engine configuration to WebEngineSettings. 

adenexter:


adenexter:
See https://github.com/sailfishos/sailfish-components-webview/pull/48

PR #687, opened 2020-06-22
[sailfish-browser] Fix opening certificate info toolbar pane. Fixes JB#50205

paveltkv:


llewelld:
A nice neat fix, and it doesn't revert the JB#47875 fix that introduced it. So LGTM.

PR #689, opened 2020-06-25
[browser] Cleanup security preferences. JB#50423

rainemak:
The "security.tls.version.min" is defaults to 1 in security-prefs.js.
Hence, not needed on the browser side.

The pref "security.csp.speccompliant" no longer exists. So we
can just remove it.

### DIFF
--- a/src/browser/declarativewebutils.cpp
+++ b/src/browser/declarativewebutils.cpp
@@ -134,9 +134,6 @@ void DeclarativeWebUtils::updateWebEngineSettings()
     // Without this pref placeholders get cleaned as soon as a character gets committed
     // by VKB and that happens only when Enter is pressed or comma/space/dot is entered.
     webEngineSettings->setPreference(QString("dom.placeholder.show_on_focus"), QVariant(false));
-
-    webEngineSettings->setPreference(QString("security.alternate_certificate_error_page"), QString("certerror"));
-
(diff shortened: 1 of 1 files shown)


rainemak:
See also https://git.sailfishos.org/mer-core/gecko-dev/merge_requests/117

src/browser/declarativewebutils.cpp
```diff
@@ -152,13 +149,6 @@ void DeclarativeWebUtils::updateWebEngineSettings()
     webEngineSettings->setPreference(QString("keyword.enabled"), QVariant(true));
 
     setRenderingPreferences();
-
-    // Disable SSLv3
-    webEngineSettings->setPreference(QString("security.tls.version.min"), QVariant(1));
```
llewelld, line 157: It doesn't affect this PR, but I notice FF 78 was due to disable support for TLS 1.0 and 1.1. So this may be something to revisit in the not-too-distant future: https://hacks.mozilla.org/2020/02/its-the-boot-for-tls-1-0-and-tls-1-1/

rainemak: Thanks. We should adjust embedding.js accordingly.

rainemak: Created JB#50522

PR #690, opened 2020-06-29
Delete startupCache if install is newer than it

llewelld:
On installation of embedlite-components-qt5 the package installs the
file /var/lib/_MOZEMBED_CACHE_CLEAN_ .

This change ensures that when the browser is run, the last modified
timestamp of the startupCache is compared against this file, and if the
installed file is newer, the cache is cleared. The startupCache will
then be recreated by xulrunner during execution. Previously the
timestamp was compared against ~/.mozilla/mozembed/_CACHE_CLEAN_ but
since it's no longer createdi it can't be used anymore.

This ensures that any changes made to embedlite javascript components
won't be ignored in favour of older versions in the cache.

### DIFF
--- a/src/browser/declarativewebutils.cpp
+++ b/src/browser/declarativewebutils.cpp
@@ -14,6 +14,7 @@
 #include <QtCore/QDir>
 #include <QtCore/QFile>
 #include <QtCore/QFileInfo>
+#include <QtCore/QFileInfoList>
 #include <QtCore/QLocale>
 #include <QtCore/QSettings>
 #include <QtCore/QStandardPaths>
@@ -35,7 +36,7 @@
 #include "browserpaths.h"
 
 static const auto gSystemComponentsTimeStamp = QStringLiteral("/var/lib/_MOZEMBED_CACHE_CLEAN_");
-static const auto gProfilePath = QStringLiteral("/.mozilla/mozembed");
+static const auto gStartupCacheDir = QStringLiteral("/.mozilla/mozembed/startupCache");
 static const auto defaultUserAgentUpdateUrl = QStringLiteral("https://browser.sailfishos.org/gecko/%APP_VERSION%/ua-update.json");
 
 static DeclarativeWebUtils *gSingleton = 0;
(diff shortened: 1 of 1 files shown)


src/browser/declarativewebutils.cpp
```diff
@@ -75,13 +77,14 @@ void DeclarativeWebUtils::clearStartupCacheIfNeeded()
...
-        if (localStamp.exists() && systemStamp.lastModified() > localStamp.lastModified()) {
-            QDir cacheDir(mostProfilePath + "/startupCache");
-            cacheDir.removeRecursively();
-            QFile(localStampString).remove();
+        QString startupCacheDir = QDir::homePath() + gStartupCacheDir;
+        QDir cacheDir(startupCacheDir);
+        if (cacheDir.exists()) {
+            QFileInfoList files = cacheDir.entryInfoList(QDir::Files, QDir::Time);
+            if (files.count() > 0 && systemStamp.lastModified() > files.first().lastModified()) {
+                qDebug() << "Removing cache to synchronise with fresh embedlite-components-qt5";
```
okodron, line 85: You can just use printf-like `qDebug("Removing cache to synchronise with fresh embedlite-components-qt5");` as it's quite simple here. Then you don't need to include `<QtCore/QDebug>` as well.

src/browser/declarativewebutils.cpp
```diff
@@ -75,13 +77,14 @@ void DeclarativeWebUtils::clearStartupCacheIfNeeded()
...
     if (systemStamp.exists()) {
-        QString mostProfilePath = QDir::homePath() + gProfilePath;
-        QString localStampString(mostProfilePath + QString("/_CACHE_CLEAN_"));
-        QFileInfo localStamp(localStampString);
-        if (localStamp.exists() && systemStamp.lastModified() > localStamp.lastModified()) {
-            QDir cacheDir(mostProfilePath + "/startupCache");
-            cacheDir.removeRecursively();
-            QFile(localStampString).remove();
+        QString startupCacheDir = QDir::homePath() + gStartupCacheDir;
+        QDir cacheDir(startupCacheDir);
```
okodron, line 81: I think it's better to omit intermediate `QString startupCacheDir` variable. Then init `QDir cacheDir(QDir::homePath() + gStartupCacheDir);`.

llewelld: Thanks, yes, the intermediate string is unnecessary. I've combined the two lines.

src/browser/declarativewebutils.cpp
llewelld: Interesting, I wasn't aware this was possible. I've changed it.

llewelld:
Thank you both for the useful review comments. I'll briefly look into a oneshot before merging.

llewelld:
This PR has been superseded by https://github.com/sailfishos/sailfish-browser/pull/693 and https://git.sailfishos.org/mer-core/gecko-dev/merge_requests/126 (both now merged), so I'm closing this as it's no longer needed.

PR #692, opened 2020-07-21
[browser] Do not force chrome to visible when page loaded. JB#50422

rainemak:


rainemak:
@llewelld could you give a look to this? This would alleviate the problem a bit. Without this toolbar is not lowered by panning the content if the content length is right (on clean browser opening https://techcrunch.com). Opening and closing the overlay can be used to workaround the issue.

rainemak:
This is not about padding or margin. It's just that toolbar cannot go away when page loaded and contentheight is "correct".
To be noted that padding doesn't help either on some problematic pages.

llewelld:
> This is not about padding or margin. It's just that toolbar cannot go away when page loaded and contentheight is "correct".

Ah, yes, I don't think we're in disagreement here. I simply meant that this is a nicer solution than the equivalent one we used for email (which was to add additional padding to the email to allow it to scroll).

rainemak:
There's still that extra margin in place if the content length short enough.

rainemak:
Yes @llewelld we're not disagreeing :+1:

PR #693, opened 2020-07-28
Remove startupCache removal code

llewelld:
The startupCache needs to be removed if it's stale (e.g. after component
installation). The code to do this has been moved into xulrunner and so
is no longer needed here.

This change effectively reverts commit
469d6a7dfe43efacf866d7f28d32eda24f3a0a9d


rainemak:
Rebase if needed and merge. LGTM.

PR #696, opened 2020-08-23
[browser] Add browsing history page. Contributes to JB#50675

atatarov:


src/pages/components/HistoryItem.qml
```diff
@@ -0,0 +1,106 @@
...
+    Component {
+        id: contextMenuComponent
+
+        ContextMenu {
+            MenuItem {
+                //% "Open in private tab"
+                text: qsTrId("sailfish_browser-me-open-private-tab")
+                onClicked: {
+                    webView.privateMode = true
+                    openLink()
```
rainemak, line 82: We do not have concept of private tab -- only private and non private mode. I think this deserves a bit more so that "private" flag would be passed to the newly created page (and created so regardless of active model) and the model into which it will be added would be handled separately.

src/pages/components/HistoryPage.qml
```diff
@@ -0,0 +1,44 @@
...
+                EnterKey.onClicked: focus = false
+                onTextChanged: {
+                    historyModel.search(searchField.text)
+                    view.search = searchField.text
+                }
+            }
+        }
+        section {
+            property: "sectionName"
+            delegate: SectionHeader { text: section }
```
rainemak, line 33: This should be time based on the qml side. On the model you should have something like "dataTimeSection" and for the 
```
SectionHeader {
    text: Format.formatDate(section, Formatter.TimepointSectionRelative)
}
```

src/storage/link.cpp
```diff
@@ -10,19 +10,41 @@
...
+    if (fullDate == today.toString(fullDateFormat)) {
+        //% "Today"
+        m_sectionName = qtTrId("sailfish_browser-se-today");
+    } else if (fullDate == yesterday.toString(fullDateFormat)) {
+        //% "Yesterday"
+        m_sectionName = qtTrId("sailfish_browser-se-yesterday");
+    } else {
+        QString standaloneMonthName = QLocale::system().standaloneMonthName(date.date().month());
+        QString year = QString::number(date.date().year());
+        m_sectionName = QString("%1 %2").arg(standaloneMonthName, year);
```
rainemak, line 32: This should be not fixed on the model side like rather platform style time formatting should be used.

src/storage/dbworker.cpp
```diff
@@ -550,13 +550,13 @@ void DBWorker::getHistory(const QString &filter)
 
     if (!filter.isEmpty()) {
         filterQuery = filterQuery.arg(QString("(url LIKE :search OR title LIKE :search)"));
-        order = QString("visited_count DESC, date ASC, LENGTH(url), title");
+        order = QString("visited_count ASC, date DESC, LENGTH(url), title");
```
rainemak, line 553: Both previous and this are off. For both visited_count and date we should have DESC ordering so that most visited would be still preferred. That said, problem is that visited count is take higher presendence than date (most recently visited).

Hence, I'd say that this actually should be like: `order = QString("date DESC, visited_count DESC, LENGTH(url), title");`

rainemak: While you're at this could you fix also 453 that is behind debug logging: "qDebug() << "link id:" << linkId;"

src/storage/dbworker.cpp
```diff
@@ -550,13 +550,13 @@ void DBWorker::getHistory(const QString &filter)
...
     } else {
         filterQuery = filterQuery.arg(1);
         order = QString("date DESC");
     }
 
-    QString queryString = QString("SELECT DISTINCT id, url, title "
+    QString queryString = QString("SELECT id, url, title, date "
```
rainemak, line 559: Add also visited count here after date

src/storage/dbworker.cpp
```diff
@@ -571,10 +571,12 @@ void DBWorker::getHistory(const QString &filter)
...
     while (query.next()) {
+        qint64 timestamp = query.value(3).toLongLong();
         Link link(query.value(0).toInt(),
                   query.value(1).toString(),
                   "",
-                  query.value(2).toString());
+                  query.value(2).toString(),
+                  QDateTime::fromMSecsSinceEpoch(timestamp*1000));
 #if DEBUG_LOGS
         qDebug() << &link;
```
rainemak, line 581: qDebug() << &link << visitedCount;

src/pages/components/HistoryLabel.qml
```diff
@@ -0,0 +1,10 @@
+import QtQuick 2.0
+import Sailfish.Silica 1.0
+
+Label {
+    textFormat: Text.StyledText
+    color: highlighted ? Theme.highlightColor : Theme.primaryColor
```
jpetrell, line 6: Not needed nowadays.

src/pages/components/HistoryItem.qml
```diff
@@ -0,0 +1,106 @@
...
+    function remove() {
+        view.model.remove(model.index)
+    }
+
+    function openLink() {
+        webView.load(model.url, model.title)
+        pageStack.pop()
+    }
+
+    width: view.width
```
jpetrell, line 19: Default Theme.itemSizeSmall is too small for two label row list item, especially if user changes system font size to "Huge". -> contentHeight: Math.max(Theme.itemSizeMedium, column.height + 2*Theme.paddingMedium)

src/storage/link.cpp
rainemak: As noted in an another comment date itself is enough. Further note, translations on the C++ are not preferred. That said, this can be more or less dropped.

src/storage/link.cpp
```diff
@@ -73,14 +95,39 @@ bool Link::isValid() const
...
+QString Link::sectionName() const
+{
+    return m_sectionName;
+}
+
+void Link::setSectionName(const QString &sectionName)
+{
+    m_sectionName = sectionName;
+}
+
```
rainemak, line 130: While you're at this please add link->date() to the qdebug of the Link above.

src/pages/components/HistoryPage.qml
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        delegate: HistoryItem { search: view.search }
+
+        ViewPlaceholder {
+            //% "Websites you visited show up here"
```
jpetrell, line 39: Maybe drop past tense "Websites you visit show up here"

src/storage/link.cpp
```diff
@@ -73,14 +95,39 @@ bool Link::isValid() const
...
+{
+    return m_date;
+}
+
+void Link::setDate(const QDateTime &date)
+{
+    m_date = date;
+}
+
+QString Link::sectionName() const
```
rainemak, line 121: Date is enough from model section point of view.

src/pages/components/HistoryItem.qml
```diff
@@ -0,0 +1,106 @@
...
+    Column {
+        id: rightColumn
+        width: dateLabel.implicitWidth
+        anchors {
+            right: deleteButton.left
+            rightMargin: Theme.paddingSmall
+        }
+        HistoryLabel {
+            id: dateLabel
+            text: Format.formatDate(model.date, Formatter.DateMediumWithoutYear)
```
rainemak, line 50: This is not at design. Only time and time should be following platform current clock.

Section header will depicts relative time view => "Today", "Yesterday", "Last week", and so on.

src/pages/components/HistoryItem.qml
```diff
@@ -0,0 +1,106 @@
...
+    function openLink() {
+        webView.load(model.url, model.title)
+        pageStack.pop()
+    }
+
+    width: view.width
+    menu: contextMenuComponent
+
+    ListView.onAdd: AddAnimation { target: root }
+
```
rainemak, line 23: Icon missing.

Icon should have... Horizontal page margin at the left and Theme.paddingMedium between icon and those two rows.

src/pages/components/HistoryItem.qml
rainemak: Discussed this with design and this clearly needs more thinking.

1) Transition from non-private to private whilst opening a tab
2) When transition happens people will wonder where did my tabs went
3) No new tabs created... always replacing previously opened page => based on the menu item text I would assume new tabs.

src/pages/components/HistoryPage.qml
atatarov: If I change to data, then each element of the model will be in a own section, so every date will be unique.

atatarov: If I change to data, then each element of the model will be in own section, so each date will be unique

src/storage/link.cpp
atatarov: Unique date will not merge sections elements - Today, Yesterday, Month

src/storage/link.cpp
atatarov: qml side doesn't allow me to determine what the section will be

monich: I would check how it's done in the Messages app. Messages are grouped by date quite reasonably.

atatarov: I haven't found the option I needed in the message app. There, the role for the section property is grouped for 1 day. I need to group by month, and today and yesterday

monich: Right, grouping there has recently changed. Before, it was more like what you're doing and looked more reasonable to me. Oh well...

atatarov: `

        // Return the date for the past week, and group all older items together under an
        // arbitrary older date
        const int daysDiff = QDate::currentDate().toJulianDay() - dateTime.date().toJulianDay();
        if (daysDiff < 7)
            return dateTime.date();

        // Arbitrary static date for older items..
        return QDate(2000, 1, 1);
`
In this case, 8 sections are returned to the model - the first 7 days, and everything after. For a browser design, this option will not work, you need to group it like this: Today, Yesterday, September 2020, August 2020, etc...

atatarov: For my case, I don't see other way to store role for the section group

src/pages/components/HistoryItem.qml
atatarov: At the moment I don't know how to store icons for history. Maybe add it in next PR..

src/pages/components/HistoryItem.qml
```diff
@@ -0,0 +1,91 @@
...
+                //% "Copy to clipboard"
+                text: qsTrId("sailfish_browser-me-copy-to-clipboard")
+                onClicked: Clipboard.text = model.url
+            }
+
+            MenuItem {
+                //: Delete history entry
+                //% "Delete"
+                text: qsTrId("sailfish_browser-me-delete")
+                onClicked: remove()
```
rainemak, line 87: Maybe this should have a remorse item. Will think about it but let's handle that separately as existing history view didn't have remorse either.

src/pages/components/SecondaryBar.qml
```diff
@@ -41,6 +41,21 @@ Column {
...
     }
 
+    OverlayListItem {
+        height: overlay.toolBar.toolsHeight
+
+        iconWidth: root.iconWidth
+        horizontalOffset: root.horizontalOffset
+        //% "History"
+        text: qsTrId("sailfish_browser-la-history")
+        iconSource: "image://theme/icon-m-backup"
```
rainemak, line 51: Lemme check this one as well. Would be better to have the icon duplicated so that we'd have something that would depict history/time i.e. would not refer to the backup.

src/pages/components/HistoryItem.qml
```diff
@@ -0,0 +1,91 @@
...
+ListItem {
+    id: root
+
+    property string search
+
+    function remove() {
+        view.model.remove(model.index)
+    }
+
+    function openLink() {
```
rainemak, line 14: I'd not expect opening to a new tab. Open in a new tab could be in the context menu.

This should be imo loading to the active tab by default. Lemme check.

src/pages/components/HistoryItem.qml
```diff
@@ -0,0 +1,65 @@
+/****************************************************************************
+**
+** Copyright (C) 2020 Open Mobile Platform LLC
```
rainemak, line 3: Lower case c is preferred. Noticed that other blocks have capital C but let's use lower case for new ones.

rainemak, line 3:
```suggestion
** Copyright (c) 2020 Open Mobile Platform LLC.
```

src/pages/components/HistoryPage.qml
```diff
@@ -0,0 +1,68 @@
+/****************************************************************************
+**
+** Copyright (C) 2020 Open Mobile Platform LLC
```
rainemak, line 3:
```suggestion
** Copyright (c) 2020 Open Mobile Platform LLC.
```

src/pages/components/SecondaryBar.qml
```diff
@@ -41,6 +41,22 @@ Column {
...
```
rainemak, line 51:
```suggestion
        iconSource: "image://theme/icon-m-history"
```

src/pages/components/HistoryItem.qml
atatarov: What about it? Should I add a remorse item?

rainemak: Not now. Could file a task about it.

rainemak:
LGTM

PR #697, opened 2020-08-26
[sailfish-browser] Add button to save page as PDF file. Contributes to JB#50923

paveltkv:


src/pages/components/SecondaryBar.qml
```diff
@@ -49,7 +49,14 @@ Item {
         // Spacer for pushing Search, Favorite, Share, Downloads to the right hand side
         Item {
             height: parent.height
-            width: parent.width - addTabButton.width - forwardButton.width - midIconWidth * 3 - downloadsButton.width
+            width: parent.width - addTabButton.width - forwardButton.width - midIconWidth * 4 - downloadsButton.width
+        }
+
+        Browser.IconButton {
```
rainemak, line 55: Checked the design and this could located between sharing and downloads icons so that we'd have similar kind of actions grouped together.

src/pages/components/Overlay.qml
```diff
@@ -486,6 +487,27 @@ Background {
                         pageStack.animatorPush("com.jolla.settings.system.CertificateDetailsPage", {"website": webView.security.subjectDisplayName, "details": webView.security.serverCertDetails})
                     }
                 }
+                onSavePageAsPDF: {
+                    var filename = "unnamed_file"
+                    if (webView.title && webView.title.length !== 0) {
+                        filename = webView.title
+                    } else {
+                        // drop query string from URL and split to sections
+                        var urlSections = webView.url.split("?")[0].split("/")
```
rainemak, line 496: WebUtils.displayableUrl(url) does conceptually the same. It's ok to move that WebUtils.displayableUrl to the declarativewebpage as a property and refer here as "webView.contentItem.displayableUrl"

src/pages/components/Overlay.qml
```diff
@@ -486,6 +487,27 @@ Background {
...
+                        var leafName = urlSections[urlSections.length - 1]
+                        if (leafName.length !== 0) {
+                            filename = leafName
+                        }
+                    }
+
+                    filename = filename.replace(/[/\\?%*:|"<>]/g, '-').replace(/ /g, '_') + ".pdf"
+                    var targetUrl = DownloadHelper.createUniqueFileUrl(filename, StandardPaths.download)
```
rainemak, line 504: filename.replace(/[/\\?%*:|"<>]/g, '-').replace(/ /g, '_') => should be added/moved to the createUniqueFileUrl itself.

If filename has already extension this would concatenate 2nd .pdf e.g. via title "myExample.pdf"

src/pages/components/SecondaryBar.qml
```diff
@@ -49,7 +49,14 @@ Item {
...
+            width: midIconWidth
+            icon.source: "image://theme/icon-m-file-download-as-pdf"
+            active: webView.contentItem && !webView.loading
```
rainemak, line 58: active: webView.contentItem &&  webView.contentItem.active && !webView.contentItem.loading

I'd assume that contentItem must be active. Maybe loading guard could be ignored -- just wondering case where loading doesn't complete or/and there's already content in the view.

src/pages/components/Overlay.qml
paveltkv: WebUtils.displayableUrl(url) returns host instead of filename. For example, if pass http://xxxx.xxx/aaa/bbb.html?zzz, it will return xxxx.xxx instead of bbb.html.

rainemak (approved):
LGTM :+1:

PR #698, opened 2020-08-27
[browser] Add bookmark page. Contributes to JB#50676

atatarov:


src/pages/components/BookmarkItem.qml
```diff
@@ -0,0 +1,116 @@
...
+            left: parent.left
+            leftMargin: Theme.horizontalPageMargin
+        }
+
+        icon: model.favicon
+
+        sourceSize.width: Theme.iconSizeSmall
+        sourceSize.height: Theme.iconSizeSmall
+        width: Theme.iconSizeSmall
+        height: Theme.iconSizeSmall
```
rainemak, line 32: Imo this is too small also in the design drafts.

src/pages/components/BookmarkItem.qml
```diff
@@ -0,0 +1,116 @@
...
+
+ListItem {
+    id: root
+
+    function openLink() {
+        webView.load(model.url, model.title)
+        pageStack.pop()
+    }
+
+    contentHeight: titleText.height * 2 + Theme.paddingMedium
```
rainemak, line 13: More or less same as https://github.com/sailfishos/sailfish-browser/pull/696/files#r482034311

src/pages/components/BookmarkPage.qml
```diff
@@ -0,0 +1,57 @@
+import QtQuick 2.0
+import Sailfish.Silica 1.0
+import Sailfish.Browser 1.0
+
+Page {
+
```
rainemak, line 6: Extra empty line.

src/bookmarks/bookmarkfiltermodel.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+bool BookmarkFilterModel::filterAcceptsRow(int sourceRow, const QModelIndex &sourceParent) const
+{
+    bool contains = false;
+    QModelIndex index = sourceModel()->index(sourceRow, 0, sourceParent);
+
+    for (int i = 0; i < sourceModel()->roleNames().count(); i++) {
+        int it = i + static_cast<int>(Qt::UserRole);
+        QString roleName = sourceModel()->roleNames().value(it);
+        if (roleName == "url" || roleName == "title") {
```
okodron, line 24: I would prefer this way:
```suggestion
        if (roleName == QLatin1String("url") || roleName == QLatin1String("title")) {
```
Also I think it's possible to search once for these two role indexes by their names and then just check data only for these two found role values.

src/bookmarks/bookmarkfiltermodel.cpp
```diff
@@ -0,0 +1,30 @@
...
+            if (sourceModel()->data(index, it).toString().trimmed().contains(filterRegExp().pattern(), Qt::CaseInsensitive))
```
okodron, line 25: Why `filterRegExp().pattern()`?
By the way, does it mean that user has provided the search string as a regexp pattern string (instead of the wildcard one)?

src/bookmarks/bookmarkfiltermodel.cpp
```diff
@@ -0,0 +1,30 @@
...
+                contains = true;
```
okodron, line 26: Maybe just:
```suggestion
                return true;
```

okodron (changes requested):

src/bookmarks/bookmarkfiltermodel.cpp
monich: Since role ids are fixed at compile time, they could be simply queried directly from the source model. If this filter model is supposed to work with any source model that has `"url"` and `"title"` roles (which I doubt), then indices should be queried when the model is being set. But that's when performance matters. In case of browser it obviously doesn't :)

src/pages/components/BookmarkPage.qml
```diff
@@ -0,0 +1,57 @@
...
+    property string search
+
+    BookmarkModel { id: bookmarkModel }
```
monich, line 9: Would it make sense to make `BookmarkModel` a singleton? Maybe not, if under normal circumstances no instances of this model gets allocated. But if at least one instance always exists, making it a singleton might be a good idea from the performance prospective.

okodron (changes requested):

src/pages/components/SecondaryBar.qml
```diff
@@ -26,75 +26,90 @@ Item {
...
-            onTapped: enterNewTabUrl()
-        }
 
-        Browser.ExpandingButton {
-            id: forwardButton
-            expandedWidth: iconWidth
-            icon.source: "image://theme/icon-m-forward"
-            active: webView.canGoForward
-            onTapped: webView.goForward()
+        SecondaryBarItem {
```
rainemak, line 32: Maybe OverlayListItem instead. The "Secondary" part of above is a bit misleading as this is kind of third row.

src/pages/components/OverlayListItem.qml
```diff
@@ -0,0 +1,56 @@
...
+        Icon {
+            id: icon
+
+            anchors {
+                centerIn: parent
+                horizontalCenterOffset: horizontalOffset
+            }
+            highlighted: root.highlighted
+            color: Theme.primaryColor
+        }
```
rainemak, line 39: As label is anchored to the right of the iconContainer (reserving the space for the icon) it is cleaner to close icoContainer  here.

src/pages/components/OverlayListItem.qml
```diff
@@ -0,0 +1,56 @@
...
+                right: parent.right
+                rightMargin: Theme.horizontalPageMargin
+                verticalCenter: parent.verticalCenter
+            }
+            color: root.highlighted ? Theme.highlightColor : Theme.primaryColor
+            truncationMode: TruncationMode.Fade
+
+            Behavior on opacity { FadeAnimation {} }
+        }
+    }
```
rainemak, line 55: After above remove one bracket here and fix indentation.

src/pages/components/BookmarkItem.qml
```diff
@@ -0,0 +1,119 @@
...
+            truncationMode: TruncationMode.Fade
+            width: parent.width
+        }
+
+        Label {
+            text: Theme.highlightText(title, searchText, Theme.highlightColor)
+            textFormat: Text.StyledText
+            color: highlighted ? Theme.highlightColor : Theme.primaryColor
+            font.pixelSize: Theme.fontSizeSmall
+            wrapMode: Text.WordWrap
```
rainemak, line 71: Let's start this with fading as well so that we have a fixed list item height.

src/pages/components/BookmarkItem.qml
```diff
@@ -0,0 +1,119 @@
...
+            left: favoriteIcon.right
+            leftMargin: Theme.paddingMedium
+            right: parent.right
+            rightMargin: Theme.horizontalPageMargin
+        }
+
+        Label {
+            text: Theme.highlightText(url, searchText, Theme.highlightColor)
+            textFormat: Text.StyledText
+            opacity: Theme.opacityHigh
```
rainemak, line 59: This looks odd to me as the "primary" text line is now a bit dimmed and so called "description label" has the primary text's color. Let's have it like this for now and I'll sync with design still.

Thanks @atatarov !

PR #709, opened 2020-09-28
[sailfish-browser] User agent overrides update for esr60. Fixes JB#50980

mkenttala:


rainemak (approved):
LGTM. Could you log a task about checking needed overrides.

PR #710, opened 2020-09-30
[sailfish-browser] Add descriptions for select test case. JB#51225

rainemak:


tests/manual/testselect.html
```diff
@@ -11,6 +11,7 @@
     <title>TestSelect</title>
   </head>
   <body>
+    <h1>Single select without group headers<h1>
```
llewelld, line 14: Should be `</h1>` to close in all cases, I think.

llewelld (changes requested):
The headings are clear, but just the end tags need correcting.

tests/manual/testselect.html
rainemak: Sure... too quickie :+1:

llewelld (approved):
Perfect, thanks :+1:

PR #712, opened 2020-10-01
Replace dead link in readme with a copy from the wayback machine

tekktonic:
Just looked at the repository and the referenced blog no longer exists. I couldn't find it anywhere else so I linked it via the wayback machine.


llewelld:
A good observation, and linking to the wayback machine seems like a decent stop-gap solution.

In the slightly longer term it would be worth moving the content (and updating it at the same time) to the browser wiki and linking to that instead. It would also be good to get some links in to the docs on the developer wiki:

https://sailfishos.org/wiki/Browser
https://sailfishos.org/wiki/WorkingWithBrowser

PR #719, opened 2020-10-14
Add a CSS test page

llewelld:
The page just has a couple of tests:

1. CSS font tests.
2. CSS shapes (since the features straddle ESR60 and ESR62).

Probably more CSS tests should be added in the future.


rainemak:
LGTM

PR #722, opened 2020-10-20
[browser] Show toolbar properly when entering back to browser page. Fixes JB#45651

rainemak:
In case the contentItem does not exist, the fullscreen mode cannot be
true either. Fix showing up toolbar accordingly.


llewelld:
LGTM.

PR #724, opened 2020-10-25
[sailfish-browser] Use remorse for item deletion in HistoryList. Contributes to JB#51702

atatarov:


src/pages/components/HistoryList.qml
```diff
@@ -28,6 +28,12 @@ SilicaListView {
         search: view.search
         showDeleteButton: view.showDeleteButton
 
+        function remove() {
+            remorseDelete(function() {
+                view.model.remove(model.index)
```
rainemak, line 33: History page's remove button is not using this. Please adjust that as well.

Further, would be better to change history model.remove to be model.url based as that is unique. This way async db && model.index should not be able cause delete for a wrong index. That said, you could store function local variable of the "model.url" in context of the remove function.

src/pages/components/HistoryItem.qml
```diff
@@ -18,6 +18,12 @@ ListItem {
     property string search
     property bool showDeleteButton
 
+    function remove() {
+        remorseDelete(function() {
+            view.model.remove(model.url)
```
rainemak, line 23: You should fix the historymodel accordingly and all the calls the model.

src/pages/components/HistoryItem.qml
```diff
@@ -18,6 +18,12 @@ ListItem {
     property string search
     property bool showDeleteButton
 
+    function remove() {
```
rainemak, line 21: var url = model.url

src/pages/components/HistoryItem.qml
rainemak, line 23: Here refer to the function local url variable.

src/history/declarativehistorymodel.cpp
```diff
@@ -54,6 +54,18 @@ void DeclarativeHistoryModel::remove(int index)
     emit countChanged();
 }
 
+void DeclarativeHistoryModel::remove(const QString &url)
+{
+    int index = 0;
+    for (const auto &link: m_links) {
```
rainemak, line 60:
```suggestion
    for (const auto &link : m_links) {
```
A space missing before colon.

rainemak (approved):
Approved with changes.

PR #725, opened 2020-10-25
[sailfish-browser] Add page for site permissions. Contributes to JB#50927

atatarov:


src/pages/components/SitePermissionPage.qml
```diff
@@ -0,0 +1,85 @@
...
+                    text: qsTrId("sailfish_browser-me-allow")
+                    onClicked: PermissionManager.add(url, model.type, PermissionManager.Allow)
+                }
+                MenuItem {
+                    //% "Block"
+                    text: qsTrId("sailfish_browser-me-block")
+                    onClicked: PermissionManager.add(url, model.type, PermissionManager.Deny)
+                }
+                MenuItem {
+                    //% "Global (Ask to allow)"
```
rainemak, line 78:
```suggestion
                    //% "Always ask"
```

src/pages/components/SitePermissionPage.qml
```diff
@@ -0,0 +1,85 @@
...
+                    text: qsTrId("sailfish_browser-me-global")
```
rainemak, line 79:
```suggestion
                    text: qsTrId("sailfish_browser-me-always_ask")
```

src/pages/components/SitePermissionPage.qml
```diff
@@ -0,0 +1,85 @@
...
+                    if (type === "geolocation") {
+                        permissionTypesModel.append({
+                                                        //% "Location"
+                                                        title: qsTrId("sailfish_browser-ti-location"),
+                                                        permission: parseInt(permissions[i].capability),
+                                                        type: type
+                                                    })
+                    } else if (type === "popup") {
+                        permissionTypesModel.append({
+                                                        //% "Pop-up windows"
```
rainemak, line 26:
```suggestion
                                                        //% "Pop-up"
```
I'd not use term "windows" as these are not windows. Further, "Pop-up" is quite well understood in context of web. Comments?

src/pages/components/SitePermissionPage.qml
```diff
@@ -0,0 +1,106 @@
+/****************************************************************************
```
rainemak, line 1: As this is a page let's have it under pages directory.

rainemak:
This should be e2e integrated. We should drop "Cookies", and "Popups" until we have means to control those via respected dialogs leaving only "Location" here which is rather clumsy but that is what we have now.

Tested location and it seems to work (Geo location permission prompt && site permission combination).

atatarov:
> This should be e2e integrated. We should drop "Always ask" from "Cookies", and "Popups" until we have means to control those via respected dialogs leaving only "Location" here which is rather clumsy but that is what we have now.
> 
> Tested location and it seems to work (Geo location permission prompt && site permission combination).

atatarov:
> This should be e2e integrated. We should drop "Always ask" from "Cookies", and "Popups" until we have means to control those via respected dialogs leaving only "Location" here which is rather clumsy but that is what we have now.
> 
> Tested location and it seems to work (Geo location permission prompt && site permission combination).

Still working on this pr. Add a button to connections security and add indication of permissions for the site

rainemak (commented):
This is not yet integrated to the user interface. So, with that this maybe could be approved.

However, with the Cookie test case and disabling cookies with this cookies still keep on working.
https://browser.sailfishos.org/tests/testcookies.html

```
                    pageStack.animatorPush(Qt.resolvedUrl("../SitePermissionPage.qml"), {
                                            "title": "Test",
                                               "url": "https://browser.sailfishos.org"
                                           })
```
Currently as far as see implementation, taking into account embedlite-component, is lacking network.cookie.* preference handling for instance.

When it comes to dom/base/Navigator.cpp it checked cookie permissions via nsICookiePermission (nsICookiePermission.idl). Here is one example usage gecko-dev/browser/base/content/test/general/browser_bug578534.js

If we have this kind of Cookie setting it should really work as well.

Am I missing something wrong the big picture?

rainemak (commented):
This is a starting point. More site specific permissions will be added later. Also integration to the user interface missing. That said, more things can be implemented on top of this.

src/pages/SitePermissionPage.qml
```diff
@@ -0,0 +1,94 @@
...
+
+import QtQuick 2.6
+import Sailfish.Silica 1.0
+import Sailfish.WebView.Controls 1.0
+
+Page {
+    id: page
+
+    property string title
+    property string url
```
llewelld, line 19: Could these be aliases, rather than properties? If so, the `id: page` can be dropped.

src/pages/SitePermissionPage.qml
```diff
@@ -0,0 +1,94 @@
...
+        case PermissionManager.Allow:
+            return 0 // index for "Allow" menu
+        case PermissionManager.Deny:
+            return 1 // index for "Deny" menu
+        default:
+            return 2 // index for "Always ask" menu
+        }
+    }
+
+    ListModel {
```
llewelld, line 38: It looks like you have similar requirements with this model as you have in your #728 PR. Given this, it would seem sensible to create a reusable "site permission model" (or using `PermissionModel` or `PermissionFilterProxyModel` if they can be restricted to a single site).

src/pages/SitePermissionPage.qml
```diff
@@ -0,0 +1,94 @@
...
+    property string title
+    property string url
+
+    property var _permissions: {
+        "geolocation": PermissionManager.Prompt
+    }
+
+    signal requestFinished()
+
+    function getCurrentIndex(type) {
```
llewelld, line 27: The title `getCurrentIndex` causes some confusion for me, because I'm expecting it to get the index from the combobox. Perhaps this is a matter of taste, but for me something like `convertTypeToIndex` would be clearer.

src/pages/SitePermissionPage.qml
```diff
@@ -0,0 +1,94 @@
...
+
+        model: permissionTypesModel
+
+        delegate: ComboBox {
+            label: model.title
+
+            currentIndex: getCurrentIndex(model.type)
+
+            Component.onCompleted: requestFinished.connect(function() {
+                currentIndex = getCurrentIndex(model.type)
```
llewelld, line 72: I didn't test this, but is it necessary to bind this connection only `onCompleted`, rather than simply creating an `onRequestFinished` handler, or even calling it directly in the `requestUriPermission` response?

src/pages/SitePermissionPage.qml
```diff
@@ -0,0 +1,94 @@
...
+            title: page.title
+            description: page.url
+        }
+
+        model: permissionTypesModel
+
+        delegate: ComboBox {
+            label: model.title
+
+            currentIndex: getCurrentIndex(model.type)
```
llewelld, line 69: It's not clear to me what this line is doing. Won't this get immediately overwritten by line 72?

llewelld (changes requested):

src/pages/SitePermissionPage.qml
atatarov: I don't have a watcher for site permissions to follow automatically. I can ask them when I open a page or click on a button. Their logic also differs. in this pr I show all possible settings, even if they are not in the exclusion list. 
I have not found a way to unify these three things:
1. A list of exceptions for a permission type. For example geolocation list, I get it to use PermissionProxyModel. PermissionModel will filter temporary permission. Here we only store permanent permissions
2. Site permission indication.  I should show all permissions for a site, both temporary and permanent. I need to store icons in this model
3. Site permission page. I should show a static model. Also have qstrId() for each item. I showi all the possible permission settings for this site. Even if they are not. But I should match the data of the set permissions with this model

llewelld: For the case of the icon sources and the qstrId()s, I don't think these should be stored in the model, because they can be derived from elements in the model. You can therefore perform the transformation in the delegate.

src/pages/SitePermissionPage.qml
atatarov: I don't know how to do it better. Although the model is static and the indices are known, I try to avoid hardcode numbers. Do you mean, like this?
```
                permissionTypesModel.setProperty(0, "geolocation", perms[i].capability)
                permissionTypesModel.setProperty(1, "cookie", perms[i].capability) - in future
```
                
                or 
```
              permissionTypesModel.setProperty(typeToIndexModel(type), "geolocation", perms[i].capability)
```
function typeToIndexModel(type) {
case "geolocation"
return 0; // index for Location item in permissionTypesModel
}

src/pages/SitePermissionPage.qml
atatarov: OK. I don't think it will be difficult and will take a long time to change the PermissionModel

atatarov: c ++ model removes a lot of problems with ListModel

src/pages/SitePermissionPage.qml
llewelld: Ah, no, I don't have any issue with the way you implemented `typeToIndexModel`. Here I was just wondering whether you can do something like this:
```
onRequestFinished: currentIndex = getCurrentIndex(model.type)
```

src/pages/SitePermissionPage.qml
atatarov: it used in some  places

llewelld (commented):
Just some small comments/questions.

src/pages/SitePermissionPage.qml
```diff
@@ -0,0 +1,112 @@
...
+                MenuItem {
+                    //% "Block"
+                    text: qsTrId("sailfish_browser-me-block")
+                    onClicked: PermissionManager.add(url, model.type, PermissionManager.Deny)
+                }
+                MenuItem {
+                    //% "Always ask"
+                    text: qsTrId("sailfish_browser-me-always_ask")
+                    onClicked: PermissionManager.add(url, model.type, PermissionManager.Prompt)
+                    visible: model.type !== "popup" && model.type !== "cookie"
```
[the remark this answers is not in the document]
atatarov: popup and cookie don't have the ability to query. They can be deny/allow only

src/pages/SitePermissionPage.qml
```diff
@@ -0,0 +1,112 @@
...
+
+    property string title
+    property string url
+    property PermissionModel permissionModel
+
+    property var permissions: {
+        "geolocation": PermissionManager.Prompt,
+        "popup": PermissionManager.Deny,
+        "cookie": PermissionManager.Allow
+    }
```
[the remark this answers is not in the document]
atatarov: My missing.. It should take it from WenEngineSettings.cookieBehavour and WebEngineSettings.popupEnabled

src/pages/SitePermissionPage.qml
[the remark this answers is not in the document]
llewelld: Okay, thanks. That makes sense then.

src/pages/SitePermissionPage.qml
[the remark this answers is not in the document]
llewelld: Thanks, looking good now.

llewelld (approved):
Thanks. This LGTM now.

PR #728, opened 2020-10-29
[sailfish-browser] Add site permission indication and update CertificateInfo UI. Contributes to JB#50930

atatarov:


shanechko (approved):
LGTM

src/pages/components/CertificateInfo.qml
```diff
@@ -12,63 +12,61 @@ import Sailfish.Silica.private 1.0 as Private
...
+        permissionIndicationModel.clear()
+
+        PermissionManager.requestUriPermissions(toolBarRow.url, function(permissions) {
+            for (var i in permissions) {
+                var type = permissions[i].type
+                var capability = parseInt(permissions[i].capability)
+                if (capability === PermissionManager.Prompt || capability === PermissionManager.Unknown)
+                    continue
+
+                var source = "image://theme/icon-m-location"
```
rainemak, line 38: You should colorize this like "image://theme/icon-m-location?#FF0000" <= red being and example. Icons must not be colorized by default rather colorizing shall be done on the software.

src/pages/components/CertificateInfo.qml
```diff
@@ -103,146 +105,69 @@ SilicaFlickable {
...
         }
 
+        DetailItem {
+            //: Label for the issuer field of a TLS certificate
+            //% "Verified by"
+            label: qsTrId("sailfish_browser-la-cert_issuer")
+            value: security ? security.issuerDisplayName : ""
+            visible: value && value.length > 0
+        }
+
```
llewelld, line 115: In the figma designs I also see "Connection details" info shown. This is removed later in some of the screenshots, but I don't see any discussion about whether it should be kept/removed on the bug.

Should a "Connection details" line be included here?

src/pages/components/CertificateInfo.qml
```diff
@@ -103,146 +105,69 @@ SilicaFlickable {
...
-            key: qsTrId("sailfish_browser-la-cert_valid_dates")
-            value: insecure ? Format.formatDate(security.effectiveDate, Formatter.DateMedium)
-                            + " — "
-                            + Format.formatDate(security.expiryDate, Formatter.DateMedium)
-                          : ""
-            insecure: security && !security.certIsNull && security.notValidAtThisTime
-            keyWidth: root.keyWidth
-        }
+        Column {
+            width: parent.width
```
llewelld, line 137: This looks cleaner if the column has the same spacing as the parent column (i.e. `Theme.paddingMedium`). If the ideas is to group the "Current permissions" title and the permission icons together, then I think extra space should be added above the title.

llewelld (changes requested):
Generally I'd say this is looking nice and working well. I've added some suggestions.

src/pages/components/CertificateInfo.qml
```diff
@@ -12,63 +12,62 @@ import Sailfish.Silica.private 1.0 as Private
...
+                                         })
     }
 
     onSecurityChanged: {
         // Jump back to the top
         contentY = originY
     }
 
-    onMaximumKeyWidthChanged: updateKeyWidth()
+    ListModel {
```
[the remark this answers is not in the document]
atatarov: I can fill the PermissionModel outside, but there are roles that are not there - title, source. 
For a regular model (list of exceptions), these roles are not needed and it is not clear how to fill them. I don't have a specific position here and can change sailfish-components-webview if you say

src/pages/components/CertificateInfo.qml
```diff
@@ -103,146 +105,69 @@ SilicaFlickable {
...
-            value: security ? security.issuerDisplayName : ""
-            keyWidth: root.keyWidth
+        Button {
+            anchors.horizontalCenter: parent.horizontalCenter
+            //% "Certificate details"
+            text: qsTrId("sailfish_browser-sh-certificate_details")
+            enabled: security && !security.certIsNull
+            visible: enabled
+            onClicked: showCertDetail()
         }
```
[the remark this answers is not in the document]
atatarov: Yes it will. I'm waiting this PR
https://github.com/sailfishos/sailfish-browser/pull/725

src/pages/components/CertificateInfo.qml
atatarov: Yes, they want to include it. But it appeared after I started this work, and still has no agreement

src/pages/components/CertificateInfo.qml
```diff
@@ -103,146 +105,69 @@ SilicaFlickable {
...
+                rows: permissionIndicationModel.count > 5 ? 1 : 2
+                columns: permissionIndicationModel.count
+                anchors.horizontalCenter: parent.horizontalCenter
+                spacing: Theme.paddingMedium
+                Repeater {
+                    model: permissionIndicationModel
+                    delegate: Icon {
+                        source: model.iconSource
+                        color: Theme.highlightColor
+                        Icon {
```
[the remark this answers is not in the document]
atatarov: There is a reason why I could not do this. When opening site permissions, the user sees the settings for possible permissions, even if they are not in the list, he can manage them. For example, it can deny permission for cookies, but by default it will not be displayed in the indication of current permissions.

src/pages/components/CertificateInfo.qml
```diff
@@ -103,146 +105,69 @@ SilicaFlickable {
             }
         }
 
+        DetailItem {
+            //: Label for the issuer field of a TLS certificate
+            //% "Verified by"
+            label: qsTrId("sailfish_browser-la-cert_issuer")
+            value: security ? security.issuerDisplayName : ""
+            visible: value && value.length > 0
+        }
```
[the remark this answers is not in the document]
atatarov: To use wrap and bind in the center this component regardless of the width, I need to take the contentWidth of the text, but with this I get a binding loop.

src/pages/components/CertificateInfo.qml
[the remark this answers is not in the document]
llewelld: Yes, I can see what you mean. You could have the icons clickable through to the site permissions dialogue, and only show the "Site permissions" button if there are none to click?

src/pages/components/CertificateInfo.qml
llewelld: Any update on this?

atatarov: https://github.com/sailfishos/sailfish-browser/pull/725
still watinig...

src/pages/components/CertificateInfo.qml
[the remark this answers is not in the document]
llewelld: I'm not sure about the title as I don't see it in the code, but it seems to me that the source is derivable from the  `type` and `capability`, so needn't be in the model at all (i.e. move the logic on lines 41-51 into the delegate). I think that could be quite nice; what do you think?

I agree that changing sailfish-components-webview doesn't seem like a good idea, so my suggestion probably isn't the right approach if that would be necessary.

src/pages/components/CertificateInfo.qml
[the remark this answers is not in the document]
llewelld: Can you use `implicitWidth` rather than `width`? In other words, if `implicitWidth > parent.width - label.width` then use the full width of the parent, otherwise centre the items based on `implicitWidth + label.width`?

I'm simplifying a lot here, but the `implicitWidth` shouldn't depend on the wrapping, so shouldn't create a binding loop.

src/pages/components/CertificateInfo.qml
```diff
@@ -103,146 +105,69 @@ SilicaFlickable {
...
+                        color: Theme.highlightColor
+                        Icon {
+                            anchors {
+                                bottom: parent.bottom
+                                right: parent.right
+                            }
+                            source: "image://theme/icon-s-blocked"
+                            color: Theme.errorColor
+                            visible: model.permission === PermissionManager.Deny
+                        }
```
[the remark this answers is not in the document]
llewelld: I don't see this jump any more, so I'm resolving this comment.

src/pages/components/CertificateInfo.qml
```diff
@@ -103,146 +79,103 @@ SilicaFlickable {
...
 
-        CertificateLabel {
-            //: Label for the subject field of a TLS certificate
-            //% "Website"
-            key: qsTrId("sailfish_browser-la-cert_subject")
-            value: security ? security.subjectDisplayName : ""
-            keyWidth: root.keyWidth
+        Button {
+            anchors.horizontalCenter: parent.horizontalCenter
+            //% Details for certificate info
```
llewelld, line 104: I think this should be a `//:` prefix.

src/pages/components/CertificateInfo.qml
```diff
@@ -103,146 +79,103 @@ SilicaFlickable {
...
 
-        CertificateLabel {
-            //: Label for the issuer field of a TLS certificate
-            //% "Issuer"
-            key: qsTrId("sailfish_browser-la-cert_issuer")
-            value: security ? security.issuerDisplayName : ""
-            keyWidth: root.keyWidth
+        Button {
+            anchors.horizontalCenter: parent.horizontalCenter
+            //% Manage permission for current site
```
llewelld, line 114: Similarly here, should be `//:` I think.

src/pages/components/CertificateInfo.qml
```diff
@@ -103,146 +79,103 @@ SilicaFlickable {
...
+                            source: {
+                                var blocked = (model.capability === PermissionManager.Deny)
+                                if (model.type === "geolocation") {
+                                    return "image://theme/icon-m-browser-location" + (blocked ? "-template" : "")
+                                } else if (model.type === "cookie") {
+                                    return "image://theme/icon-m-browser-cookies" + (blocked ? "-template" : "")
+                                } else if (model.type === "popup") {
+                                    return "image://theme/icon-m-browser-popup" + (blocked ? "-template" : "")
+                                }
+                                highlighted: permissionArea.pressed
```
llewelld, line 162: Should the `highlighted` be outside the `source` block?

src/pages/components/CertificateInfo.qml
```diff
@@ -103,146 +79,103 @@ SilicaFlickable {
...
+                            }
+                            color: Theme.highlightColor
```
llewelld, line 164: Since these are now pressable, I think the default colour should be used instead (in which case, this line can be removed).

llewelld (changes requested):
This is looking really excellent, but there are just a few typos which I think should be fixed.

llewelld (approved):
LGTM

PR #729, opened 2020-10-29
[sailfish-browser] Add maxRowCount to overlay and renaming toolsHeight to rowHeight. Contributes to JB#51715

atatarov:


okodron (approved):

paveltkv (approved):

rainemak:
This causes odd looking flickering and change to the host that is visible in the toolbar all the time.

Reason for that is (as far as my eye can spot if) when opening secondary layer it's now animating from to bottom... not from from the top of the first row.

rainemak (changes requested):
Let's fix that flickering before merging.

rainemak (changes requested):
Almost there...

Please change CaptivePortalToolBar.qml as well...

```
    readonly property real rowHeight: height
    readonly property int maxRowCount: 1
```
There we have higher row => let's keep to API in sync that OverlayAnimator keeps on working.

src/pages/components/CaptivePortalToolBar.qml
```diff
@@ -16,7 +16,7 @@ Column {
     id: toolBarRow
 
     property string url
-    readonly property real toolsHeight: height
+    readonly property real rowHeight: height
```
rainemak, line 19:
```
readonly property int maxRowCount: 1
```
This is still missing... Please also rebase.

rainemak (commented):
More comments.

rainemak (approved):
:+1:

PR #731, opened 2020-11-04
[sailfish-browser] Hide certinfo when the page is inactive. Fixes JB#51846

atatarov:


src/pages/components/Overlay.qml
```diff
@@ -68,6 +68,11 @@ Background {
     }
 
     function enterNewTabUrl(action) {
```
rainemak, line 70: Getting there... Now I understand what you're after.

This not only about enterNewTabUrl rather all your cases are happening when a page is pushed on top of the BrowserPage.

My proposal is that you following from BrowserPage.qml:

```
@@ -92,6 +92,7 @@ Page {
 
         if (status == PageStatus.Inactive && overlay.visible) {
             overlay.animator.hide()
+            overlay.toolBar.certOverlayActive = false
         }
     }

```

Then for _showInfoOverlay property and rename usage of it like:

```
-                _showInfoOverlay = false
+                toolBar.certOverlayActive = false
```

All error is in the cert overlay of the ToolBar. It already tries release loaded item but it's only based on the visibile change but the if state takes into account certOverlayAction. Hence, it should be like:

```
@@ -88,6 +88,12 @@ Column {
 
     Item {
         id: certOverlay
+
+        readonly property bool canDestroy: !visible && !certOverlayActive
+        onCanDestroyChanged: {
+            if (canDestroy) certOverlayLoader.active = false
+        }
+
         visible: opacity > 0.0 || height > 0.0
         opacity: certOverlayActive ? 1.0 : 0.0
         height: certOverlayHeight
@@ -95,12 +101,6 @@ Column {
 
         Behavior on opacity { FadeAnimation {} }
 
-        onVisibleChanged: {
-            if (!visible && !certOverlayActive) {
-                certOverlayLoader.active = false
-            }
-        }

```

rainemak (approved):
LGTM. Thanks Aleksey.

PR #733, opened 2020-11-10
[sailfish-browser] Adjust web page size to visible area when toolbar appears. Contributes to JB#51962

paveltkv:


src/pages/BrowserPage.qml
```diff
@@ -151,7 +151,7 @@ Page {
         fullscreenHeight: portrait ? Screen.height : Screen.width
         portrait: browserPage.isPortrait
         maxLiveTabCount: 3
-        toolbarHeight: overlay.toolBar.rowHeight
+        toolbarHeight: Math.min(overlay.toolBar.rowHeight, webView.fullscreenHeight - overlay.yPosition)
```
rainemak, line 154: Why this?

paveltkv: When the SecondaryBar is open, user is focused on it. At this time, the user is not working with the page and I think there is no need to compress it to keep important elements in scope.

src/qtmozembed/declarativewebpage.cpp
```diff
@@ -396,37 +379,25 @@ void DeclarativeWebPage::thumbnailReady()
...
 {
-    qCDebug(lcCoreLog) << "WebPage: update view margins, foreground:" << (m_container && !m_container->foreground()) << "throttling:" << (m_marginChangeThrottleTimer > 0) << "moving:" << moving();
+    qCDebug(lcCoreLog) << "WebPage: update view margins, foreground:" << (m_container && !m_container->foreground()) << "moving:" << moving();
 
     // Don't update margins while panning, flicking, pinching, vkb is already open, or
     // margin update is ongoing (throttling).
-    if ((m_container && !m_container->foreground()) || m_marginChangeThrottleTimer > 0 ||
-            moving() || m_virtualKeyboardMargin > 0) {
-        return;
+    if(!force) {
```
atatarov, line 388:
```suggestion
    if (!force) {
```

src/pages/BrowserPage.qml
```diff
@@ -284,6 +284,24 @@ Page {
 
         onEnteringNewTabUrlChanged: window.opaqueBackground = webView.tabModel.waitingForNewTab || enteringNewTabUrl
 
+        // Changing of the animator state changes toolbar position to final before animation starts.
+        // This prevents content flickering by skipping invalid y values.
+        property real yPosition: 0
+        property bool skipYChange: false
```
atatarov, line 290:
```suggestion
        property bool skipYChange
```
false by default

apps/qtmozembed/declarativewebpage.cpp
```diff
@@ -86,11 +85,10 @@ DeclarativeWebPage::DeclarativeWebPage(QObject *parent)
             this, &DeclarativeWebPage::onRecvAsyncMessage);
     connect(&m_grabWritter, &QFutureWatcher<QString>::finished, this, &DeclarativeWebPage::grabWritten);
     connect(this, &DeclarativeWebPage::urlChanged, this, &DeclarativeWebPage::onUrlChanged);
-    connect(this, &QOpenGLWebPage::contentHeightChanged, this, &DeclarativeWebPage::updateViewMargins);
+    connect(this, &QOpenGLWebPage::virtualKeyboardHeightChanged, this, &DeclarativeWebPage::updateViewMargins);
```
llewelld, line 88: As far as I can tell (using this PR and https://git.sailfishos.org/mer-core/qtmozembed/merge_requests/74 ) the `QOpenGLWebPage::virtualKeyboardHeight` is never changed, and so this signal never triggers. Can this connection be removed?

apps/qtmozembed/declarativewebpage.cpp
```diff
@@ -398,35 +372,8 @@ void DeclarativeWebPage::thumbnailReady()
...
-        if (contentHeight() < threshold) {
-            margins.setBottom(m_toolbarHeight);
-        }
-    }
-
-    // Some content needed so that it makes sense to throttle content height changes.
-    if (contentHeight() > 0) {
-        m_marginChangeThrottleTimer = startTimer(200);
-    }
+    margins.setBottom(qMax(virtualKeyboardHeight(), (int)toolbarHeight()));
```
llewelld, line 376: From what I can tell, `virtualKeyboardHeight()` is always 0, so could this be removed?

apps/qtmozembed/declarativewebpage.cpp
```diff
@@ -273,16 +266,7 @@ void DeclarativeWebPage::setVirtualKeyboardMargin(qreal margin)
...
-            // Only place where we ignore view margins update guards.
-            // It must be allowed to close vkb while content is moving.
-            resetViewMargins();
-        } else {
-            QMargins margins;
-            margins.setBottom(m_virtualKeyboardMargin);
-            qCDebug(lcCoreLog) << "WebPage: set vkb margins:" << m_virtualKeyboardMargin;
-            setMargins(margins);
-        }
+        updateViewMargins();
```
llewelld, line 269: As far as I can tell, this has no effect without #741. If that is the case, I wonder if at this stage it might better to combined the two PRs, to avoid adding code (which has no effect) and then immediately removing it again.

There may be a good reason for it, but looking at this PR in isolation I'm finding it hard to figure out.

apps/qtmozembed/declarativewebpage.cpp
```diff
@@ -398,35 +372,8 @@ void DeclarativeWebPage::thumbnailReady()
...
-        }
-    }
-
-    // Some content needed so that it makes sense to throttle content height changes.
-    if (contentHeight() > 0) {
-        m_marginChangeThrottleTimer = startTimer(200);
-    }
+    margins.setBottom(qMax(virtualKeyboardHeight(), (int)toolbarHeight()));
 
     qCDebug(lcCoreLog) << "WebPage: set margins:" << margins;
```
llewelld, line 378: With the animated toolbar change, this generates a stream of debug output. It's only when `lcCoreLog` logging is enabled, but maybe it's better to remove it anyway? What do you think?

apps/qtmozembed/declarativewebpage.cpp
rainemak: Really? See QOpenGLWebPage::setVirtualKeyboardHeight(int height)

rainemak: Ah, now I understand what you're saying... Please this on top of https://github.com/sailfishos/sailfish-browser/pull/741

apps/captiveportal/qml/pages/CaptivePortalPage.qml
```diff
@@ -116,7 +116,7 @@ Page {
         fullscreenHeight: portrait ? Screen.height : Screen.width
         portrait: browserPage.isPortrait
         maxLiveTabCount: 3
-        toolbarHeight: overlay.toolBar.height
+        toolbarHeight: Math.min(overlay.toolBar.height, webView.fullscreenHeight - overlay.yPosition)
```
rainemak, line 119: Similarly as commented here https://github.com/sailfishos/sailfish-browser/pull/741/files#r640564376 we should not animate these things. Further, toolbarHeight is a fixed value for a reason.

Overall this PR looks too complex for the thing that it is trying achieve.

apps/qtmozembed/declarativewebpage.cpp
```diff
@@ -63,7 +63,6 @@ DeclarativeWebPage::DeclarativeWebPage(QObject *parent)
     , m_fullScreenHeight(0.f)
     , m_toolbarHeight(0.f)
     , m_virtualKeyboardMargin(0.f)
-    , m_marginChangeThrottleTimer(0)
```
rainemak, line 66: Throttling used to play a role and might be still needed but hopefully we can drop it. This needs a bit more thinking.

llewelld:
These changes have been merged into #741, so I suggest closing this one.

paveltkv:
Changes merged into #741.

PR #734, opened 2020-11-10
[user-agent] User agent override for forbes.com. Fixes JB#51732

mkenttala:


rainemak:
Re-direct to next branch and rebase.

data/ua-update.json
```diff
@@ -65,5 +65,6 @@
   "google.at": "Linux#Maemo",
   "twitter.com": "Linux#Android 8.1.0",
   "apkpure.com": "Linux#Android 8.1.0",
-  "zingmp3.vn": "Linux#Android 8.1.0"
+  "zingmp3.vn": "Linux#Android 8.1.0",
+  "forbes.com": "Linux#Android 8.1.0"
```
rainemak, line 69: Btw, would dropping Linux be enough?

Like:
```
"forbes.com": "Linux# "
```

Space after the hash

mkenttala: Nope, needs Android. Works without that "8.1.0" though.

rainemak: Others have that 8.1.0 => so we can be it with this as well.

rainemak (approved):
Looks good.

PR #741, opened 2020-11-16
[sailfish-browser] Improve Browser margin handling. Contributes to JB#51962

paveltkv:


src/qtmozembed/declarativewebpage.cpp
```diff
@@ -263,6 +264,21 @@ void DeclarativeWebPage::setToolbarHeight(qreal toolbarHeight)
     }
 }
 
+bool DeclarativeWebPage::virtualKeyboardOpened() const
+{
+    return m_virtualKeyboardOpened;
+}
+
+void DeclarativeWebPage::setVirtualKeyboardOpened(bool opened)
```
rainemak, line 272: Could you push all this down to the QtMozEmbed.

apps/browser/qml/pages/BrowserPage.qml
```diff
@@ -139,6 +139,9 @@ Page {
             PropertyChanges {
                 target: webView.contentItem
                 virtualKeyboardMargin: virtualKeyboardObserver.panelSize
+                virtualKeyboardHeight: virtualKeyboardObserver.opened
+                                       ? virtualKeyboardObserver.imSize
+                                       : 0
```
rainemak, line 144:
```
virtualKeyboardHeight: virtualKeyboardObserver.panelSize
```
See when condition. Further, given that virtualKeyboardHeight works is it should you should remove virtualKeyboardMargin (setting margin as well) + cleanup related sendVkbOpenCompositionMetrics from browser side and "embedui:vkbOpenCompositionMetrics" handling from the embedlite-components

apps/captiveportal/qml/pages/CaptivePortalPage.qml
```diff
@@ -105,6 +105,9 @@ Page {
             PropertyChanges {
                 target: webView.contentItem
                 virtualKeyboardMargin: virtualKeyboardObserver.panelSize
+                virtualKeyboardHeight: virtualKeyboardObserver.opened
```
rainemak, line 108: Same as above

rainemak (changes requested):
Please cleanup virtualKeyboardMargin and related code that will obsoleted once it is removed.

llewelld (approved):
This looks good to me and works nicely. It's also an important improvement for the browser, so really nice to have.

I think at this stage it might be better to combine #733 and #741 into one (possibly two) commits. Otherwise you have code added/changed that gets immediately removed.

apps/captiveportal/qml/pages/CaptivePortalPage.qml
```diff
@@ -104,7 +104,7 @@ Page {
             when: virtualKeyboardObserver.opened && webView.enabled
             PropertyChanges {
                 target: webView.contentItem
-                virtualKeyboardMargin: virtualKeyboardObserver.panelSize
+                virtualKeyboardHeight: virtualKeyboardObserver.imSize
```
rainemak, line 107: Need to check in more detail but we should not use the value that is animated -- only one single update and that's it.

apps/qtmozembed/declarativewebpage.cpp
```diff
@@ -432,40 +402,6 @@ void DeclarativeWebPage::resetViewMargins()
     setMargins(margins);
 }
```
rainemak, line 404: This doesn't build now...

See DeclarativeWebPage::updateViewMargins() and m_virtualKeyboardMargin. Just change to use virtualKeyboardHeight().

rainemak (changes requested):
See one build error. After that is fixed, let's get this one merged.

apps/captiveportal/qml/pages/CaptivePortalPage.qml
rainemak: This is good.

paveltkv:
Rebased, combined with #733 and removed margin change animation.

llewelld (approved):
This looks good to me, thank you Pavel. In testing it worked nicely, no glitches or obvious performance issues. The code looks good too. Re-approving

rainemak:
> This looks good to me and works nicely. It's also an important improvement for the browser, so really nice to have.
> 
I do agree with this comment. Also worth saying that more testing we get for this, sooner we find the issues and have them fixed.

rainemak (approved):
Looks good now and works well! Thanks @paveltkv

okodron (approved):

PR #742, opened 2020-11-16
[sailfish-browser] Rename favorites to bookmark. Fixes JB#51984

atatarov:
We have no technical differences for bookmark and favorite now. Same model used
However, the separate BookmarkPage is the bookmark for the user.
It is necessary to rename the places where are "favorite" words to "bookmark",
where is it clear. In the future, these will be different models

### DIFF
--- a/src/pages/components/BookmarkEditDialog.qml
+++ b/src/pages/components/BookmarkEditDialog.qml
@@ -23,10 +23,10 @@ UserPrompt {
 
     canAccept: urlField.text && titleField.text && (urlField.text !== url || titleField.text !== title)
 
-    //% "Edit favorite"
-    title: qsTrId("sailfish_browser-he-edit_favorite")
+    //% "Edit"
+    title: qsTrId("sailfish_browser-he-edit")
 
-    //: Save the bookmark/favorite
+    //: Accept text of bookmark edit dialog
     //% "Save"
     acceptText: qsTrId("sailfish_browser-la-accept_edit")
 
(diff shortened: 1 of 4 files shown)


src/pages/components/BookmarkEditDialog.qml
```diff
@@ -23,8 +23,8 @@ UserPrompt {
 
     canAccept: urlField.text && titleField.text && (urlField.text !== url || titleField.text !== title)
 
-    //% "Edit favorite"
-    title: qsTrId("sailfish_browser-he-edit_favorite")
+    //% "Edit"
+    title: qsTrId("sailfish_browser-he-edit")
 
     //: Save the bookmark/favorite
```
rainemak, line 29: This could be also cleaned. Something like "Accept text of bookmark edit dialog" as translation help.

src/pages/components/FavoriteContextMenu.qml
```diff
@@ -45,15 +45,15 @@ Component {
...
         MenuItem {
-            //% "Edit favorite"
-            text: qsTrId("sailfish_browser-me-edit_favorite")
+            //% "Edit"
+            text: qsTrId("sailfish_browser-me-edit")
             onClicked: delegate.editBookmark()
         }
 
         MenuItem {
             //: "Remove favorited / bookmarked web page"
```
rainemak, line 54: "Remove bookmarked web page"

rainemak:
Few minor comments. Please rebase as well.

rainemak (approved):
Thanks.

PR #745, opened 2020-11-19
[sailfish-browser] Move header from page root to SilicaListView. Fixes JB#52084

atatarov:
When panning content search field and bookmarks headers should be moving as well.
Also move HistoryPage and BookmarkPage to "pages" directory.

### DIFF
--- a/src/pages/BookmarkPage.qml
+++ b/src/pages/BookmarkPage.qml
@@ -11,6 +11,7 @@
 import QtQuick 2.0
 import Sailfish.Silica 1.0
 import Sailfish.Browser 1.0
+import "components"
 
 Page {
     property string searchText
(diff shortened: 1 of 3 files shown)


okodron (approved):

src/pages/BookmarkPage.qml
```diff
@@ -23,35 +24,28 @@ Page {
...
-            EnterKey.onClicked: focus = false
-            onTextChanged: searchText = text
-        }
-    }
-
     SilicaListView {
         id: listView
 
+        anchors.fill: parent
         clip: true
```
rainemak, line 31: No longer needed.

rainemak (changes requested):
Added a minor note about unnecessary clipping that should be dropped now.

rainemak (approved):
Thanks for fixing.

PR #751, opened 2020-11-25
[browser] Move data to new location. JB#51740

rainemak:
Move browser and captive portal profiles to .cache/org.sailfishos/sailfish-browser
and .cache/org.sailfishos/sailfish-browser-captiveportal respectively
to better adhere to freedesktop standards.

Adds oneshot that can be removed after it's been in one stop release.

### DIFF
--- a/common/browserapp.cpp
+++ b/common/browserapp.cpp
@@ -10,6 +10,7 @@
 
 #include <QCoreApplication>
 #include <QString>
+#include <QStandardPaths>
 #include "browserapp.h"
 
 bool BrowserApp::captivePortal()
@@ -18,7 +19,7 @@ bool BrowserApp::captivePortal()
     static bool argsChecked = false;
 
     if (!argsChecked) {
-        if (QCoreApplication::arguments().contains(QStringLiteral("-captiveportal")))
+        if (QCoreApplication::arguments().contains(QLatin1String("-captiveportal")))
             captivePortalMode = true;
         argsChecked = true;
     }
@@ -28,20 +29,10 @@ bool BrowserApp::captivePortal()
 
 QString BrowserApp::profileName()
 {
-    static QString profileName = QStringLiteral("mozembed");
-    static bool argsChecked = false;
-
-    if (!argsChecked) {
-        const QStringList &arguments = QCoreApplication::arguments();
-
-        if (arguments.contains(QStringLiteral("-profile"))) {
-            int index = arguments.indexOf(QStringLiteral("-profile"));
-            if (index + 1 < arguments.size()) {
-                profileName = arguments.at(index + 1);
-            }
-        } else if (BrowserApp::captivePortal()) {
-            profileName = QStringLiteral("captiveportal");
-        }
+    const QStringList &arguments = QCoreApplication::arguments();
+    int index = arguments.indexOf(QLatin1String("-profile"));
+    if (index >= 0 && index + 1 < arguments.size()) {
+        return arguments.at(index + 1);
     }
-    return profileName;
+    return QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
 }
--- a/oneshot.d/browser-move-data-to-new-location
+++ b/oneshot.d/browser-move-data-to-new-location
@@ -1,0 +1,29 @@
+#!/bin/sh
+# Moves Browser and Captive portal profiles to new location
+OLD_BROWSER_PROFILE_PATH="$HOME/.mozilla/mozembed"
+NEW_BROWSER_PROFILE_PATH="$HOME/.local/share/org.sailfishos/sailfish-browser/.mozilla"
+
+OLD_CAPTIVE_PORTAL_PROFILE_PATH="$HOME/.mozilla/captiveportal"
+NEW_CAPTIVE_PORTAL_PROFILE_PATH="$HOME/.local/share/org.sailfishos/sailfish-browser-captiveportal/.mozilla"
+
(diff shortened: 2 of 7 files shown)


rainemak:
Something somewhere in gecko is still creating .mozilla/extensions directory but that's another issue.
That directory is not coming via browser or qtmozembed.

oneshot.d/browser-move-data-to-new-location
```diff
@@ -0,0 +1,29 @@
...
+NEW_BROWSER_PROFILE_PATH="$HOME/.cache/org.sailfishos/sailfish-browser/.mozilla"
+
+OLD_CAPTIVE_PORTAL_PROFILE_PATH="$HOME/.mozilla/captiveportal"
+NEW_CAPTIVE_PORTAL_PROFILE_PATH="$HOME/.cache/org.sailfishos/sailfish-browser-captiveportal/.mozilla"
+
+if [ -e "$OLD_BROWSER_PROFILE_PATH/" ] && [ ! -e "$NEW_BROWSER_PROFILE_PATH/" ]
+then
+    mkdir -p "$NEW_BROWSER_PROFILE_PATH"
+    if mv "$OLD_BROWSER_PROFILE_PATH/"* "$NEW_BROWSER_PROFILE_PATH/"
+    rm -f $OLD_BROWSER_PROFILE_PATH/.parentlock
```
Tomin1, line 13: This looks like a mistakenly placed line. Similar on the next if then statement.

oneshot.d/browser-move-data-to-new-location
```diff
@@ -0,0 +1,29 @@
...
+    then
+        rmdir "$OLD_BROWSER_PROFILE_PATH"
+        rmdir "$HOME/.mozilla"
```
Tomin1, line 16: Probably not a big deal but this would print an error if the captive portal path is still there. The directory will be removed in the next check if the captive portal path is moved.

common/browserapp.cpp
```diff
@@ -28,20 +29,12 @@ bool BrowserApp::captivePortal()
...
-            profileName = QStringLiteral("captiveportal");
+    const QStringList &arguments = QCoreApplication::arguments();
+    if (arguments.contains(QStringLiteral("-profile"))) {
+        int index = arguments.indexOf(QStringLiteral("-profile"));
+        if (index + 1 < arguments.size()) {
+            return arguments.at(index + 1);
         }
     }
-    return profileName;
+    return QStandardPaths::writableLocation(QStandardPaths::CacheLocation);
```
Tomin1, line 39: That CacheLocation feels weird here. I would think that user profile should be in DataLocation even if there is some cache inside it. If it's really only about cache this would deserve a comment.

common/browserapp.cpp
```diff
@@ -28,20 +29,12 @@ bool BrowserApp::captivePortal()
...
-        if (arguments.contains(QStringLiteral("-profile"))) {
-            int index = arguments.indexOf(QStringLiteral("-profile"));
-            if (index + 1 < arguments.size()) {
-                profileName = arguments.at(index + 1);
-            }
-        } else if (BrowserApp::captivePortal()) {
-            profileName = QStringLiteral("captiveportal");
+    const QStringList &arguments = QCoreApplication::arguments();
+    if (arguments.contains(QStringLiteral("-profile"))) {
+        int index = arguments.indexOf(QStringLiteral("-profile"));
```
Tomin1, line 34: The same QStringLiteral defined twice.

rainemak: That's bad! Even though was there already, I'll clean it up.

common/browserapp.cpp
rainemak: There are all kind of things but yeah AppDataLocation sounds better.

oneshot.d/browser-move-data-to-new-location
rainemak: Oops. Fixed.

oneshot.d/browser-move-data-to-new-location
```diff
@@ -0,0 +1,29 @@
...
+fi
+
+if [ -e "$OLD_CAPTIVE_PORTAL_PROFILE_PATH/" ] && [ ! -e "$NEW_CAPTIVE_PORTAL_PROFILE_PATH/" ]
+then
+    rm -f $OLD_CAPTIVE_PORTAL_PROFILE_PATH/.parentlock
+    mkdir -p "$NEW_CAPTIVE_PORTAL_PROFILE_PATH"
+    if mv "$OLD_CAPTIVE_PORTAL_PROFILE_PATH/"* "$NEW_CAPTIVE_PORTAL_PROFILE_PATH/"
+    then
+        rmdir "$OLD_CAPTIVE_PORTAL_PROFILE_PATH"
+        rmdir "$HOME/.mozilla" > /dev/null 2>&1
```
Tomin1, line 27: One more thing about this: This oneshot fails because this remove fgals if ~/.mozilla/extensions is still there and this is the last command executed so the exit status of the script is whatever this command exits with. On second run (== usually next reboot) this oneshot doesn't do anything and succeeds. That might be something to consider still. Otherwise it seems to work fine.

rainemak: We could do this at the end with rm -rf :)

Yeah, that extensions dir is not nice really. Maybe we just leave the .mozilla there for now and handle final cleanup in JB#52230

Tomin1: > Maybe we just leave the .mozilla there for now and handle final cleanup in JB#52230

Sounds good to me.

Tomin1 (approved):

PR #753, opened 2020-12-01
[sailfish-browser] Add connection details to CertificateInfo. Contributes to JB#47514

atatarov:


llewelld:
This is good. The only question in my mind is how the alignment of the two CertificateLabels should work. Currently they're both centred individually, but arguably they should line up along the key-value split and then be centred as a group. Is the current "individually centred" approach the agreed design? I checked the designs on JB#50930, but they only include a single line.

I also notice that when the toolbar is tallest (all data + permission icons) there's a glitch in the animation when the toolbar opens. It opens fully, then jumps down slightly. Is this something that can be fixed?

llewelld:
Image of current version for reference.
Screenshot_20201201_001

llewelld:
Thanks for the changes @atatarov. I definitely prefer the previous alignment, but your latest changes match the design, which I think is the correct thing to do.

Screenshot_20201202_001

llewelld:
I'm still getting a judder in the animation when the pane opens. Is it possible to fix this?

atatarov:
> I'm still getting a judder in the animation when the pane opens. Is it possible to fix this?

I tested it on inoi r7 and couldn't figure out what problems with animation. 
Maybe the model didn't load before the animation started?

llewelld:
> Maybe the model didn't load before the animation started?

Yes, it's possible that's the cause. Assuming there's only one row of icons, can you put a placeholder height in while the model is loading to avoid this?

atatarov:
> > Maybe the model didn't load before the animation started?
> 
> Yes, it's possible that's the cause. Assuming there's only one row of icons, can you put a placeholder height in while the model is loading to avoid this?

trying keep min 1 row always. How does it work on your side?

atatarov:
Add pr
https://github.com/sailfishos/sailfish-components-webview/pull/85

llewelld (approved):
This now works nicely for me; thanks for persevering with the fix. The Loader seems to do the trick nicely and I don't see any animation judder now.

PR #755, opened 2020-12-01
[browser] Adapt bookmark edit dialog to use the new validation style. Contributes to JB#50329

jpetrell:


src/pages/components/BookmarkEditDialog.qml
```diff
@@ -53,26 +67,21 @@ UserPrompt {
...
-            EnterKey.onClicked: {
-                if (canAccept) {
-                    accept()
-                } else {
-                    titleField.focus = true
-                }
-            }
+            EnterKey.enabled: canAccept
+            EnterKey.iconSource: "image://theme/icon-m-enter-accept"
+            EnterKey.onClicked: accept()
```
pvuorela, line 84: An enter can still be coming from the hardware keyboard. Accept() doesn't seem to protect with canAccept -> all cases like this should be protected?

pvuorela: Or wait, was there an acceptBlocked happening.

src/pages/components/BookmarkEditDialog.qml
```diff
@@ -53,26 +67,21 @@ UserPrompt {
...
-            placeholderText: qsTrId("sailfish_browser-ph-url_edit")
+            acceptableInput: text.length > 0
+            onActiveFocusChanged: if (!activeFocus) errorHighlight = !acceptableInput
+            onAcceptableInputChanged: if (acceptableInput) errorHighlight = false
 
             //: Label for textfield to edit bookmark's URL
             //% "URL"
             label: qsTrId("sailfish_browser-la-url_editor")
+            //% "URL is required"
+            description:  errorHighlight ? qsTrId("sailfish_browser-la-url_editor_error") : ""
```
pvuorela, line 79: Extra space there.

src/pages/components/BookmarkEditDialog.qml
```diff
@@ -36,15 +46,19 @@ UserPrompt {
...
+
+            acceptableInput: text.length > 0
+            onActiveFocusChanged: if (!activeFocus) errorHighlight = !acceptableInput
+            onAcceptableInputChanged: if (acceptableInput) errorHighlight = false
 
             //: Label for bookmark/favorite's title edit field
             //% "Title"
             label: qsTrId("sailfish_browser-la-title_editor")
+            //% "Title is required"
+            description:  errorHighlight ? qsTrId("sailfish_browser-la-title_editor_error") : ""
```
pvuorela, line 61: Extra space here as well.

jpetrell:
Fixed the empty spaces.

src/pages/components/BookmarkEditDialog.qml
jpetrell: Yeah calling accept() will do same thing as attempting to accept with the dialog header buttons, show error. We would actually need dimmed mode for accept enter key that allowed clicking so could show error highlight there as well.

pvuorela:
LGTM

PR #757, opened 2020-12-01
[browser] Add sandboxing. JB#51740

rainemak:


oneshot.d/browser-cleanup-startup-cache
```diff
@@ -3,5 +3,5 @@
 rm -rf ~/.mozilla/mozembed/startupCache
 rm -rf ~/.mozilla/captiveportal/startupCache
 # New startup cache location
-rm -rf ~/.local/share/sailfish-browser/.mozilla/startupCache
```
Tomin1, line 6: Are these paths actually correct? I would think that there needs to be `org.sailfishos` directory before `browser`.

oneshot.d/browser-cleanup-startup-cache
```diff
@@ -3,5 +3,5 @@
...
-rm -rf ~/.local/share/sailfish-browser-captiveportal/.mozilla/startupCache
+rm -rf ~/.local/share/browser/.mozilla/startupCache
```
Tomin1, line 6: Are these paths actually correct? I would think that there needs to be org.sailfishos directory before browser.

sailfish-browser-captiveportal.desktop
```diff
@@ -5,5 +5,9 @@ Name=Captive Portal Browser
...
 X-MeeGo-Translation-Catalog=sailfish-browser-captiveportal
 Icon=icon-launcher-browser
-Exec=invoker -s --type=browser -G /usr/bin/sailfish-browser-captiveportal -captiveportal %U
-Comment=Sailfish UI application
+Exec=/usr/bin/sailjail -p sailfish-browser-captiveportal.desktop /usr/bin/sailfish-browser-captiveportal -captiveportal %U
+Comment=Sailfish captive portal
+[X-Sailjail]
+Permissions=CaptivePortal;Privileged;Audio;Location;Internet;Downloads
+OrganizationName=org.sailfishos
+ApplicationName=browser.captiveportal
```
Tomin1, line 13: Originally I envisioned that ApplicationName can't have a dot. I hoped dashes and underscores would suffice. Well, if it works...

rainemak: Me too.

settings/browser.qml
```diff
@@ -12,24 +11,18 @@
 
 import QtQuick 2.0
 import Sailfish.Silica 1.0
+import org.sailfishos.browser.settings 1.0
 import org.nemomobile.dbus 2.0
+import com.jolla.settings 1.0
 
-Page {
+ApplicationSettings {
     id: page
```
Tomin1, line 19: No need to keep this id.

oneshot.d/browser-cleanup-startup-cache
rainemak: Well spotted!

sailfish-browser-captiveportal.desktop
rainemak: why didn't I just have "captiveportal" there's no need for this "browser.". You're right.

Tomin1 (approved):
LGTM

PR #758, opened 2020-12-03
Split browser between captiveportal and browser

rainemak:


rainemak:
Small update... when moving last bits missed this:
```
-    <file alias="WebView.qml">../../../apps/qml/WebView.qml</file>
+    <file alias="WebView.qml">../../../apps/shared/WebView.qml</file>
```
Points now to shared.

Directory structure could be aligned probably even more so that this "shared" would work better. But I think that this is already much better.

apps/captiveportal/captiveportalservice.cpp
```diff
@@ -0,0 +1,52 @@
+#include "captiveportalservice.h"
```
Tomin1, line 1: Missing copyright header.

apps/captiveportal/captiveportal.pro
```diff
@@ -0,0 +1,54 @@
...
+    message("Building with qdeclarative-boostable support")
+    DEFINES += HAS_BOOSTER
+    PKGCONFIG += qdeclarative5-boostable
+} else {
+    warning("qdeclarative5-boostable not available; startup times will be slower")
+}
+
+# Translations
+TS_PATH = $$PWD
+TS_FILE = $$OUT_PWD/sailfish-browser.ts
```
Tomin1, line 24: A question: is it a problem that both captiveportal and browser define the same translation files? Is there a conflict?

rainemak: I had in my notes that this should be split as well but I'd not like to disturb translations now with this. Need to think about.

I think this can be solved by moving the translation.pri stuff to the apps directory so that we run it through only once.

rainemak:
Pushed to review fixes on top that will be squashed before merging.

Tomin1 (approved):

PR #761, opened 2020-12-08
[sailfish-browser] Fix backup. Contributes to JB#52318

Tomin1:
This only fixes backing up. Older backups can not be restored correctly
with this change.


rainemak (approved):
This is one step forward. Hence, approving.

PR #762, opened 2020-12-08
[sailfish-browser] Fix restoring old backups. Fixes JB#52318

Tomin1:
This moves the files around in the in memory copy of the archive before
restoring it. This doesn't modify the archive itself because it is
mounted with nosave option.

### DIFF
--- a/backup-unit/browserunit.cpp
+++ b/backup-unit/browserunit.cpp
@@ -6,6 +6,8 @@
 #include <QVariantMap>
 #include <QDebug>
 #include <QLoggingCategory>
+#include <QFileInfo>
+#include <QDir>
 #include <sys/types.h>
 #include <signal.h>
 #include <set>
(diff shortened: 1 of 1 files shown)


backup-unit/browserunit.cpp
```diff
@@ -70,21 +72,67 @@ const QString browser_dir = ".local/share/org.sailfishos/browser";
...
+}
+
+void fix(QString common, QString file, QString oldPath, QString newPath)
+{
+    const auto pathTemplate = QStringLiteral("%1/%3/%2").arg(common).arg(file);
+    QFileInfo source(pathTemplate.arg(oldPath));
+    if (source.exists()) {
+        QFileInfo dest(pathTemplate.arg(newPath));
+        dest.dir().mkpath(".");
+        if (!source.dir().rename(source.fileName(), dest.absoluteFilePath())) {
```
Tomin1, line 110: I wonder if I should instead copy the files with vault::unit::copy so that it can keep the timestamps and such. I don't know if it works properly inside the archive though.

Tomin1: Ah, it seems it might just keep those already.

rainemak (approved):
Didn't test on device but change looks good.

PR #765, opened 2020-12-10
[sailfish-browser] Add dialog for create new permission exception. Contributes to JB#52422

atatarov:


apps/browser/qml/pages/components/PermissionCreateDialog.qml
```diff
@@ -0,0 +1,67 @@
...
+
+    Column {
+        width: parent.width
+        anchors.top: header.bottom
+
+        TextField {
+            id: textField
+
+            width: parent.width
+            //% "Website addres"
```
rainemak, line 37:
```suggestion
            //% "Website address"
```

apps/browser/qml/pages/components/PermissionCreateDialog.qml
```diff
@@ -0,0 +1,67 @@
...
+                if (errorHighlight) {
+                    //% "Should begin with http:// or https://"
+                    return qsTrId("sailfish_browser-he-error-website")
+                } else if (text == "") {
+                    return ""
+                } else {
+                    return placeholderText
+                }
+            }
+            onTextChanged: errorHighlight = (!_regExp.test(text) && text !== "")
```
rainemak, line 49: Why this is not a binding like:

errorHighlight: !_regExp.test(text) && text !== ""

apps/browser/qml/pages/components/PermissionCreateDialog.qml
```diff
@@ -0,0 +1,67 @@
...
+        //: Accept text when add new permission
+        //% "Add"
+        acceptText: qsTrId("sailfish_browser-he-add")
+    }
+
+    Column {
+        width: parent.width
+        anchors.top: header.bottom
+
+        TextField {
```
llewelld, line 33: It'd be good to add `inputMethodHints: Qt.ImhUrlCharactersOnly` to the TextField.

apps/browser/qml/pages/components/PermissionCreateDialog.qml
```diff
@@ -0,0 +1,67 @@
...
+            id: textField
+
+            width: parent.width
+            //% "Website addres"
+            placeholderText: qsTrId("sailfish_browser-he-website-address")
+            label: {
+                if (errorHighlight) {
+                    //% "Should begin with http:// or https://"
```
llewelld, line 41: I'd suggest changing this to "Must begin with" (since "should" implies it's optional).

llewelld, line 41: Does the transfer prefix actually affect behaviour (i.e. do permissions for http and https versions of the same domain act independently)? If not, perhaps the requirement could be dropped?

If they do, perhaps the field could be pre-populated with "https://" so the user doesn't have to type it in each time? Or alternatively add the domain of the currently open website?

apps/browser/qml/pages/components/PermissionCreateDialog.qml
```diff
@@ -0,0 +1,67 @@
...
+                } else if (text == "") {
+                    return ""
+                } else {
+                    return placeholderText
+                }
+            }
+            onTextChanged: errorHighlight = (!_regExp.test(text) && text !== "")
+        }
+
+        ComboBox {
```
llewelld, line 52: This is a design issue, but it would be nice to include the icon for the permission type here, to reflect the design of the Permissions page:

permission-icons

apps/browser/qml/pages/components/PermissionCreateDialog.qml
atatarov: PermissionManager http and https accept as different hosts

apps/browser/qml/pages/components/PermissionListItem.qml
```diff
@@ -0,0 +1,32 @@
...
+    width: parent.width
+    contentHeight: Theme.itemSizeMedium
+    onClicked: openMenu()
+
+    Row {
+        width: parent.width - 2*Theme.horizontalPageMargin
+        x: Theme.horizontalPageMargin
+        anchors.verticalCenter: parent.verticalCenter
+
+        Icon { id: icon }
```
llewelld, line 20: The Icon and ValueButton look misaligned vertically to me. The should both be vertically centred in the parent row.

apps/browser/qml/pages/components/PermissionCreateDialog.qml
```diff
@@ -0,0 +1,82 @@
...
+        //: Accept button text for adding a new permission
+        //% "Add"
+        acceptText: qsTrId("sailfish_browser-he-add")
+    }
+
+    Column {
+        width: parent.width
+        anchors.top: header.bottom
+
+        TextField {
```
llewelld, line 34: Did you decide against pre-populating the text entry field? I think you could still add "https://" to it, in order to make it easier for the user.

llewelld (changes requested):
Just two very small things that I think still need looking at here:
1. The icon/text vertical alignment in the PermissionListItem.
2. Consider pre-populating with https:// as we discussed.

llewelld (approved):
Thanks; this LGTM now.

PR #766, opened 2020-12-10
Allow some D-Bus APIs for unprivileged clients

Tomin1:
Privileged check doesn't work inside sandbox. There is little reason to
keep requestTab privileged API. This allow it for unprivileged clients.

It would be bad to allow closeTab for all. Mark the process that called
requestTab as owner of the tab and allow the owner to call it.

### DIFF
--- a/apps/browser/browserservice.cpp
+++ b/apps/browser/browserservice.cpp
@@ -21,20 +21,21 @@
 #include <QFileInfo>
 #include <QCoreApplication>
 
-#define SAILFISH_BROWSER_UI_SERVICE QLatin1String("org.sailfishos.browser.ui")
+namespace {
+const auto ProcDir = QStringLiteral("/proc/%1");
+const auto ErrorPidIsNotPrivileged = QStringLiteral("PID %1 is not in privileged group");
+const auto SailfishBrowserUiService = QStringLiteral("org.sailfishos.browser.ui");
+}
 
-#define IS_PRIVILEGED \
+#define GET_PID() connection().interface()->servicePid(message().service())
+
+#define IS_PRIVILEGED() \
     if (!calledFromDBus()) { \
         return true; \
     } \
-    uint pid = connection().interface()->servicePid(message().service()); \
-    QFileInfo info(QString("/proc/%1").arg(pid)); \
-    if (info.group() != "privileged" && info.owner() != "root") { \
-        sendErrorReply(QDBusError::AccessDenied, \
-            QString("PID %1 is not in privileged group").arg(pid)); \
-        return false; \
-    } \
-    return true;
+    uint pid = GET_PID(); \
+    QFileInfo info(ProcDir.arg(pid)); \
+    return info.group() == "privileged" || info.owner() == "root";
 
 BrowserService::BrowserService(QObject * parent)
     : QObject(parent)
@@ -106,7 +107,15 @@ void BrowserService::dumpMemoryInfo(const QString &fileName)
 
 bool BrowserService::isPrivileged() const
 {
-    IS_PRIVILEGED;
+    auto isPrivileged = [=] {
+        IS_PRIVILEGED();
+    };
+    if (!isPrivileged()) {
+        sendErrorReply(QDBusError::AccessDenied,
+                ErrorPidIsNotPrivileged.arg(GET_PID()));
+        return false;
+    }
+    return true;
 }
 
 BrowserUIServicePrivate::BrowserUIServicePrivate()
@@ -124,7 +133,7 @@ BrowserUIService::BrowserUIService(QObject *parent)
     new UIServiceDBusAdaptor(this);
     QDBusConnection connection = QDBusConnection::sessionBus();
     if(!connection.registerObject("/ui", this) ||
-            !connection.registerService(SAILFISH_BROWSER_UI_SERVICE)) {
+            !connection.registerService(SailfishBrowserUiService)) {
         d->registered = false;
     }
 }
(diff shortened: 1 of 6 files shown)


Tomin1:
This is kind of Request For Comments PR, I'm not totally sure if this makes sense from all angles. Also this still leaves a few non-working privileged APIs.

This is also missing removing that ownership if user starts to use the tab for something else. I left TODO comment for that but it can be fixed of course, just need to understand how to detect that.

apps/core/declarativewebcontainer.cpp
```diff
@@ -496,17 +496,30 @@ void DeclarativeWebContainer::closeTab(int tabId)
...
+{
+    return requestTabWithOwner(tabId, url, 0);
+}
+
+int DeclarativeWebContainer::requestTabWithOwner(int tabId, const QString &url, uint ownerPid)
 {
     bool activated = m_model->activateTabById(tabId);
     if (!activated) {
         tabId = m_model->newTab(url);
+        // TODO: remove ownership of the tab if user changes URL
```
rainemak, line 508: Maybe navigation on the requested tab is fine and changing url should be nice as we really cannot assume much about usage of the requested tab. That said, if there's onOpenUrlRequested or user herself enters a url (function loadPage(url) of the Overlay) then I think it'd make sense to remove ownership.

How does that sound to you?

Tomin1: Pushed a commit that might work for this purpose.

apps/browser/browserservice.cpp
```diff
@@ -23,15 +23,19 @@
 
 #define SAILFISH_BROWSER_UI_SERVICE QLatin1String("org.sailfishos.browser.ui")
 
-#define IS_PRIVILEGED \
+#define GET_PID connection().interface()->servicePid(message().service())
```
spiiroin, line 26: nit: macro that looks like constant, but does complex call sequence -> GET_PID() would feel more appropriate

apps/browser/browserservice.cpp
```diff
@@ -23,15 +23,19 @@
...
+
+#define IS_PRIVILEGED(reply) \
```
spiiroin, line 28: and another nit: a macro that looks like a predicate but is a fancy return with double semicolon / other syntactical issues depending on where and how it is used.

well, it was like this originally but imo the end result is that places where it is used are confusing and can't be understood without looking at the macro def

Tomin1:
I think I need to improve those macros and also it seems that my last commit to clear ownership doesn't work at all -> more work needed.

apps/browser/qml/pages/BrowserPage.qml
```diff
@@ -343,7 +343,9 @@ Page {
             }
 
             webView.grabActivePage()
-            if (!webView.tabModel.activateTab(url)) {
+            if (webView.tabModel.activateTab(url)) {
+                webView.unownActiveTab()
```
rainemak, line 347: Whilst unown can be understood, maybe this should be called like releaseActiveTabOwnership or takeActiveTabOwnership.  Latter wouldn't actually "take" anything but would freeing already set ownership.

Summary: to me releasing / freeing feels quite natural.

Tomin1:
> I think I need to improve those macros and also it seems that my last commit to clear ownership doesn't work at all -> more work needed.

Duh. It works. It just didn't apply the logging rules.

apps/browser/qml/pages/BrowserPage.qml
Tomin1: Renamed the method.

apps/browser/browserservice.cpp
Tomin1: Improved these. Now it doesn't take an argument but it's simply a check.

rainemak (approved):
This should be ok now. I'm fine if ok for @spiiroin as well. Approving.

apps/browser/browserservice.cpp
```diff
@@ -106,7 +107,15 @@ void BrowserService::dumpMemoryInfo(const QString &fileName)
 
 bool BrowserService::isPrivileged() const
 {
-    IS_PRIVILEGED;
+    auto isPrivileged = [=] {
+        IS_PRIVILEGED();
```
spiiroin, line 111: That "double semicolon" thing == what happens here. 'IS_PRIVILEGED();' expands to '... == "root";;' - which is harmless, but can also trigger empty statement warnings / stte. The canonical thing to do would be to enclose macro def in "do { multiple; statements; } while(0)" -> semicolon comes from where the macro is used, and as a bonus also things like "if (foo) MACRO(x);" behaves like one might expect when looking at the code.

On a closer look. What is this? Lambda method to work around a macro that looks like a predicate but actually does a return? I guess it works and static function / similar is out of the question due to calledFromDBus() method requiring some suitable object context?

Since this was just something that caught by eye, not what the PR was about: It is fine as it is, but if you wish to make places where the macro is used look more like regular code, one alternative might be to use statement expression (gnu extension) so that it evaluates into boolean value instead of doing a value return -> could avoid that lambda here / have return IS_PRIVILEGED() in that other place. Then again I'm not 100% sure if this c extensions also c++ ok and whether it is better / worse than what there already is depends on the reader...

```
#define IS_PRIVILEGED() ({\
    bool isPrivileged = true;\
    if (calledFromDBus()) {\
        uint pid = GET_PID(); \
        QFileInfo info(QString("/proc/%1").arg(pid)); \
        isPrivileged = info.group() == "privileged" || info.owner() == "root";\
    }\
    isPrivileged;\
})

```

spiiroin:
Ignoring the macro nits, what I understand of the other changes: LGTM

PR #768, opened 2020-12-14
[sailfish-browser] Add desktop site mode. Contributes to JB#50924

paveltkv:


apps/history/declarativetabmodel.cpp
```diff
@@ -454,6 +454,20 @@ void DeclarativeTabModel::onUrlChanged()
...
 }
 
+void DeclarativeTabModel::onDesktopModeChanged()
+{
+    DeclarativeWebPage *webPage = qobject_cast<DeclarativeWebPage *>(sender());
+    if (webPage) {
+        int tabIndex = findTabIndex(webPage->tabId());
+        if (tabIndex >= 0 && m_tabs.at(tabIndex).desktopMode() != webPage->desktopMode()) {
+            QVector<int> roles;
+            roles << UrlRole;
```
rainemak, line 464: This is not yet changing url as there will be another load when the mode changes. If url did change due to desktop mode change, then that will be handled onUrlChanged slot.

As you are introducing already desktop mode to the tab, let's add DesktopModeRole role (exposed as "desktopMode"). You need to fix/update data method to return the desktop mode as well.

rainemak (changes requested):
Some comments added. Please check and update accordingly. Will still check on device how does this TextSwitch look like.

apps/browser/qml/pages/components/SecondaryBar.qml
```diff
@@ -74,6 +74,21 @@ Column {
         }
     }
 
+    TextSwitch {
+        height: overlay.toolBar.rowHeight
+        enabled: webView.contentItem
+        automaticCheck: false
+
+        //: Label for text switch that reloads page in desktop mode
+        //% "Desktop site"
```
rainemak, line 83: This introduces a text alignment issue but good starting point.

We should have icons (mobile & desktop), text changing accordingly (Show mobile site, Show desktop site), and use the OverlayListItem to get the alignment fixed. Filed JB#53202. Please fix that one after this has been landed.

rainemak (approved):
Functionality is good now. Please check JB#53202 after this has been landed so that we get text alignment fixed as well.

PR #770, opened 2020-12-17
[browser] Add Android ua overrides to fix video playback on various domains. JB#51967, fixes JB#51909

abranson:


rainemak (approved):
Use "Fixes JB#51967" as well. Usually fixes is starting with a capital letter (Fixes).

Approving already.

PR #776, opened 2021-01-14
[browser] Add the ability to add bookmark to app grid. Contributes to JB#52719

atatarov:


apps/browser/qml/pages/components/BookmarkItem.qml
```diff
@@ -101,6 +94,16 @@ ListItem {
                 text: qsTrId("sailfish_browser-me-copy-to-clipboard")
                 onClicked: Clipboard.text = model.url
             }
+            MenuItem {
+                //% "Add to App Grid"
```
rainemak, line 98: No need to re-define. You can leave a normal comment that referring to the FavoriteContextMenu.

apps/browser/qml/pages/components/BookmarkItem.qml
```diff
@@ -120,4 +123,34 @@ ListItem {
...
+    Component {
+        id: editDialog
+        BookmarkEditDialog {
+            onAccepted: bookmarkModel.edit(index, editedUrl, editedTitle)
+        }
+    }
+
+    Component {
+        id: addToLauncherDialog
+        BookmarkEditDialog {
```
rainemak, line 143: This could even be wrapped into a component AddToAppGridDialog and then FavoriteGrid updated accordingly. Now code is copy-paste.

rainemak (commented):
Two minor notes. Please fix those and let's get this one merged.

apps/browser/qml/pages/components/AddToAppGridDialog.qml
```diff
@@ -0,0 +1,16 @@
+import QtQuick 2.0
+import Sailfish.Silica 1.0
+
+BookmarkEditDialog {
+    property string icon
+    property var desktopBookmarkWriter
```
rainemak, line 6:
```suggestion
    property Component desktopBookmarkWriter
```

As this is now DesktopBookmarkWriter component I'd also call this such way.

apps/browser/qml/pages/components/AddToAppGridDialog.qml
```diff
@@ -0,0 +1,16 @@
...
+    property var view
```
rainemak, line 7:
```suggestion
    property QtObject bookmarkWriterParent
```

This is in reality an owner / parent of the to be created desktopBookmarkWriter not a view per se. Hence, above suggestion.

apps/browser/qml/pages/components/AddToAppGridDialog.qml
```diff
@@ -0,0 +1,16 @@
...
+
+    //% "Add to App Grid"
+    title: qsTrId("sailfish_browser-he-add_bookmark_to_launcher")
+    canAccept: editedUrl !== "" && editedTitle !== ""
+    onAccepted: {
+        var bookmarkWriter = desktopBookmarkWriter.createObject(view)
```
rainemak, line 13: Big fat warning if there is no component passed whilst both cases should be working fine... i.e.
```
if (desktopBookmarkWriter) {
    var bookmarkWriter = desktopBookmarkWriter.createObject(view)
    bookmarkWriter.save(editedUrl, editedTitle, icon)
} else {
    console.log("Cannot save to launcher without bookmark writter!!")
}
```

apps/browser/qml/pages/components/BookmarkItem.qml
```diff
@@ -101,6 +94,17 @@ ListItem {
...
             }
+            MenuItem {
+                text: qsTrId("sailfish_browser-me-add_to_launcher")
+                onClicked: pageStack.animatorPush("AddToAppGridDialog.qml",
+                                                  {
+                                                      "url": url,
+                                                      "title": title,
+                                                      "icon": favicon,
+                                                      "desktopBookmarkWriter": desktopBookmarkWriter,
+                                                      "view": listView
```
rainemak, line 105:
```suggestion
                                                      "bookmarkWriterParent": pageStack
```

There's potential object life-cycle problem now. The case where you have ""view": favoriteGrid" is fine from life-cycle point of view as the favoriteGrid is always there. However, in this case the saving will be aborted if you pop the Bookmark page before saving has finished.

Hmm, have to correct a bit... DesktopBookmarkWriter will block if there pending save upon destruction. Even so it be better not to block. Hence, ```bookmarkWriterParent: pageStack```.

apps/browser/qml/pages/components/FavoriteGrid.qml
```diff
@@ -55,11 +55,13 @@ IconGridViewBase {
...
             // url, title, favicon
-            pageStack.animatorPush(addToLauncherDialog,
+            pageStack.animatorPush("AddToAppGridDialog.qml",
                                    {
                                        "url": url,
                                        "title": title,
-                                       "icon": favicon
+                                       "icon": favicon,
+                                       "desktopBookmarkWriter": desktopBookmarkWriter,
+                                       "view": favoriteGrid
```
rainemak, line 64:
```suggestion
                                       "bookmarkWriterParent": favoriteGrid
```

rainemak (changes requested):
One iteration needed still. Added few notes and suggestions. Please check them all.

apps/browser/qml/pages/components/FavoriteGrid.qml
```diff
@@ -55,11 +55,13 @@ IconGridViewBase {
...
             // url, title, favicon
-            pageStack.animatorPush(addToLauncherDialog,
+            pageStack.animatorPush("AddToAppGridDialog.qml",
                                    {
                                        "url": url,
                                        "title": title,
-                                       "icon": favicon
+                                       "icon": favicon,
+                                       "desktopBookmarkWriter": desktopBookmarkWriter,
+                                       "bookmarkWriterParent": pageStack
```
rainemak, line 64: here the favorite grid was just fine... as lifetime to of this was the same as the pagestack's. But yes, pageStack works here as well.

rainemak (approved):

PR #779, opened 2021-01-28
[browser] Set BROWSER environment variable correctly. Fixes JB#52900

Tomin1:


rainemak (approved):
There's a difference in the cover behaviour but that can be fixed later. LGTM.

PR #781, opened 2021-02-02
[sailfish-browser] Add a button to select the default download folder. Contributes to JB#52931

atatarov:


apps/browser/qml/pages/components/BrowserListItem.qml
```diff
@@ -0,0 +1,57 @@
...
+            id: icon
+
+            anchors.verticalCenter: parent.verticalCenter
+        }
+
+        Column {
+            width: parent.width
+            anchors.verticalCenter: parent.verticalCenter
+
+            ValueButton {
```
llewelld, line 33: `ValueButton` will apply an indent of `Theme.horizontalPageMargin` by default, so this should be set to `Theme.paddingMedium` in this case to align with `secondaryLabel`:
```
leftMargin: Theme.paddingMedium
```

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -150,6 +151,47 @@ Page {
                 }
                 onClicked: pageStack.push("PermissionPage.qml")
             }
+
+            BrowserListItem {
+                //% "Save destination"
```
llewelld, line 156: It may be that this is part of a fixed design already, but for me "Download destination" would be clearer than "Save destination".

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -150,6 +151,47 @@ Page {
...
+                label: qsTrId("settings_browser-la-save_destination")
+                iconSource: "image://theme/icon-m-download"
+                value: {
+                    if (WebEngineSettings.useDownloadDir) {
+                        //% "Download dir"
+                        return qsTrId("sailfish_browser-me-download_dir")
```
llewelld, line 162: Again, this may be part of the fixed design, but why not just give the folder name here, and avoid the use of `secondaryText`?:

"Download to: /home/defaultuser/Downloads"
"Download to: Always ask"

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -150,6 +151,47 @@ Page {
...
+                    } else {
+                        //% "Ask"
+                        return qsTrId("sailfish_browser-me-ask")
+                    }
+                }
+
+                secondaryText: {
+                    if (WebEngineSettings.useDownloadDir) {
+                        //% "Downloaded files will be saved to %1"
```
llewelld, line 171: Could be "Files will be downloaded to %1"

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -150,6 +151,47 @@ Page {
...
+                        return qsTrId("sailfish_browser-me-will_be_saved_to_download").arg(WebEngineSettings.downloadDir)
+                    } else {
+                        //% "Always ask where save to"
```
llewelld, line 174: Should be "Always ask where to save" or "Always ask where to save to" (I prefer the former personally).

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -150,6 +151,47 @@ Page {
...
+                        return qsTrId("sailfish_browser-me-ask_where_save_to")
+                    }
+                }
+
+                menu: ContextMenu {
+                    MenuItem {
+                        // "Downloads"
```
llewelld, line 181: Missing `%`:
```
//% "Downloads"
```

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -150,6 +151,47 @@ Page {
...
+                        text: qsTrId("sailfish_browser-me-downloads")
+                        onClicked: {
+                            WebEngineSettings.useDownloadDir = true
+                            pageStack.animatorPush(folderPickerPage)
+                        }
+                    }
+                    MenuItem {
+                        // "Custom"
```
llewelld, line 189: Missing `%` here too.

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -150,6 +151,47 @@ Page {
...
+                    if (WebEngineSettings.useDownloadDir) {
+                        //% "Downloaded files will be saved to %1"
+                        return qsTrId("sailfish_browser-me-will_be_saved_to_download").arg(WebEngineSettings.downloadDir)
+                    } else {
+                        //% "Always ask where save to"
+                        return qsTrId("sailfish_browser-me-ask_where_save_to")
+                    }
+                }
+
+                menu: ContextMenu {
```
llewelld, line 179: The phrasing is a bit confusing for me here too. Could it be something like:
"Select a download folder"
"Always ask"

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -188,4 +230,16 @@ Page {
             searchInstalledNotice.show()
         }
     }
+
+    Component {
+        id: folderPickerPage
+
+        FolderPickerPage {
```
llewelld, line 237: Could this be set to start in the Downloads folder by default, or the folder last chosen by the user?

apps/browser/qml/pages/components/PermissionListItem.qml
```diff
@@ -1,37 +0,0 @@
-import QtQuick 2.0
```
llewelld, line 1: Nice change that you're able to combine `PermissionListItem` with `BrowserListItem` here :1st_place_medal:

apps/browser/qml/pages/components/BrowserListItem.qml
```diff
@@ -0,0 +1,57 @@
...
+                id: secondaryLabel
+
+                anchors {
+                    left: valueButton.left
+                    leftMargin: Theme.paddingMedium
+                    right: valueButton.right
+                }
+                color: root.highlighted ? palette.secondaryHighlightColor : palette.secondaryColor
+                font.pixelSize: Theme.fontSizeSmall
+                wrapMode: Text.WordWrap
```
llewelld, line 52: `Text.Wrap` would be safer than `Text.WordWrap`.

llewelld (changes requested):
Looking nice. Just some small suggestions/changes.

apps/browser/qml/pages/components/BrowserListItem.qml
```diff
@@ -0,0 +1,57 @@
+import QtQuick 2.0
```
llewelld, line 1: This needs a copyright header, and the other files need them updating for 2021 too.

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -150,6 +151,47 @@ Page {
...
+                    } else {
+                        //% "Ask"
+                        return qsTrId("sailfish_browser-me-ask")
+                    }
+                }
+
+                secondaryText: {
+                    if (WebEngineSettings.useDownloadDir) {
+                        //% "Downloaded files will be saved to %1"
+                        return qsTrId("sailfish_browser-me-will_be_saved_to_download").arg(WebEngineSettings.downloadDir)
```
llewelld, line 172: Looking at the design, this should only show the folder below the user's home folder (i.e. "Downloads" rather than "/home/defaultuser/Downloads".

Presumably the file picker should also be restricted to the user's home as well; there shouldn't be any reason to try to save a file below that.

apps/browser/qml/pages/components/BrowserListItem.qml
```diff
@@ -0,0 +1,57 @@
...
+                id: valueButton
+
+                width: parent.width - icon.width
+                labelColor: root.highlighted ? palette.highlightColor : palette.primaryColor
+                valueColor: Theme.highlightColor
+                opacity: 1.0
+                enabled: false
+            }
+            Label {
+                id: secondaryLabel
```
llewelld, line 43: Rather than creating a separate `Label` here, there is the `description` text of the `ValueButton`. It looks like this can be used instead.

apps/browser/qml/pages/SettingsPage.qml
llewelld: Now that I've seen the design, please ignore this comment; this is the correct text following the design.

apps/browser/qml/pages/SettingsPage.qml
llewelld: Again, please ignore more earlier comment.

It looks like the string used here should be either "Downloads" or the leafname of the folder. Personally I think the leafname would be clearer.

apps/browser/qml/pages/SettingsPage.qml
llewelld: Please ignore my earlier comment. Based on the design the string should be "Downloaded files will be saved to %1 folder".

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -150,6 +151,47 @@ Page {
...
+                //% "Save destination"
+                label: qsTrId("settings_browser-la-save_destination")
+                iconSource: "image://theme/icon-m-download"
+                value: {
+                    if (WebEngineSettings.useDownloadDir) {
+                        //% "Download dir"
+                        return qsTrId("sailfish_browser-me-download_dir")
+                    } else {
+                        //% "Ask"
+                        return qsTrId("sailfish_browser-me-ask")
```
llewelld, line 165: Following the design, this text should be "Always ask".

llewelld:
I've now seen the design for this, and amended some of my earlier comments as a result. Apologies for the confusion, but I hope the comments still make sense.

apps/browser/qml/pages/SettingsPage.qml
atatarov: I think it may work, but it requires changes to sailfish-pickers. Now if set default directory /home/user/MyFolder/, I will not be able to go to /home/user/

apps/browser/qml/pages/components/BrowserListItem.qml
llewelld: I'm unresolving this as I still see the additional space here I'm afraid. According to the design, the "Save destination" text should align with the text "Permissions" directly above, but currently the "Save destination" text is too far to the right.

apps/browser/qml/pages/components/BrowserListItem.qml
```diff
@@ -0,0 +1,54 @@
...
+        id: row
+
+        width: parent.width - 2*Theme.horizontalPageMargin
+        x: Theme.horizontalPageMargin
+        anchors.verticalCenter: parent.verticalCenter
+
+        Icon {
+            id: icon
+
+            anchors.verticalCenter: parent.verticalCenter
```
llewelld, line 36: The "value" text of `valueButton` is animated nicely if it moves onto a new line, whereas the icon's vertical position will jump without any animation, which looks jarring. It would be good to add a behavior to the icon's `y` value, so that it animates nicely. Something like the following will match the animation of ValueButton:

`Behavior on y { NumberAnimation{ easing.type: Easing.InOutQuad; duration: valueButton._duration } }`

Note that you'll need to set `y` explicitly for this to work, rather than setting the `verticalCenter`.

llewelld (changes requested):
Thanks, this is looking better. I've added some further comments.

apps/browser/qml/pages/components/BrowserListItem.qml
atatarov: what should be` y`? `Y` position is not clear to me .. can it be calculated correctly

llewelld: Currently the verticalCenter value is equivalent to this, I think:
```
y: valueButton.y + (valueButton.height + height) / 2
```

apps/browser/qml/pages/SettingsPage.qml
llewelld: New bug added to cover this; thanks @atatarov; I'm resolving this thread.

llewelld (approved):
This LGTM.

PR #786, opened 2021-02-12
[browser] Remove obsoleted BrowserSettings. Fixes JB#53137

rainemak:


apps/shared/WebView.qml
```diff
@@ -61,13 +61,7 @@ WebContainer {
...
     property var linkHandler: LinkHandler {}
 
-    property var searchEngine: SearchEngine {
-        property var browserSettings: BrowserSettings {}
-
-        function searchEngines() {
-            return browserSettings.searchEngineList;
-        }
-    }
+    property var searchEngine: SearchEngine {}
```
rainemak, line 64: We shouldn't need this here either as we could installing of a search engine view search engine model as well. That said, leaving it here for the time being.

mkenttala (approved):

PR #787, opened 2021-02-12
[browser] Refactored OpenSearch engine download support. Contributes to JB#40380

mkenttala:
…to JB#40380


apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -82,7 +82,11 @@ Page {
 
                 onCurrentItemChanged: {
```
rainemak, line 83: May I propose to actually move this to the delegate itself.

Feel free to adapt from here:
https://github.com/sailfishos/sailfish-browser/commit/6d7e1fc4eadbc7b41255d8ad625ef12056b354da#diff-c27387fb9a98ac07eb836e3e56d5275ebc64fb1e685ab856f63946d4303e1c08R85

There are already roles regarding item being installed or not. Also url role is there.

So MenuItem click handled could be turned into somehting like:

```
onClicked: {
    if (title !== searchEngineConfig.value) {
        if (installed) {
            searchEngineConfig.value = title
        } else {
            searchFetcher.fetch(url)
        }
    }
}

```
No need for isInstalled and url invokables.

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -166,4 +170,15 @@ Page {
         key: "/apps/sailfish-browser/settings/home_page"
         defaultValue: "http://jolla.com/"
     }
+
+    DataFetcher {
```
rainemak, line 174: I'd prefer seeing "install" invokable on the searchenginemodel which would be using datafetcher.

As after installing it, you need to update the datamodel as well.

apps/shared/WebView.qml
```diff
@@ -319,13 +317,9 @@ WebContainer {
...
                 }
                 case "Link:AddSearch": {
-                    // TODO: proper implementation
-                    /*if (searchEngine.searchEngines().indexOf(data.engine.title)) {
-                        searchEngine.url = data.url
-                        searchEngine.href = data.engine.href
-                        searchEngine.title = data.engine.title
-                        searchEngine.addSearchEngine()
-                    }*/
+                    if (SearchEngineModel.searchEngines.indexOf(data.engine.title)) {
```
rainemak, line 320: SearchEngineModel.contains(data.engine) instead. No, need to expose that old string list.

apps/browser/settings/searchenginemodel.cpp
```diff
@@ -68,3 +68,46 @@ void SearchEngineModel::componentComplete()
...
+    QStringList engines;
+    for (int i = 0; i < m_searchEngines.count(); i++)
+        engines.append(m_searchEngines[i].title);
+    return engines;
+}
+
+void SearchEngineModel::addSearchEngine(const QString& title, const QString& url)
+{
+    bool found = false;
+    for (int i = 0; i < m_searchEngines.count(); i++) {
```
rainemak, line 83:
```
for (const SearchEngine &searchEngine : m_searchEngines) {
    if (searchEngine.title == title) {
        found = true;
        break;
    }
}
...
```

apps/browser/settings/searchenginemodel.cpp
```diff
@@ -68,3 +68,46 @@ void SearchEngineModel::componentComplete()
...
+
+QStringList SearchEngineModel::searchEngines()
+{
+    QStringList engines;
+    for (int i = 0; i < m_searchEngines.count(); i++)
+        engines.append(m_searchEngines[i].title);
+    return engines;
+}
+
+void SearchEngineModel::addSearchEngine(const QString& title, const QString& url)
```
rainemak, line 80: Just call this "add"

apps/browser/settings/searchenginemodel.cpp
```diff
@@ -68,3 +68,46 @@ void SearchEngineModel::componentComplete()
...
+{
+    bool found = false;
+    for (int i = 0; i < m_searchEngines.count(); i++) {
+        if (m_searchEngines[i].title == title) {
+            found = true;
+            break;
+        }
+    }
+    if (!found) {
+        SearchEngine engine(url, title, false);
```
rainemak, line 90: SearchEngine engine(url, title, false);
beginInsertRows(QModelIndex(), rowCount(), rowCount());
m_searchEngines.append(engine);
endInsertRows();
emit countChanged();

apps/browser/settings/searchenginemodel.cpp
```diff
@@ -68,3 +68,46 @@ void SearchEngineModel::componentComplete()
...
+        m_searchEngines.append(engine);
+        emit countChanged();
+        emit searchEnginesChanged();
```
rainemak, line 93: As the searchEngines string list is only for "contains" => this should not be needed either.

apps/shared/WebView.qml
rainemak: Actually, addSearchEngine check as well if it's there already and adds if not found.

rainemak (changes requested):
Quite nice already. Requires a bit of tweaking.

Let's keep WebView's call to add to the model still in comments as the combobox is not yet there.

case "Link:AddSearch": {
    // SearchEngineModel.add...
}

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -67,22 +67,26 @@ Page {
...
 
                 onCurrentItemChanged: {
                     if (currentItem.text !== searchEngineConfig.value) {
-                        searchEngineConfig.value = currentItem.text
+                        //TODO: downloaded is not updated until settings page is reopened
+                        if (currentItem.downloaded) {
+                            searchEngineConfig.value = currentItem.text
+                        } else {
+                            SearchEngineModel.install(currentItem.text)
+                        }
```
rainemak, line 89: I'd still move this whole onCurrentItem change log to the onClicked handler of the MenuItem component itself. There's no need to have this handler here as all the steps can be done already on onClicked inside the delegate.

apps/browser/qml/pages/SettingsPage.qml
mkenttala: Had problems using installed property here, it doesn't change until page is reloaded. Might be a problem with my dataChanged signal.

apps/browser/qml/pages/SettingsPage.qml
mkenttala: Yep, that move solved my problems.

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -166,4 +169,10 @@ Page {
         key: "/apps/sailfish-browser/settings/home_page"
         defaultValue: "http://jolla.com/"
     }
+
+    Connections {
+        target: SearchEngineModel
+
+        onInstalled: searchEngineConfig.value = title
```
rainemak, line 176: Even this could be written to the dconf value directly on the C++ side. However, I think it's good already like this.

This Connections can be improved in another task. Will file one.

apps/browser/settings/searchenginemodel.cpp
```diff
@@ -68,3 +69,40 @@ void SearchEngineModel::componentComplete()
 {
 
 }
+
+void SearchEngineModel::add(const QString& title, const QString& url)
+{
+    for (SearchEngine engine : m_searchEngines) {
```
rainemak, line 75: This should be like "const SearchEngine &engine : m_searchEngines" => now there will be copied.

apps/browser/settings/searchenginemodel.cpp
```diff
@@ -68,3 +69,40 @@ void SearchEngineModel::componentComplete()
...
+    SearchEngine engine(url, title, false);
+    beginInsertRows(QModelIndex(), rowCount(), rowCount());
+    m_searchEngines.append(engine);
+    endInsertRows();
+    emit countChanged();
+}
+
+void SearchEngineModel::install(const QString& title)
+{
+    for (int i = 0; i < m_searchEngines.count(); i++) {
```
rainemak, line 89:
```suggestion
    for (const SearchEngine &engine : m_searchEngines)
```

apps/browser/settings/searchenginemodel.cpp
```diff
@@ -68,3 +69,40 @@ void SearchEngineModel::componentComplete()
...
+        if (m_searchEngines[i].title == title && !m_searchEngines[i].installed) {
```
rainemak, line 90:
```suggestion
        if (engine.title == title && !engine.installed) {
```

apps/browser/settings/searchenginemodel.cpp
rainemak, line 75:
```suggestion
    for (const SearchEngine &engine : m_searchEngines) {
```

rainemak (approved):
Nice! Minor suggestion still regarding one extra copy on a loop. Check that. LGTMing already.

apps/browser/settings/searchenginemodel.cpp
```diff
@@ -10,6 +10,12 @@
 
 #include "searchenginemodel.h"
 #include "opensearchconfigs.h"
+#include "../bookmarks/datafetcher.h"
+
+#include <QString>
+#include <MGConfItem>
+
+const auto SEARCH_ENGINE_CONFIG = QStringLiteral("/apps/sailfish-browser/settings/search_engine");
```
rainemak, line 18: Regarding this we have JB#49923 but that's another improvement.

PR #792, opened 2021-02-23
[browser] Ensure rendering is resumed when the window becomes visible. Contributes to JB#53152

adenexter:
Rendering can be suspended by means other than setting readyToPaint to
false so the previous value has limited meaning.


rainemak (approved):
Well spotted! LGTM

PR #793, opened 2021-02-23
[browser] Add translation generation script. Fixes JB#53252

Tomin1:
Add translation generation script for property files. This is intended
to be used when engine is updated and the result can be then committed
to this repository. The script outputs a single C++ source file from
input files.

### DIFF
--- a/tools/generate_property_translations.py
+++ b/tools/generate_property_translations.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python3
+#
+# Generates translation strings from gecko property files for lupdate
+# to process
+
+import os, os.path, sys, re, dataclasses, functools
+
+PREFIX = "sailfish_components_webview_popups-la"
+
+class Generator:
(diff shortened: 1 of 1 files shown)


tools/generate_property_translations.py
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env python3
+#
+# Generates translation strings from gecko property files for lupdate
+# to process
+
+import os, os.path, sys, re, dataclasses, functools
+
+PREFIX = "sailfish_components_webview_popups-la"
```
rainemak, line 8: This is correct prefix... but the script itself should be pushed to the sailfish-components-webview repository. Similarly under tools I'd say.

Will give a spin for this one already.

Tomin1: Right, thanks! I opened a new PR to the correct repository.

Tomin1:
Replaced by https://github.com/sailfishos/sailfish-components-webview/pull/92.

PR #794, opened 2021-02-24
[sailfish-browser] Add icon for open-url .desktop entry. Contributes to JB#44657

pvuorela:
Can be needed if there are multiple apps handling a url opening
and the user is presented a list to choose the app.

See also https://git.sailfishos.org/mer-core/libcontentaction/merge_requests/18

@rainemak


rainemak (approved):
LGTM

PR #795, opened 2021-02-24
[sailfish-browser] Handle opening dialog when download click. Contributes to JB#51673

atatarov:


apps/browser/qml/pages/DownloadPickerPage.qml
```diff
@@ -0,0 +1,13 @@
+import QtQuick 2.0
+import Sailfish.Silica 1.0
+import Sailfish.Pickers 1.0
+
+FolderPickerPage {
```
llewelld, line 5: This page looks redundant to me: you can connect directly to `onSelectedPathChanged` in the BrowserPage.qml code to avoid the need for the additional `selected` signal (and then use `page.selectedPath` instead of passing `path`.

llewelld (changes requested):

apps/browser/qml/pages/BrowserPage.qml
```diff
@@ -17,6 +17,7 @@ import Sailfish.Silica 1.0
 import Sailfish.Silica.private 1.0 as Private
 import Sailfish.Browser 1.0
 import Sailfish.Policy 1.0
+import Sailfish.WebEngine 1.0
```
llewelld, line 20: Copyright header needs updating for 2021.

llewelld (approved):
This looks great, thank you!

llewelld:
As on the other commits, ideally the dialog should open in the same place as the previous safe. Let me know if I should merge this directly, or if you prefer to address that first.

PR #797, opened 2021-03-01
[sailfish-browser] Adjust historylistitem for correct display of search results. Fixes JB#53177

aroditelev:

PR #798, opened 2021-03-01
[sailfish-browser] Implement bookmark search only when user has entered text to the search field. Fixes JB#53174

aroditelev:


apps/browser/bookmarks/bookmarkfiltermodel.h
```diff
@@ -17,21 +17,28 @@ class BookmarkFilterModel : public QSortFilterProxyModel
...
+    void setMaxDisplayedItems(const int maxDisplayedItems);
 
 signals:
     void searchChanged(QString search);
+    void maxDisplayedItemsChanged(int maxDisplayedItems);
 
 private:
     QString m_search;
+    int m_maxDisplayedItems;
+    mutable int m_countFilterAccepts;
```
xlise, line 41: Maybe you should add initialization

aroditelev: This is done by adding source model

apps/browser/qml/pages/components/Overlay.qml
```diff
@@ -541,11 +541,18 @@ Shared.Background {
...
                     id: bookmarkModel
                     activeUrl: toolBar.url
                 }
 
+                model: BookmarkFilterModel {
+                    id: bookmarkFilterModel
+                    sourceModel: bookmarkModel
+                    maxDisplayedItems: search ? favoriteGrid.columns : bookmarkModel.count
+                    search: searchField.text !== webView.url ? searchField.text : ""
+                }
```
llewelld, line 554: When searching, the `FavoriteGrid` is immediately hidden. Should there be some change to [`historyContainer.showFavorites` on line 223](https://github.com/sailfishos/sailfish-browser/pull/798/files#diff-2ddfbf1787333eb21930f77d91dc8b92fc930d6dd5eb8d801df0c9499fe06bf8R223) to avoid this?

apps/browser/qml/pages/components/Overlay.qml
```diff
@@ -541,11 +541,18 @@ Shared.Background {
...
+                BookmarkModel {
                     id: bookmarkModel
                     activeUrl: toolBar.url
                 }
 
+                model: BookmarkFilterModel {
+                    id: bookmarkFilterModel
+                    sourceModel: bookmarkModel
+                    maxDisplayedItems: search ? favoriteGrid.columns : bookmarkModel.count
+                    search: searchField.text !== webView.url ? searchField.text : ""
```
llewelld, line 553: Would it make more sense to use `searchField.edited ? searchField.text : ""` here instead?

apps/browser/bookmarks/bookmarkfiltermodel.cpp
```diff
@@ -28,24 +28,53 @@ bool BookmarkFilterModel::filterAcceptsRow(int sourceRow, const QModelIndex &sou
...
     return false;
 }
 
+void BookmarkFilterModel::setSourceModel(QAbstractItemModel *sourceModel)
+{
+    if (sourceModel) {
+        m_countFilterAccepts = 0;
+        m_maxDisplayedItems = sourceModel->rowCount();
+        QSortFilterProxyModel::setSourceModel(sourceModel);
+        invalidate();
```
llewelld, line 46: According to [the docs](https://doc.qt.io/qt-5/qabstractproxymodel.html#setSourceModel):
> Subclasses should call beginResetModel() at the beginning of the method, disconnect from the old model, call this method, connect to the new model, and call endResetModel().

so presumably you should be doing this here rather than invalidating.

apps/browser/bookmarks/bookmarkfiltermodel.cpp
```diff
@@ -28,24 +28,53 @@ bool BookmarkFilterModel::filterAcceptsRow(int sourceRow, const QModelIndex &sou
 {
     QModelIndex index = sourceModel()->index(sourceRow, 0, sourceParent);
 
-    if (sourceModel()->data(index, DeclarativeBookmarkModel::UrlRole).toString().trimmed().contains(search(), Qt::CaseInsensitive)
-            || sourceModel()->data(index, DeclarativeBookmarkModel::TitleRole).toString().trimmed().contains(search(), Qt::CaseInsensitive)) {
+    if ((sourceModel()->data(index, DeclarativeBookmarkModel::UrlRole).toString().trimmed().contains(search(), Qt::CaseInsensitive)
+         || sourceModel()->data(index, DeclarativeBookmarkModel::TitleRole).toString().trimmed().contains(search(), Qt::CaseInsensitive))
+         && m_countFilterAccepts < m_maxDisplayedItems) {
```
llewelld, line 33: It's likely to be more efficient to move the `m_countFilterAccepts < m_maxDisplayedItems` clause to the start of the condition, and allow the heavier string comparisons to be short-circuited.

You could also use `m_search` directly rather than calling `search()` each time.

apps/browser/bookmarks/bookmarkfiltermodel.cpp
```diff
@@ -28,24 +28,53 @@ bool BookmarkFilterModel::filterAcceptsRow(int sourceRow, const QModelIndex &sou
...
+}
+
 void BookmarkFilterModel::setSearch(const QString &search)
 {
     if (m_search == search)
         return;
 
+    m_countFilterAccepts = 0;
     m_search = search;
+    invalidate();
```
llewelld, line 67: Rather than invalidating the entire model each time the search term changes, it would be better to perform a diff and only update the rows that have been added/removed. For example see the code in [`DeclarativeHistoryModel::updateModel()`](https://github.com/sailfishos/sailfish-browser/blob/1cd9c437f411a9159c7f1551f339522a61a6e2d0/apps/history/declarativehistorymodel.cpp#L113).

apps/browser/bookmarks/bookmarkfiltermodel.cpp
```diff
@@ -28,24 +28,53 @@ bool BookmarkFilterModel::filterAcceptsRow(int sourceRow, const QModelIndex &sou
...
+}
+
+void BookmarkFilterModel::setMaxDisplayedItems(const int maxDisplayedItems)
+{
+    if (m_maxDisplayedItems == maxDisplayedItems)
+        return;
+
+    m_countFilterAccepts = 0;
+    m_maxDisplayedItems = maxDisplayedItems;
     invalidate();
```
llewelld, line 78: Similarly, if `maxDisplayedItems` changes you will only ever be adding/remove rows to/from the end of the model, so it would be better not to invalidate the entire model.

llewelld (changes requested):
I have some queries here; two things in particular. First I don't actually see the bookmark filter in use; is there some other PR needed to enable it, or should this PR include the changes required for that too? Second it looks like the filtering could be performed without invalidating the model each time. I've added some comments related to this in the code.

aroditelev:
> I have some queries here; two things in particular. First I don't actually see the bookmark filter in use; is there some other PR needed to enable it, or should this PR include the changes required for that too? Second it looks like the filtering could be performed without invalidating the model each time. I've added some comments related to this in the code.

bookmark filter in use: https://github.com/aroditelev/sailfish-browser/blob/4a4c4802fa50b5a85998bb51f428d251692a5b77/apps/browser/qml/pages/components/Overlay.qml#L549

llewelld:
> bookmark filter in use: https://github.com/aroditelev/sailfish-browser/blob/4a4c4802fa50b5a85998bb51f428d251692a5b77/apps/browser/qml/pages/components/Overlay.qml#L549

Thanks @aroditelev. I see the model in the code, but in practice it doesn't show because `historyContainer.showFavorites` hides it as soon as I start typing (see my comment [here](https://github.com/sailfishos/sailfish-browser/pull/798#discussion_r589487257)).

aroditelev:
> > bookmark filter in use: https://github.com/aroditelev/sailfish-browser/blob/4a4c4802fa50b5a85998bb51f428d251692a5b77/apps/browser/qml/pages/components/Overlay.qml#L549
> 
> Thanks @aroditelev. I see the model in the code, but in practice it doesn't show because `historyContainer.showFavorites` hides it as soon as I start typing (see my comment [here](https://github.com/sailfishos/sailfish-browser/pull/798#discussion_r589487257)).

@llewelld to display model when searching, you must also use https://github.com/sailfishos/sailfish-browser/pull/797, I specifically set the order from 1 to 4 in the chat.Initially, these changes were in one pr, but @rainemak asked to split them into 4 pr.

llewelld:
> @llewelld to display model when searching, you must also use #797, I specifically set the order from 1 to 4 in the chat.Initially, these changes were in one pr, but @rainemak asked to split them into 4 pr.

Thanks for the clarification, but I'm unable to rebase #797 onto #798 (or vice versa) as I get conflicts when I try to do that.

apps/browser/bookmarks/bookmarkfiltermodel.cpp
aroditelev: @llewelld  `invalidate()` it was before the changes https://github.com/sailfishos/sailfish-browser/blob/1cd9c437f411a9159c7f1551f339522a61a6e2d0/apps/browser/bookmarks/bookmarkfiltermodel.cpp#L50

aroditelev:
> I have some queries here; two things in particular. First I don't actually see the bookmark filter in use; is there some other PR needed to enable it, or should this PR include the changes required for that too? Second it looks like the filtering could be performed without invalidating the model each time. I've added some comments related to this in the code.

The order of changes is presented here:  https://github.com/sailfishos/sailfish-browser/pull/783#discussion_r574574086

apps/browser/bookmarks/bookmarkfiltermodel.cpp
llewelld: It looks like `invalidateFilter()` would apply here too.

apps/browser/bookmarks/bookmarkfiltermodel.cpp
llewelld: Apologies for not explaining clearly. As I understand it, using `QAbstractListModel` would allow the model to be manipulated by Adding/Removing rows rather than invalidating the entire model each time the search term changes. This is the approach that's used for the [history model](https://github.com/sailfishos/sailfish-browser/blob/1cd9c437f411a9159c7f1551f339522a61a6e2d0/apps/history/declarativehistorymodel.cpp) for example. In that case, when the search term changes, a diff is performed in the `updateModel()` method and only changes are emitted.

It would be nice to do the same with the bookmarks, but it's not essential for your changes here.

rainemak: @llewelld you mind filing a task about it. Would be good to get it fixed for the next release. Should not be very big task to implement.

aroditelev: @llewelld  `invalidateFilter ()` aren't these operations optimized? Why create another `QAbstractListModel` that will hold the filtered items.  `QSortFilterProxyMode` for this it is also intended.

llewelld: Yes, you're right @aroditelev and I was mistaken. If you use `invalidateFilter()` it'll be optimised, so in that case there's only limited benefit to changing to `QAbstractListModel()` (`QSortFilterProxyModel` doesn't know when it's a narrowing of the search term, but I guess the benefit that can be gained from that is small).

apps/browser/bookmarks/bookmarkfiltermodel.cpp
aroditelev: Done

apps/browser/bookmarks/bookmarkfiltermodel.cpp
aroditelev: Done

llewelld:
This is looking really nice now, thank you. I'm afraid I noticed one last issue that I'd not noticed before.

Currently you assume that `BookmarkFilterModel::filterAcceptsRow()` will be called just once for each row as the model is built up. You need this in order for `m_countFilterAccepts` to represent the number of items shown in the model.

Unfortunately `QSortFilterProxyModel` will also calle `filterAcceptsRow()` on item insertions (see [here](https://git.sailfishos.org/mer-core/qtbase/blob/43a7947c26b79e9976e82957d3a0c48cf6e7596c/src/corelib/itemmodels/qsortfilterproxymodel.cpp#L790)) and dataChange events (see [here](https://git.sailfishos.org/mer-core/qtbase/blob/43a7947c26b79e9976e82957d3a0c48cf6e7596c/src/corelib/itemmodels/qsortfilterproxymodel.cpp#L1176)).

When a bookmark is added the count is updated, which invalidates the filter, but also triggers a call to `DeclarativeBookmarkModel::updateFavoriteIcon()`, which itself then emits `dataChanged()`. As a consequence of all this, when a bookmark is added the count gets messed up, and so the new bookmark doesn't appear in the bookmark grid (at least, not until the user enters a character and invalidates the filter).

I can see a couple of easy solutions:
1. Invalidate the filter on a `dataChanged()` signal from the source model.
2. Set `QSortFilterProxyModel::dynamicSortFilter` to false.

I'm not sure whether either of these will have other side effects though, and you may have a better idea.

If you prefer we can merge this as it is and file a separate bug about bookmark additions not appearing?

llewelld (approved):
> If you prefer we can merge this as it is and file a separate bug about bookmark additions not appearing?

We agreed to to this route and create a separate PR to address this. Therefore approving this.

atatarov (approved):

PR #799, opened 2021-03-01
[sailfish-browser] Add history item below the search field. Fixes JB#53171

aroditelev:
https://bz.jollamobile.com/show_bug.cgi?id=53171

### DIFF
--- a/apps/browser/qml/pages/components/Overlay.qml
+++ b/apps/browser/qml/pages/components/Overlay.qml
@@ -313,10 +313,10 @@ Shared.Background {
                 }
 
                 // Follow grid / list position.
-                y: -((historyContainer.showFavorites ? favoriteGrid.contentY : historyList.contentY) + height)
+                y: -((historyContainer.showFavorites ? favoriteGrid.contentY : historyList.contentY) + (historyContainer.showFavorites ? favoriteGrid.headerItem.height : height))
                 // On top of HistoryList and FavoriteGrid
                 z: 1
-                height: Theme.itemSizeMedium
+                height: toolBar.rowHeight
                 textLeftMargin: Theme.paddingLarge
                 textRightMargin: Theme.paddingLarge
                 focusOutBehavior: FocusBehavior.ClearPageFocus
@@ -394,6 +394,28 @@ Shared.Background {
                     if (!edited && text !== webView.url) {
                         edited = true
                     }
+                }
+            }
+
+            OverlayListItem {
+                id: historyButton
+                height: toolBar.rowHeight
+                iconWidth: toolBar.iconWidth
+                horizontalOffset: toolBar.horizontalOffset
+                // Follow grid / list position.
+                y: -((historyContainer.showFavorites ? -searchField.y : historyList.contentY) - height)
+                // On top of HistoryList and FavoriteGrid
+                z: 1
+
+                text: qsTrId("sailfish_browser-la-history")
+                iconSource: "image://theme/icon-m-history"
+                visible: historyContainer.showFavorites
+                opacity: visible && toolBar.opacity < 0.9 ? 1.0 : 0.0
+                enabled: overlayAnimator.atTop
+
+                onClicked: {
+                    var historyPage = pageStack.push("../HistoryPage.qml", { model: historyModel })
+                    historyPage.loadPage.connect(loadPage)
                 }
             }
 
@@ -549,7 +571,7 @@ Shared.Background {
 
                 header: Item {
                     width: parent.width
-                    height: searchField.height
+                    height: searchField.height + historyButton.height
                 }
 
                 model: BookmarkModel {


apps/browser/qml/pages/components/Overlay.qml
```diff
@@ -321,7 +321,11 @@ Shared.Background {
                 textTopMargin: height/2 - _editor.implicitHeight/2
                 labelVisible: false
                 inputMethodHints: Qt.ImhUrlCharactersOnly
-                background: null
+                background: Rectangle{
```
xlise, line 324:
```suggestion
                background: Rectangle {
```

apps/browser/qml/pages/components/Overlay.qml
```diff
@@ -530,7 +558,8 @@ Shared.Background {
             Browser.FavoriteGrid {
                 id: favoriteGrid
 
-                height: historyList.height
+                y: historyButton.height
+                height: historyList.height - historyButton.height
```
llewelld, line 562: The underlying `IconGridViewBase` component adds extra margins to the left and right hand side, larger than the margin for other components in the toobar. This is accentuated with the addition of the History button.

Could we maybe have the `horizontalMargin: Theme.paddingLarge` set here so that they align?

apps/browser/qml/pages/components/Overlay.qml
```diff
@@ -386,6 +390,30 @@ Shared.Background {
...
+
+                text: qsTrId("sailfish_browser-la-history")
+                iconSource: "image://theme/icon-m-history"
+                opacity: visible && toolBar.opacity < 0.9 ? 1.0 : 0.0
+                enabled: overlayAnimator.atTop
+                visible: historyContainer.showFavorites && !historyContainer.showHistoryList
+
+                onClicked: {
+                    var historyPage = pageStack.push("../HistoryPage.qml", { model: historyModel })
+                    historyPage.loadPage.connect(loadPage)
```
llewelld, line 411: When cancelling the history page (i.e. swipe left), the toolbar has descended, and it feels a bit strange returning to the webpage when previously the start page was open (when going back, I expect to be returned to the same place I cam from). Is it possible to have the overlay remain opening when swiping back from the history page?

apps/browser/qml/pages/components/Overlay.qml
```diff
@@ -321,7 +321,11 @@ Shared.Background {
                 textTopMargin: height/2 - _editor.implicitHeight/2
                 labelVisible: false
                 inputMethodHints: Qt.ImhUrlCharactersOnly
-                background: null
+                background: Rectangle {
```
llewelld, line 324: Just as a note, this will conflict with the changes in #800.

llewelld (changes requested):
This is looking nice; just a few minor comments.

apps/browser/qml/pages/components/Overlay.qml
aroditelev: @llewelld   before the changes,  behavior was the same, is there no such behavior in the design?

llewelld: I'm don't think the design goes in to this level of detail. You're right though that other users have the same behaviour as this. It's just especially pronounced in this case because the toolbar has grown to take up so much of the screen, so it seems like a separate page. Let me double check on this.

apps/browser/qml/pages/components/Overlay.qml
```diff
@@ -386,6 +390,30 @@ Shared.Background {
...
 
+            OverlayListItem {
+                id: historyButton
+                height: toolBar.rowHeight
+                iconWidth: toolBar.iconWidth
+                horizontalOffset: toolBar.horizontalOffset
+
+                y: -((historyContainer.showFavorites ? -searchField.y : historyList.contentY) - height)
+                // On top of HistoryList and FavoriteGrid
+                z: 1
```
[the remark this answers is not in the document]
aroditelev: @llewelld  `visible` of this element depends on whether the query is entered in the searchField

apps/browser/qml/pages/components/Overlay.qml
aroditelev: @llewelld they are not on the same level in design:
https://www.figma.com/file/boHTFFYJBsRQjZQjWispiJ/Browser?node-id=3743%3A1432

apps/browser/qml/pages/components/Overlay.qml
aroditelev: done

apps/browser/qml/pages/components/Overlay.qml
aroditelev: But adding `horizontalMargin: Theme.paddingLarge` makes them look more like the design. Thanks.

llewelld (changes requested):
Thank you for making the changes. After trying out your latest version some more, it's become clear why the extra margin on the grid is needed. Please see the screenshot, where you can see the text extending outside the screen bounds on the left hand side.

I'm afraid I think you'll need to revert this to the spacing from your original version.

browser

aroditelev:
> Thank you for making the changes. After trying out your latest version some more, it's become clear why the extra margin on the grid is needed. Please see the screenshot, where you can see the text extending outside the screen bounds on the left hand side.
> 
> I'm afraid I think you'll need to revert this to the spacing from your original version.
> 
> browser

@llewelld  Removed custom horizontalMargin

llewelld (changes requested):
This is looking great. There's just one minor adjustment needed: currently the background highlight of the History button is offset horizontally, but it should stretch across the width of the parent.

aroditelev:
> This is looking great. There's just one minor adjustment needed: currently the background highlight of the History button is offset horizontally, but it should stretch across the width of the parent.

@llewelld  lthese spaces are related to the wrapping of the button in the header  `FavoriteGrid`

llewelld (approved):
Thank you, this is looking and acting nicely. Looks good to me.

PR #800, opened 2021-03-01
[sailfish-browser] Make overlay to open to top (y: 0) if no tabs exists. Fixes JB#53170

aroditelev:
https://bz.jollamobile.com/show_bug.cgi?id=53170

### DIFF
--- a/apps/browser/qml/pages/BrowserPage.qml
+++ b/apps/browser/qml/pages/BrowserPage.qml
@@ -210,7 +210,7 @@ Page {
         target: AccessPolicy.browserEnabled && webView && webView.tabModel || null
         ignoreUnknownSignals: true
         // Animate overlay to top if needed.
-        onCountChanged: webView.handleModelChanges(false)
+        onCountChanged: webView.tabModel.count !== 0 ? overlay.animator.showChrome() : webView.handleModelChanges(false)
         onWaitingForNewTabChanged: window.opaqueBackground = webView.tabModel.waitingForNewTab
     }
 
@@ -295,7 +295,11 @@ Page {
         onActiveChanged: {
             var isFullScreen = webView.contentItem && webView.contentItem.fullscreen
             if (!isFullScreen && active && !overlay.enteringNewTabUrl) {
-                overlay.animator.showChrome()
+                if (webView.tabModel.count !== 0 || (WebUtils.homePage !== "about:blank" && WebUtils.homePage.length > 0)) {
+                    overlay.animator.showChrome()
+                } else {
+                    overlay.startPage()
+                }
             }
 
             if (!active) {
--- a/apps/browser/qml/pages/components/Overlay.qml
+++ b/apps/browser/qml/pages/components/Overlay.qml
@@ -1,6 +1,6 @@
 /*
  * Copyright (c) 2014 - 2019 Jolla Ltd.
- * Copyright (c) 2019 Open Mobile Platform LLC.
+ * Copyright (c) 2019 - 2021 Open Mobile Platform LLC.
  *
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
@@ -40,7 +40,7 @@ Shared.Background {
     property bool _showUrlEntry
     readonly property bool _topGap: _showUrlEntry || _showFindInPage
 
-    function loadPage(url)  {
+    function loadPage(url) {
         if (url == "about:config") {
             pageStack.animatorPush(Qt.resolvedUrl("ConfigWarning.qml"), {"browserPage": browserPage})
         } else if (url == "about:settings") {
@@ -74,6 +74,13 @@ Shared.Background {
         _overlayHeight = Qt.binding(function () { return overlayAnimator._fullHeight })
         searchField.resetUrl("")
         overlayAnimator.showOverlay(action === PageStackAction.Immediate)
+    }
+
+    function startPage() {
+        searchField.enteringNewTabUrl = true
+        _showUrlEntry = true
+        searchField.resetUrl("")
+        overlay.animator.updateState("startPage", PageStackAction.Immediate)
     }
 
     function dismiss(canShowChrome, immediate) {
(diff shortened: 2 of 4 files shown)


apps/shared/OverlayAnimator.qml
```diff
@@ -162,6 +167,22 @@ Item {
             ]
         },
 
+        State {
+            name: _startPage
+            changes: [
+                PropertyChanges {
+                    target: overlay
+                    y: webView.privateMode ? _fullHeight : 0
+                    height: webView.fullscreenHeight
```
xlise, line 176: May be write values in other states

apps/browser/qml/pages/components/Overlay.qml
```diff
@@ -321,7 +328,11 @@ Shared.Background {
                 textTopMargin: height/2 - _editor.implicitHeight/2
                 labelVisible: false
                 inputMethodHints: Qt.ImhUrlCharactersOnly
-                background: null
+                background: Rectangle{
```
xlise, line 331:
```suggestion
                background: Rectangle {
```

apps/shared/OverlayAnimator.qml
```diff
@@ -190,15 +211,15 @@ Item {
...
+            to: "fullscreenWebPage,chromeVisible,loadProgressOverlay,fullscreenOverlay,noOverlay,doubleToolBar,certOverlay,startPage"
 
             SequentialAnimation {
                 NumberAnimation { target: webView; property: "height"; duration: transitionDuration; easing.type: Easing.InOutQuad }
                 ScriptAction {
                     script: {
                         if (animator.state === _chromeVisible || animator.state === _fullscreenWebPage || animator.state === _doubleToolBar) {
                             atBottom = true
-                        } else if (animator.state === _fullscreenOverlay || animator.state === _certOverlay) {
+                        } else if (animator.state === _fullscreenOverlay || animator.state === _certOverlay ||animator.state === _startPage ) {
```
xlise, line 222:
```suggestion
                        } else if (animator.state === _fullscreenOverlay || animator.state === _certOverlay || animator.state === _startPage ) {
```

llewelld (commented):
This looks and works nicely generally. I have a few minor style comments.

One issue in relation to behaviour is that, on opening the browser when there are no existing tabs, the start page opens full screen. Shortly after the homepage loads and the start page disappears. Is this something that's been fixed elsewhere already? It looks like it's related to [this code](https://github.com/sailfishos/sailfish-browser/blob/master/apps/core/declarativewebcontainer.cpp#L902).

aroditelev:
> This looks and works nicely generally. I have a few minor style comments.
> 
> One issue in relation to behaviour is that, on opening the browser when there are no existing tabs, the start page opens full screen. Shortly after the homepage loads and the start page disappears. Is this something that's been fixed elsewhere already? It looks like it's related to [this code](https://github.com/sailfishos/sailfish-browser/blob/master/apps/core/declarativewebcontainer.cpp#L902).

Если в настройках оставить пустое поле домашней страницы, то будет:
void DeclarativeWebContainer::initialize() homePage... "about:blank"
Если задать домашнюю страницу, то :

apps/browser/qml/pages/BrowserPage.qml
```diff
@@ -294,7 +294,11 @@ Page {
         onActiveChanged: {
             var isFullScreen = webView.contentItem && webView.contentItem.fullscreen
             if (!isFullScreen && active && !overlay.enteringNewTabUrl) {
-                overlay.animator.showChrome()
+                if (webView.tabModel.count !== 0 || (WebUtils.homePage !== "about:blank" && WebUtils.homePage.length > 0)) {
```
aroditelev, line 297: @llewelld 
> This looks and works nicely generally. I have a few minor style comments.
> 
> One issue in relation to behaviour is that, on opening the browser when there are no existing tabs, the start page opens full screen. Shortly after the homepage loads and the start page disappears. Is this something that's been fixed elsewhere already? It looks like it's related to [this code](https://github.com/sailfishos/sailfish-browser/blob/master/apps/core/declarativewebcontainer.cpp#L902).

This change resolves the issue when using the home page.

llewelld: Great, this is much better. Thank you.

llewelld (approved):
This looks good to me. There is one further point I note, which is that it would be better if the start page opened fully after the user selects "Open new tab" (especially if there are no open tabs, since in this case it can't be dragged downwards anyway). Would this be easy to implemented?

If this requires notable new work, then I suggest we merge this as it is.

aroditelev:
> This looks good to me. There is one further point I note, which is that it would be better if the start page opened fully after the user selects "Open new tab" (especially if there are no open tabs, since in this case it can't be dragged downwards anyway). Would this be easy to implemented?
> 
> If this requires notable new work, then I suggest we merge this as it is.

@llewelld  This will be relevant only if there is an active tab, if there are no tabs, then overlay will be opened in full screen.

llewelld:
> @llewelld This will be relevant only if there is an active tab, if there are no tabs, then overlay will be opened in full screen.

This works correctly if I swipe back from the tab page, but not if I choose the "Open new tab" option. So, following these steps:
1. Open browser.
2. Press tab button on toolbar.
3. Select "Close all tabs" from context menu (wait for remorse timer to complete).
4. Select "Open new tab" from the context menu.

Now the overlay is not opened full screen.

apps/shared/OverlayAnimator.qml
aroditelev: Done

aroditelev:
> > @llewelld This will be relevant only if there is an active tab, if there are no tabs, then overlay will be opened in full screen.
> 
> This works correctly if I swipe back from the tab page, but not if I choose the "Open new tab" option. So, following these steps:
> 
> 1. Open browser.
> 2. Press tab button on toolbar.
> 3. Select "Close all tabs" from context menu (wait for remorse timer to complete).
> 4. Select "Open new tab" from the context menu.
> 
> Now the overlay is not opened full screen.

@llewelld Thank you, I really did not take into account such a case. Added a condition, now the home page opens when there are no active tabs.

llewelld:
Reaffirming that these changes look good to me. Could you please rebase on next.

PR #803, opened 2021-03-05
[ua] User agent fix for sogou.com. Fixes JB#51734

mkenttala:


rainemak (approved):
If this works, then it works. Change looks good.

PR #804, opened 2021-03-12
[browser] Migrate potential leftover bookmarks. Fixes JB#53083

rainemak:


apps/browser/bookmarks/bookmarkmanager.cpp
```diff
@@ -113,6 +114,45 @@ QList<Bookmark*> BookmarkManager::load() {
...
+    // Cleanup after next stop release. See JB#53083 and JB#52736
+    bookmarkFile = QStandardPaths::writableLocation(QStandardPaths::GenericDataLocation)
+            + QLatin1String("/org.sailfishos/sailfish-browser/bookmarks.json");
+    file.reset(new QFile(bookmarkFile));
+    if (file->exists() && file->open(QIODevice::ReadOnly | QIODevice::Text)) {
+        QJsonDocument doc = QJsonDocument::fromJson(file->readAll());
+        if (doc.isArray()) {
+            QJsonArray array = doc.array();
+            QJsonArray::iterator i;
+            for (i=array.begin(); i != array.end(); ++i) {
```
Tomin1, line 127: Documentation seems to recommend using `QJsonArray::const_iterator` in cases when the array is not modified. Also I would think this could be `for (const QJsonValue value : array)` for example instead of using iterators. Anyways, this is similar to the code above so fine by me like this too.

Tomin1 (approved):
LGTM

apps/browser/bookmarks/bookmarkmanager.cpp
rainemak: Will add a commit on top and cleanup them both.

Tomin1 (approved):

PR #805, opened 2021-03-12
Separate browser and captiveportal translation files

Tomin1:
Use separate translation files for sailfish-browser and
sailfish-captiveportal. They share strings in apps/shared but that
contains only one string at the moment.


apps/captiveportal/main.cpp
```diff
@@ -77,11 +77,11 @@ Q_DECL_EXPORT int main(int argc, char *argv[])
...
-    engineeringEnglish.load("sailfish-browser_eng_en", translationPath);
+    engineeringEnglish.load("sailfish-captiveportal_eng_en", translationPath);
     qApp->installTranslator(&engineeringEnglish);
 
     QTranslator translator;
-    translator.load(QLocale(), "sailfish-browser", "-", translationPath);
+    translator.load(QLocale(), "sailfish-captiveportal", "-", translationPath);
     qApp->installTranslator(&translator);
 
     //% "Browser"
```
rainemak, line 87:
```suggestion
    //% "Captive Portal"
    view->setTitle(qtTrId("sailfish-captiveportal-ap-name"));
```
+ updating sailfish-captiveportal.desktop accordingly.

rainemak: Regarding translation itself if you have suggestions, I'm all ears. It could indicate more towards WLAN login.

rainemak (changes requested):
Separation itself looks good but let's update also translation ids themselves.

Regarding BrowserWindow's sailfish_browser-la-web_browsing we should add "disableByMdmTitle" like property that could then be used in Browser side as it is currently and as "Captive Portal" when used via captive portal. The "activity" renders engineering English text like "Captive Portal disabled by Sailfish OS Device Manager'

apps/captiveportal/main.cpp
Tomin1: Well,
-  WLAN portal,
-  Login portal, or
- Network login portal

even perhaps. That "login portal" could be even "login page" I think but maybe not in this context (app title).

apps/captiveportal/main.cpp
```diff
@@ -77,15 +78,15 @@ Q_DECL_EXPORT int main(int argc, char *argv[])
...
 
     QTranslator translator;
-    translator.load(QLocale(), "sailfish-browser", "-", translationPath);
+    translator.load(QLocale(), "sailfish-captiveportal", "-", translationPath);
     qApp->installTranslator(&translator);
 
-    //% "Browser"
-    view->setTitle(qtTrId("sailfish-browser-ap-name"));
+    //% "Network login portal"
+    view->setTitle(qtTrId("sailfish-captiveportal-ap-name"));
```
rainemak, line 89: :+1:

rainemak (approved):
Rebase before landing if not already done.

PR #807, opened 2021-03-16
[sailfish-browser] Only save working pages to the history. Fixes JB#20436

mkenttala:
…0436

### DIFF
--- a/apps/core/declarativewebcontainer.cpp
+++ b/apps/core/declarativewebcontainer.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013 - 2019 Jolla Ltd.
+ * Copyright (c) 2013 - 2021 Jolla Ltd.
  * Copyright (c) 2019 Open Mobile Platform LLC.
  *
  * This Source Code Form is subject to the terms of the Mozilla Public
@@ -19,6 +19,7 @@
 #include "browserpaths.h"
 #include "browserapp.h"
 #include "logging.h"
+#include "declarativehistorymodel.h"
 
 #include <webengine.h>
 #include <QTimerEvent>
@@ -192,6 +193,16 @@ void DeclarativeWebContainer::setWebPage(DeclarativeWebPage *webPage, bool trigg
                     this, &DeclarativeWebContainer::updateActiveTabRendered, Qt::UniqueConnection);
             connect(m_webPage.data(), &DeclarativeWebPage::domContentLoadedChanged,
                     this, &DeclarativeWebContainer::updateActiveTabRendered, Qt::UniqueConnection);
+
+            connect(m_webPage.data(), &DeclarativeWebPage::neterror, [this]() {
+                if (m_historyModel)
+                    m_historyModel->remove(m_webPage->url().toString());
+            });
+
+            connect(m_webPage.data(), &DeclarativeWebPage::urlChanged, [this]() {
+                if (!BrowserApp::captivePortal() && !m_privateMode && m_historyModel)
+                    m_historyModel->add(m_webPage->url().toString(), QString());
+            });
 
             if (m_webPage->completed() && m_webPage->active() && m_webPage->domContentLoaded()) {
                 m_webPage->update();
@@ -1104,5 +1115,18 @@ void DeclarativeWebContainer::handleContentOrientationChanged(Qt::ScreenOrientat
 {
     if (orientation == pendingWebContentOrientation()) {
         emit webContentOrientationChanged(orientation);
+    }
+}
+
+DeclarativeHistoryModel *DeclarativeWebContainer::historyModel() const
+{
+    return m_historyModel;
+}
+
+void DeclarativeWebContainer::setHistoryModel(DeclarativeHistoryModel *model)
+{
+    if (model != m_historyModel) {
+        m_historyModel = model;
+        emit historyModelChanged();
     }
 }
--- a/apps/core/declarativewebcontainer.h
+++ b/apps/core/declarativewebcontainer.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013 - 2019 Jolla Ltd.
+ * Copyright (c) 2013 - 2021 Jolla Ltd.
  * Copyright (c) 2019 Open Mobile Platform LLC.
  *
  * This Source Code Form is subject to the terms of the Mozilla Public
@@ -31,6 +31,7 @@ class DeclarativeTabModel;
 class DeclarativeWebPage;
 class WebPages;
 class Tab;
+class DeclarativeHistoryModel;
 
 class DeclarativeWebContainer : public QWindow, public QQmlParserStatus, protected QOpenGLFunctions {
     Q_OBJECT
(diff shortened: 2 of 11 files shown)


rainemak (changes requested):
Please give a look to the review comments. There are some issues that need to be addresses before this can be landed.

apps/qtmozembed/declarativewebpage.h
```diff
@@ -88,6 +88,7 @@ class DeclarativeWebPage : public QOpenGLWebPage {
     void toolbarHeightChanged();
     void virtualKeyboardMarginChanged();
     void securityChanged();
+    void contentLoaded();
```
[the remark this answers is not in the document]
mkenttala: domContentLoaded is not useful in this case because it's sent even when page fails.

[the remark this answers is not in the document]
rainemak: But for instance in case of invalid url (server not found) argument is about:neterror from which we can know that this was an error.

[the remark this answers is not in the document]
rainemak: The error state could actually go to the qtmozembed and qmozview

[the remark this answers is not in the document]
rainemak: That would actually imply that domContentLoaded signal should be moved to the qtmozembed as well.

mkenttala:
New version with some of the review comments addressed. Handling every possible way to load (and fail) a page proved quite difficult so there might be some holes left.

rainemak (changes requested):
There's one bigger issue and that's the fact that a tab might disappear after browser restart. Proposed solution on a comment.

apps/qtmozembed/declarativewebpage.cpp
```diff
@@ -217,7 +220,7 @@ void DeclarativeWebPage::restoreHistory() {
 
 void DeclarativeWebPage::setContentLoaded()
 {
-    if (!m_domContentLoaded) {
+    if (!m_domContentLoaded && m_urlReady) {
```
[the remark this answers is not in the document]
mkenttala: This gets called multiple times even when the url() is still the old one, not the url currently loading. Just waiting for the proper url.

apps/history/declarativetabmodel.cpp
```diff
@@ -73,6 +72,19 @@ void DeclarativeTabModel::addTab(const QString& url, const QString &title, int i
     m_nextTabId = tab.tabId() + 1;
 }
 
+void DeclarativeTabModel::saveTab(int tabId)
+{
+    int index = findTabIndex(tabId);
+    if (index >= 0) {
+        if (!m_tabs.at(index).persisted()) {
+            createTab(m_tabs.at(index));
```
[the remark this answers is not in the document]
mkenttala: Yep, now just adding to the history database if page load is ok, tab history remains.

rainemak (commented):
Another round.

apps/history/declarativehistorymodel.cpp
```diff
@@ -66,6 +67,13 @@ void DeclarativeHistoryModel::remove(const QString &url)
     }
 }
 
+void DeclarativeHistoryModel::add(const QString &url, const QString &title)
+{
+    beginInsertRows(QModelIndex(), 0, 1);
```
rainemak, line 72: No need start model row insert here... search will query new set of history entries from database and sorting is handled by `updateModel`.

Just adding should be enough. There's no really means to checking history entries without  entering a search term.

apps/core/declarativewebcontainer.h
```diff
@@ -154,6 +156,9 @@ class DeclarativeWebContainer : public QWindow, public QQmlParserStatus, protect
 
     bool event(QEvent *event);
 
+    DeclarativeHistoryModel *historyModel();
```
rainemak, line 159:
```suggestion
    DeclarativeHistoryModel *historyModel() const;
```

apps/core/declarativewebcontainer.h
```diff
@@ -75,6 +76,7 @@ class DeclarativeWebContainer : public QWindow, public QQmlParserStatus, protect
     Q_PROPERTY(Qt::ScreenOrientation pendingWebContentOrientation READ pendingWebContentOrientation NOTIFY pendingWebContentOrientationChanged FINAL)
 
     Q_PROPERTY(QMozSecurity *security READ security NOTIFY securityChanged)
+    Q_PROPERTY(DeclarativeHistoryModel* historyModel READ historyModel WRITE setHistoryModel)
```
rainemak, line 79: NOTIFY signal missing

apps/history/declarativehistorymodel.cpp
```diff
@@ -66,6 +65,23 @@ void DeclarativeHistoryModel::remove(const QString &url)
     }
 }
 
+void DeclarativeHistoryModel::add(const QString &url, const QString &title)
+{
+    QDateTime time = QDateTime::currentDateTimeUtc();
+    int id = DBManager::instance()->addHistoryEntry(url, title, time);
```
rainemak, line 71: I meant this single line should be enough. When you start typing to the search field, model updates on every single character and dbworker emits historyAvailable.

Problem with this approach is that appending to the end can be wrong place as sorting comes from dbworker. It should be inserted to the right place if any. Hence, I'd not add anything extra here rather just call addHistoryEntry(...).

If you think that's worth to update model here, you could store current search term to member variable (see search) and call search(m_searchTerm);

apps/core/declarativewebcontainer.cpp
```diff
@@ -193,6 +194,16 @@ void DeclarativeWebContainer::setWebPage(DeclarativeWebPage *webPage, bool trigg
...
 
+            connect(m_webPage.data(), &DeclarativeWebPage::neterror, [this]() {
+                if (m_historyModel)
+                    m_historyModel->remove(m_webPage->url().toString());
+            });
+
+            connect(m_webPage.data(), &DeclarativeWebPage::urlChanged, [this]() {
+                if (!BrowserApp::captivePortal() && !m_privateMode && m_historyModel)
+                    m_historyModel->add(m_webPage->url().toString(), m_webPage->title());
+            });
```
rainemak, line 205: This same should be done for the title change. Sounds that update(url, title) would be good for that.

rainemak: Already handled by the historymodel itself.

apps/history/declarativehistorymodel.cpp
```diff
@@ -1,7 +1,6 @@
 /****************************************************************************
 **
-** Copyright (C) 2013 Jolla Ltd.
-** Contact: Petri M. Gerdt <petri.gerdt@jollamobile.com>
+** Copyright (C) 2021 Jolla Ltd.
```
rainemak, line 3:
```suggestion
** Copyright (c) 2013 - 2021 Jolla Ltd.
```

rainemak: Apply similar update for other changed files as well.

rainemak (approved):
Thanks Marko for pushing this through!

PR #809, opened 2021-03-18
[sailfish-browser] Ability to choose start page as default instead of URL. Fixes JB#50678

aroditelev:


apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -40,21 +46,32 @@ Page {
...
+                    MenuItem {
+                        //% "Default home page"
+                        text: qsTrId("sailfish_browser-me-home_page_default")
+                        onClicked: homePageConfig.value = "about:blank"
+                    }
+                    MenuItem {
+                        readonly property string site: removeProtocolTypeFromUri(homePageConfig.value)
+                        //% "Custom website %1"
+                        property string title: homePageConfig.value === "about:blank" ? qsTrId("sailfish_browser-me-home_page_custom").arg("")
+                                                                                      : qsTrId("sailfish_browser-me-home_page_custom").arg(site)
```
xlise, line 68:
```suggestion
                        property string title: qsTrId("sailfish_browser-me-home_page_custom").arg(homePageConfig.value === "about:blank" ? "" : site)
```

aroditelev: Yes it looks better thanks)

apps/browser/qml/pages/components/AddHomePageDialog.qml
```diff
@@ -0,0 +1,63 @@
...
+import QtQuick 2.0
+import Sailfish.Silica 1.0
+import Sailfish.Policy 1.0
+import org.nemomobile.configuration 1.0
+
+Dialog {
+    id: root
+    property alias text: textField.text
+    property alias placeholderText: textField.placeholderText
+    property alias description: textField.description
```
llewelld, line 20: The three aliases and id look unused, so can be removed.

apps/browser/qml/pages/components/AddHomePageDialog.qml
```diff
@@ -0,0 +1,63 @@
...
+
+    function containsSpaces(text) {
+       return /^\s+|\s+$|\s*(\s)/.test(text) && text.length > 0
+    }
+
+    canAccept: (textField.text !== "" && !textField.errorHighlight)
```
llewelld, line 26: As far as I can tell the `textField.text !== "" ` condition is already covered by `containsSpaces()` so can be removed from here.

apps/browser/qml/pages/components/AddHomePageDialog.qml
```diff
@@ -0,0 +1,63 @@
...
+    onAccepted: {
+        homePageConfig.value = textField.text || "about:blank"
+    }
+
+    TextField {
+        id: textField
+
+        width: parent.width
+        anchors.top: header.bottom
+        enabled: AccessPolicy.browserEnabled
```
llewelld, line 44: With the settings pages included as part of the browser, are these AccessPolicy restrictions still needed?

apps/browser/qml/pages/components/AddHomePageDialog.qml
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    onAccepted: {
+        homePageConfig.value = textField.text || "about:blank"
+    }
+
+    TextField {
+        id: textField
+
+        width: parent.width
```
llewelld, line 42: I believe `TextField` defaults to the parent width, so this can be removed.

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -41,21 +47,32 @@ Page {
...
+                iconSource: "image://theme/icon-m-home"
+                menu: ContextMenu {
+                    MenuItem {
+                        //% "Default home page"
+                        text: qsTrId("sailfish_browser-me-home_page_default")
+                        onClicked: homePageConfig.value = "about:blank"
+                    }
+                    MenuItem {
+                        readonly property string site: removeProtocolTypeFromUri(homePageConfig.value)
+                        //% "Custom website %1"
```
llewelld, line 67: It might be helpful for the translators if you could explain what the `%1` will be replaced with.

apps/browser/qml/pages/components/AddHomePageDialog.qml
```diff
@@ -0,0 +1,63 @@
...
+        text: homePageConfig.value === "about:blank" ? "" : homePageConfig.value
+        //% "Enter website address that should be shown when browser is opened with no tabs to load."
+        description: qsTrId("sailfish_browser-he-home_page_website-address")
+        //% "Website address"
+        placeholderText: qsTrId("settings_browser-la-website_address")
+        label: placeholderText
+        inputMethodHints: Qt.ImhNoPredictiveText | Qt.ImhNoAutoUppercase | Qt.ImhUrlCharactersOnly
+        EnterKey.iconSource: "image://theme/icon-m-enter-close"
+        EnterKey.onClicked: focus = false
+        errorHighlight: containsSpaces(text)
```
llewelld, line 54: It's customary to switch out the description with some info about why there's an error in case the dialog can't be accepted. For example, something like "Spaces can't be included in the URL".

Is it necessary to exclude spaces from the URL?

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -41,21 +47,32 @@ Page {
...
-
-                inputMethodHints: Qt.ImhNoPredictiveText | Qt.ImhNoAutoUppercase | Qt.ImhUrlCharactersOnly
-
-                onTextChanged: homePageConfig.value = text || "about:blank"
-
-                EnterKey.iconSource: "image://theme/icon-m-enter-close"
-                EnterKey.onClicked: focus = false
+                //% "Default"
+                value: homePageConfig.value === "about:blank" ? qsTrId("sailfish_browser-la-home_page_default") : removeProtocolTypeFromUri(homePageConfig.value)
+                iconSource: "image://theme/icon-m-home"
```
llewelld, line 58: There's an animation glitch when the item is changed from a long URL, to "Default home page". In this case the little "Home" icon jumps to the floor before moving up to the centre of the item. It must be an issue with `BrowserListItem`; would it be possible to fix it?

llewelld (changes requested):
This is already looking really nice. I've added a few comments.

apps/browser/qml/pages/components/AddHomePageDialog.qml
aroditelev: containsSpaces will be relevant when `text.length > 0`

llewelld: Yes, you're absolutely right, my mistake. Thanks for clarifying.

In this case, maybe the error highlight should show if the textField is empty, in order to clarify to the user why the dialogue can't be accepted. What do you think?

apps/browser/qml/pages/components/AddHomePageDialog.qml
aroditelev: @llewelld  Done

apps/browser/qml/pages/components/AddHomePageDialog.qml
aroditelev: @llewelld  Done

apps/browser/qml/pages/components/AddHomePageDialog.qml
aroditelev: @llewelld  Done

apps/browser/qml/pages/components/AddHomePageDialog.qml
aroditelev: @llewelld  Done.  Added by "Spaces can't be included in the URL", but there was no corresponding inscription in the design. 
`Is it necessary to exclude spaces from the URL?` Сombobox will have an extra space at the beginning of the line, and desktop browsers remove spaces in the address automatically.

apps/browser/qml/pages/SettingsPage.qml
aroditelev: Most likely this is a problem in `BrowserListItem`

apps/browser/qml/pages/SettingsPage.qml
aroditelev: @llewelld Done

apps/browser/qml/pages/components/AddHomePageDialog.qml
aroditelev: @llewelld  `In this case, maybe the error highlight should show if the textField is empty, in order to clarify to the user why the dialogue can't be accepted. What do you think?` I added to the description: ` //% "Enter website address that should be shown when browser is opened with no tabs to load."
        description: !errorHighlight ? qsTrId("sailfish_browser-he-home_page_website-address")
                                       //% "Spaces can't be included in the URL."
                                     : qsTrId("sailfish_browser-he-home_page_website-spaces_error")`

apps/browser/qml/pages/SettingsPage.qml
aroditelev: @llewelld  added an overload alias: `property alias animationDuration: valueButton._duration` in` BrowserListItem`. Now "Home" icon not jumps.

llewelld: Thanks for taking a look at this. I'm still getting the icon jumping position (but in a slightly different way). It's more obvious if text size is set to Huge.

I agree with you that this is a problem in `BrowserListItem`, so if you prefer we can leave it as it is here and look into the jump separately.

apps/browser/qml/pages/components/AddHomePageDialog.qml
```diff
@@ -0,0 +1,57 @@
...
+
+    onAccepted: {
+        homePageConfig.value = textField.text || "about:blank"
+    }
+
+    TextField {
+        id: textField
+        anchors.top: header.bottom
+        text: homePageConfig.value === "about:blank" ? "" : homePageConfig.value
+        //% "Enter website address that should be shown when browser is opened with no tabs to load."
```
llewelld, line 38: Thank you for adding this. I'd like to suggest some small changes to simplify the wording:
```suggestion
        //% "Shown when the browser is opened with no tabs to load"
```

apps/browser/qml/pages/components/AddHomePageDialog.qml
llewelld: Could you use `trim()` to remove the extra spaces instead for the combo box instead (or maybe I'm misunderstanding)?

It's just that we allow spaces in the address bar, and in favourites, so it would seem reasonable to allow them here too (otherwise the user has to manually change each of them to a %20 instead).

llewelld (commented):
Thanks for making the changes.

apps/browser/qml/pages/SettingsPage.qml
aroditelev: @llewelld it seems to me better to abandon the `BrowserListItem` in favor of the ComboBox. The problem is being solved.

apps/browser/qml/pages/components/AddHomePageDialog.qml
aroditelev: @llewelld  regexp finds all spaces, if, for example, the user enters https: // goo gle.com, then when the browser starts, not the page `https: // google.com` will open, but a search query, in the field of which there will be` https: // goo gle .com`. It checks and warns the user that this should not be done.

llewelld (approved):
Thanks, this works nicely and looks good to me.

atatarov (approved):

PR #813, opened 2021-03-19
Animate overlay when closing last tab.

Tomin1:
Animate overlay when last tab is closed from toolbar.

### DIFF
--- a/apps/browser/qml/pages/components/Overlay.qml
+++ b/apps/browser/qml/pages/components/Overlay.qml
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014 - 2019 Jolla Ltd.
+ * Copyright (c) 2014 - 2021 Jolla Ltd.
  * Copyright (c) 2019 - 2021 Open Mobile Platform LLC.
  *
  * This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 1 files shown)


apps/browser/qml/pages/components/Overlay.qml
```diff
@@ -76,11 +76,12 @@ Shared.Background {
...
     }
 
-    function startPage() {
+    function startPage(action) {
         searchField.enteringNewTabUrl = true
         _showUrlEntry = true
+        _overlayHeight = Qt.binding(function () { return overlayAnimator._fullHeight })
         searchField.resetUrl("")
-        overlay.animator.updateState("startPage", PageStackAction.Immediate)
+        overlayAnimator.showStartPage(action !== PageStackAction.Animated)
```
Tomin1, line 84: I wasn't quite sure if I should really do this the other way around (`action === PageStackAction.Immediate`) like in that `enterNewTabUrl()` above and then change all calls to `startPage()`. Anyway, feel free to comment.

rainemak: Looks like what I was expecting.

apps/browser/qml/pages/components/Overlay.qml
```diff
@@ -496,7 +497,7 @@ Shared.Background {
                     // Activates (loads) the tab next to the currect active.
                     webView.tabModel.closeActiveTab()
                     if (webView.tabModel.count === 0) {
-                        overlay.startPage()
+                        overlay.startPage(PageStackAction.Animated)
```
rainemak, line 500: :+1:

rainemak (approved):

PR #816, opened 2021-03-24
[browser] Allow transfers to cancel and restart downloads. Fixes JB#53658

rainemak:


apps/browser/browserservice.cpp
```diff
@@ -118,6 +119,22 @@ bool BrowserService::isPrivileged() const
...
+    uint callerServicePid = GET_PID().value();
+
+    // Test this against pid of 'org.nemo.transferengine' which works also inside
+    // sandbox. If that matches, then the caller is transferengine.
+    if (callerServicePid == connection().interface()->servicePid(TransferEngine).value()
+            || isPrivileged()) {
+        return true;
+    }
+
+    sendErrorReply(QDBusError::AccessDenied,
```
Tomin1, line 133: isPrivileged() call above already sends error reply so this may try to send it twice in some cases.

apps/browser/browserservice.cpp
```diff
@@ -118,6 +119,22 @@ bool BrowserService::isPrivileged() const
...
+                   QString("PID %1 is not the owner of '%2' or in privileged group").arg(callerServicePid).arg(TransferEngine));
```
Tomin1, line 134: Why QString and not QStringLiteral?

rainemak: No reason... will fix.

apps/browser/browserservice.cpp
rainemak: I was pondering if checking that caller is transferengine would be enough. That's not 100% solid of course in case somebody/something fakes itself. However, cancelling and restarting download is not that harmful. How do you see it?

rainemak: Like this.

Tomin1: Well, I think it's probably better to allow all privileged for development purposes. isPrivileged() in BrowserServiceUI works a little bit differently as it doesn't send the error reply but leaves that to the caller. Also it's possible to use that IS_PRIVILEGED() macro directly in isCalledFromTransfers() if that seems like a better fit although it will result in having yet another copy of that same code in the binary.

rainemak: Updated...

Tomin1 (approved):
LGTM

PR #817, opened 2021-03-24
[browser] Add tabs for mode private. Contributes to JB#50860

xlise:


apps/browser/qml/pages/components/TabView.qml
```diff
@@ -40,131 +32,144 @@ SilicaGridView {
...
-                    //% "Normal browsing"
-                    qsTrId("sailfish_browser-me-normal_browsing") :
-                    //: Menu item switching to private browser
-                    //% "Private browsing"
-                    qsTrId("sailfish_browser-me-private_browsing")
-            onDelayedClick: {
-                if (remorsePopup) {
-                    remorsePopup.trigger()
-                }
+    PrivateModeTexture {
```
llewelld, line 61: Currently the private mode background texture appears/disappears across the whole page as the tab is changed. I'm wondering if maybe it would look better if it slid in along with the page?

apps/browser/qml/pages/components/TabView.qml
```diff
@@ -40,131 +32,144 @@ SilicaGridView {
...
 
+    Component {
+        id: persistentTabView
+
+        Private.TabItem {
+            TabGridView {
+                id: _persistentTabView
+                privateMode: false
+                portrait: tabs.portrait
+                model: !tabs.privateMode ? tabs.tabModel : null
```
llewelld, line 77: Using the same `tabModel` and switching between them like this seems to be causing problems for the animation when swiping between the normal and private tab listings. The items appear suddenly half way through the swipe.

To get this to work nicely, I think it'll be necessary to maintain both models simultaneously and separately here.

apps/browser/qml/pages/components/TabView.qml
```diff
@@ -11,26 +11,18 @@
...
+import Sailfish.Silica.private 1.0 as Private
+import Sailfish.Browser 1.0
 import org.nemomobile.configuration 1.0
 import "." as Browser
+import "../../shared" as Shared
 
-SilicaGridView {
-    id: tabView
+Private.TabView {
+    id: tabs
```
llewelld, line 21: There's an error from Overlay.qml attempting to access `Browser.TabView.positionViewAtIndex()` It doesn't look like it's being caused by your changes here, but do you think it would be possible to fix it along with these changes?
```
[W] unknown:656 - file:///usr/share/sailfish-browser/pages/components/Overlay.qml:656: ReferenceError: positionViewAtIndex is not defined
[W] unknown:665 - file:///usr/share/sailfish-browser/pages/components/Overlay.qml:665: TypeError: Property 'positionViewAtIndex' of object TabView_QMLTYPE_144
```

apps/browser/qml/pages/components/TabGridView.qml
```diff
@@ -0,0 +1,141 @@
...
+        }
+    ]
+
+    ConfigurationValue {
+        id: showCloseAllAction
+        key: "/apps/sailfish-browser/settings/show_close_all"
+        defaultValue: true
+    }
+
+    PullDownMenu {
```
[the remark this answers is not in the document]
xlise: Return to the previous page blocked, the menu is still left

apps/browser/qml/pages/components/TabGridView.qml
```diff
@@ -0,0 +1,141 @@
...
+    signal closeAll
+    signal closeAllCanceled
+    signal closeAllPending
+
+    onCountChanged: if (count > 0) closingAllTabs = false
+    onClosingAllTabsChanged: if (closingAllTabs) closeAllPending()
+
+    width: parent.width - Theme.horizontalPageMargin
+    height: parent.height
+    x: Theme.horizontalPageMargin
```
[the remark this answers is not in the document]
xlise: It turned out to be problematic to fix, besides, this problem will go away when switching to the popup menu

apps/browser/qml/pages/components/TabView.qml
```diff
@@ -40,131 +32,144 @@ SilicaGridView {
...
+        // TODO: Add remorsePopup
+        // if (remorsePopup) {
+        //     remorsePopup.trigger()
+        // }
+        privateMode = currentIndex !== 0
+    }
+    Component.onCompleted: {
+        persistentIcon.updateGrubImage()
+        privateIcon.updateGrubImage()
+        currentIndex = privateMode
```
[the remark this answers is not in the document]
xlise: done

apps/browser/qml/pages/components/TabView.qml
xlise: I agree, corrected

apps/browser/qml/pages/components/TabView.qml
xlise: I agree, corrected

apps/browser/qml/pages/components/TabView.qml
xlise: Applied the hack, as there are no ideas how to do it correctly 
```
        Component.onCompleted:  {
            if ((index / columns * cellHeight) > tabView.height) {
                tabView.loaded = true
            }
        }
```

apps/browser/qml/pages/components/TabGridView.qml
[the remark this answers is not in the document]
adenexter: The gesture conflict could be reduced by disabling wrapping in the tab view. Then if you're on the first tab the gesture would return you to the previous page.

`wrapMode: PagedView.NoWrap` would do I think.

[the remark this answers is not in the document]
xlise: Good idea, I will consult with the designer

xlise:
Снимок_Экрана_20210520_001

apps/browser/qml/pages/components/TabGridView.qml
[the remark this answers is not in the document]
xlise: Fixed
`interactive: false` It turned out like this

apps/browser/qml/pages/components/TabGridView.qml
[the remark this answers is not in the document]
adenexter: The convention is to put the menu above the header. The tab header background rectangle is going to create a bit of a problem there because it would need to be below the pull down menu so it doesn't occlude the highlight.

But for reference the way to position the menu at the top of the window would be to assign the tabs grid view to the TabItem flickable property and add a header item to the grid view that is the same height as TabView header.

```
Private.TabItem {
    flickable: _persistentTabView
    TabGridView {
         id: _persistentTabView
         header: Item {
              width: 1
              height: tabs.headerHeight
         }
   }
}
```

The horizontal offset might still be a problem even then though. Provided there's no clipping though, you might be able to fix it by adding a negative x to the PullDownMenu as a counter margin and increasing the width of the menu to resolve the margin on the right.

Alternatively including the margin in the GridView delegates instead of the view itself would work too.

apps/browser/qml/pages/components/TabView.qml
```diff
@@ -40,131 +34,204 @@ SilicaGridView {
...
+                height: _privateIcon.implicitHeight
+                width: _privateIcon.implicitWidth
+
+                Image {
+                    id: _privateIcon
+
+                    source: webView.privateTabModel.count > 0 ? "image://theme/icon-m-incognito-selected" : "image://theme/icon-m-incognito"
+                    visible: false
+                }
+                Rectangle {
```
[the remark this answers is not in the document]
xlise: Yes you are right. done

apps/browser/qml/pages/components/TabGridView.qml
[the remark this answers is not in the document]
xlise: Thanks for your suggestions for improvement. Fixed gap. Moved the pulley menu above the title, but this can only be considered as a temporary solution before the implementation of the popup menu

adenexter:
It's looking pretty good to me now. I am seeing a single pixel uneven line just below the pulley menu indicator. I think this is the normal tab header background item not being quite hidden by the black rectangle background and on just that one line you can see through to the web page below the page.

Hiding the tab header background with `_headerBackgroundVisible: false` in the silica TabView fixed that for me.

The scroll bar also gets hidden behind the tab header when the view is scrolled to the top but I don't have a solution for that, it is within a tab while needing to appear above a sibling item to the tab in the header.

apps/browser/qml/pages/components/TabGridView.qml
```diff
@@ -0,0 +1,153 @@
...
+    signal closeAll
+    signal closeAllCanceled
+    signal closeAllPending
+
+    onCountChanged: if (count > 0) closingAllTabs = false
+    onClosingAllTabsChanged: if (closingAllTabs) closeAllPending()
+
+    width: parent.width - Theme.horizontalPageMargin
+    height: parent.height
+    y: Theme.paddingLarge
```
llewelld, line 49: Shifting the `SilicaGridView` down by `Theme.paddingLarge` here is unfortunately causing the pulley menu to be shifted down as well, so there's a gap at the top when you pull the items down.

I recommend removing this `y` shift here, and instead adding `Theme.paddingLarge` to both `TabGridView.header` items in `TabView.qml`.
```suggestion
```

apps/browser/qml/pages/components/TabView.qml
```diff
@@ -40,131 +34,230 @@ SilicaGridView {
...
 
-                tabView.privateMode = !tabView.privateMode
+                TabGridView {
+                    id: _privateTabView
+                    privateMode: true
+                    portrait: root.portrait
+                    model: webView.privateTabModel
+                    header: Item {
+                        width: 1
+                        height: tabs.tabBarHeight
```
llewelld, line 131: Similarly, the same extra space would need to be added here:
```suggestion
                        height: tabs.tabBarHeight + Theme.paddingLarge
```

apps/browser/qml/pages/components/TabView.qml
```diff
@@ -40,131 +34,230 @@ SilicaGridView {
...
+                flickable: _persistentTabView
+
+                TabGridView {
+                    id: _persistentTabView
+                    privateMode: false
+                    portrait: root.portrait
+                    model: webView.persistentTabModel
+                    header: Item {
+                        width: 1
+                        height: tabs.tabBarHeight
```
llewelld, line 98: In order to compensate for removing the y offset from `TabGridView.qml` I suggest adding the extra space here instead.
```suggestion
                        height: tabs.tabBarHeight + Theme.paddingLarge
```

llewelld (changes requested):
This is so much nicer, thank you. You even managed to solve the "swipe back" problem, which is fantastic. I've made just a few small comments, mostly about fixing some spacing issues. There also seem to be a couple of comments from Andrew still unresolved if I'm reading correctly.

xlise:
> This is so much nicer, thank you. You even managed to solve the "swipe back" problem, which is fantastic. I've made just a few small comments, mostly about fixing some spacing issues. There also seem to be a couple of comments from Andrew still unresolved if I'm reading correctly.

Notes from Andrew will still edit

apps/browser/qml/pages/components/TabView.qml
```diff
@@ -40,131 +34,230 @@ SilicaGridView {
...
+            if (_remorsePopup) {
+                _remorsePopup.trigger()
+            }
+            privateMode = currentIndex !== 0
+        }
+        Component.onCompleted: {
+            persistentIcon.updateGrubImage()
+            privateIcon.updateGrubImage()
+            currentIndex = privateMode ? 1 : 0
         }
```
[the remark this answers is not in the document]
xlise: Thank you, done

apps/browser/qml/pages/components/TabGridView.qml
```diff
@@ -0,0 +1,153 @@
...
+    x: Theme.horizontalPageMargin
+    currentIndex: -1
+    header: spacer
```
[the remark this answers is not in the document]
xlise: Thank you, it's really neat. done

apps/browser/qml/pages/components/TabGridView.qml
```diff
@@ -0,0 +1,153 @@
...
+        //: Hint to create a new tab from pull down menu.
+        //% "Pull down to create a new tab"
+        text: qsTrId("sailfish_browser-la-pull_down_to_create_tab_hint")
+    }
+
+    Component {
+        id: spacer
+        Item {
+            width: tabView.width
+            height: Theme.paddingMedium
```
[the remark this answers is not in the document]
xlise: done

apps/browser/qml/pages/components/TabGridView.qml
xlise: done

apps/browser/qml/pages/components/TabView.qml
xlise: done

apps/browser/qml/pages/components/TabView.qml
xlise: done

llewelld:
> The scroll bar also gets hidden behind the tab header when the view is scrolled to the top but I don't have a solution for that, it is within a tab while needing to appear above a sibling item to the tab in the header.

Maybe this is really nasty, but placing the `VerticalScrollBarDecorator` inside an `Item` that's only as high as the area beneath the toolbar could work. Something like this (this would need tweaking though because `tabsToolBar.height` isn't currently available inside the `TabViewGrid`s).

```
Item {
    y: tabsToolBar.height
    height: parent.height - y
    anchors.right: parent.right
    width: Math.round(Theme.paddingSmall/2)

    VerticalScrollDecorator {
        _forcedParent: parent
        flickable: tabView
    }
}
```

apps/browser/qml/pages/components/TabView.qml
```diff
@@ -40,131 +34,230 @@ SilicaGridView {
...
-        GridView.onRemove: RemoveAnimation {
-            target: tabItem
+
+        header: Private.TabBar {
+            model: headerModel
+            Rectangle {
+                anchors {
+                    fill: parent
+                    topMargin: _pulleyMenuApparentHeight
```
llewelld, line 61: I really like the way you've handled hiding the top menu highlight; it's very ingenious. However it seems to cause some issues, because `_pulleyMenuApparentHeight` doesn't update when the normal/private tab changes. I tried calling `_changeMenuApparentHeight()` when the `currentIndex` changed, but that resulted in other animation artefacts.

Perhaps as an alternative consider adding in a covering bar to the two `Private.TabItem` items instead. This way the bar will slide left/right when the grids slide. Something like this (just a rough idea):
```
Rectangle {
    color: Theme.colorScheme == Theme.LightOnDark ? "black" : "white"
    width: parent.width
    height: Math.max(0, _pulleyMenuHeight - y)
    y: Math.max(0, _pulleyMenuHeight - _persistentTabView.contentY + _persistentTabView.originY)
}   
```
It makes the code a bit messy, but then this is only a temporary solution until the menu is changed anyway.

xlise: Thank you, so much better. done

llewelld: Thanks @xlise. Unfortunately I don't see the change in practice (did you push it to the repo? I see a force-push, but I don't see any changes; it looks like a rebase).

xlise: Sorry, I haven't added any files. done

llewelld: Thank you! You should be safe to remove the `_changeMenuApparentHeight()` logic now, and replace the header topMargin to be a fixed height. I'll add a comment about it above.

apps/browser/qml/pages/components/TabView.qml
```diff
@@ -40,131 +34,248 @@ SilicaGridView {
...
-            target: tabItem
+
+        header: Private.TabBar {
+            id: headerTabs
+            model: headerModel
+
+            Rectangle {
+                anchors {
+                    fill: parent
+                    topMargin: _pulleyMenuApparentHeight
```
llewelld, line 64: This should now be safe:
```suggestion
                    topMargin: _pulleyMenuHeight
```
Which would allow you to remove the `_changeMenuApparentHeight()` logic I think.

xlise: I agree, removed

llewelld (approved):
Thanks for your work on this, it looks good to me, and is working really great in my opinion.

Please rebase before we merge it.

xlise:
> Thanks for your work on this, it looks good to me, and is working really great in my opinion.
> 
> Please rebase before we merge it.

Yes, it works, but I still haven't corrected one comment from Andrew, but it doesn't work there as Andrew suggested

llewelld:
Thanks, and understood. I won't merge anything until you've completed your discussion with Andrew and you say it's ready (it still needs approval from someone at OMP too before we can merge it).

apps/browser/qml/pages/components/TabView.qml
```diff
@@ -40,131 +34,204 @@ SilicaGridView {
...
             }
         }
-        MenuItem {
-            //% "New tab"
-            text: qsTrId("sailfish_browser-me-new_tab")
-            onClicked: tabView.enterNewTabUrl()
+
+        function _updateHeaderModel() {
+            headerModel.set(0, { "icon": persistentIcon.grabIcon })
+            headerModel.set(1, { "icon": privateIcon.grabIcon })
```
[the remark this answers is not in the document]
xlise: Thank you, this option is really better. done

apps/browser/qml/pages/components/TabView.qml
```diff
@@ -40,131 +34,204 @@ SilicaGridView {
...
+                    text: webView.persistentTabModel.count
+                    font.pixelSize: Theme.fontSizeExtraSmall
+                    font.bold: true
+                }
+            },
+            Item {
+                id: privateIcon
+                property string grabIcon
+
+                function updateGrubImage() {
```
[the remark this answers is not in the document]
xlise: done

adenexter (commented):
> Yes, it works, but I still haven't corrected one comment from Andrew, but it doesn't work there as Andrew suggested. Rebased

Which comment was this? The only thing I can see that might be unresolved is that you could use a delegate and model for the Silica TabView the same as with the header so as to not have two mostly the same TabItem components.

apps/browser/qml/pages/components/TabView.qml
```diff
@@ -40,131 +34,221 @@ SilicaGridView {
...
+                    fill: parent
+                    topMargin: _pulleyMenuHeight
                 }
 
-                tabView.privateMode = !tabView.privateMode
-
+                z: -100
+                color: Theme.colorScheme == Theme.LightOnDark ? "black" : "white"
             }
+            Component.onCompleted: tabs.headerHeight = Qt.binding(function() { return height })
```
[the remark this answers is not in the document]
adenexter: Correction, Silica TabView has a tabBarHeight property so you can use that instead of defining your own headerHeight property.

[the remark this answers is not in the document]
xlise: Thanks!  done

xlise:
> > Yes, it works, but I still haven't corrected one comment from Andrew, but it doesn't work there as Andrew suggested. Rebased
> 
> Which comment was this? The only thing I can see that might be unresolved is that you could use a delegate and model for the Silica TabView the same as with the header so as to not have two mostly the same TabItem components.

Yes, I mean this remark, I corrected the TabBar. delegate for components today I will try to make

adenexter:
> Yes, I mean this remark, I corrected the TabBar. delegate for components today I will try to make

Okay. Something to watch for is TabGridView has both model and privateMode properties, so if you try and reference the model data from within the TabGridView declaration both model.privateMode and privateMode will resolve to its properties and not the model data.

The best way to address that is probably bind model.privateMode to a property in the delegate top level TabItem and reference that.

xlise:
> > Yes, I mean this remark, I corrected the TabBar. delegate for components today I will try to make
> 
> Okay. Something to watch for is TabGridView has both model and privateMode properties, so if you try and reference the model data from within the TabGridView declaration both model.privateMode and privateMode will resolve to its properties and not the model data.
> 
> The best way to address that is probably bind model.privateMode to a property in the delegate top level TabItem and reference that.

Great thanks a lot for your hint. The code is much better. done

adenexter (approved):

atatarov (approved):

aroditelev:
LGTM

PR #820, opened 2021-04-01
[browser] Fix tabview margins. Fixes jb#53673

mkenttala:


mkenttala:
Got notified there's already PR that is likely to fix this one too: https://github.com/sailfishos/sailfish-browser/pull/817

apps/browser/qml/pages/components/TabView.qml
```diff
@@ -43,7 +43,7 @@ SilicaGridView {
     onCountChanged: if (count > 0) closingAllTabs = false
     onClosingAllTabsChanged: if (closingAllTabs) closeAllPending()
 
-    width: parent.width - Theme.horizontalPageMargin
+    width: parent.width - 2 * Theme.horizontalPageMargin
```
rainemak, line 46: Other way around. Both SilicaGridView and PullDownMenu should expand to full width and only the actual content should have y:  Theme.horizontalPageMargin and width: parent.width - 2 * Theme.horizontalPageMargin

rainemak: That implying also that background texture used in private browser mode should expand to full width.

xlise: Please don't need to fix it, I'll fix it in https://github.com/sailfishos/sailfish-browser/pull/817

mkenttala:
Closing, will get fixed with #817.

PR #821, opened 2021-04-05
[sailfish-browser] Remove handling downloadpicker from WebEngine. Contributes to JB#53798

atatarov:
It is no longer required to do this here, its already in the PickerOpener.


llewelld (approved):
This looks good to me. I'll merge all three PRs at the same time once the other two are also approved.

okodron (approved):

PR #823, opened 2021-04-12
Small cleanups for the repository

rainemak:


rainemak:
@okodron @atatarov maybe you'd like to give a look as well?

Will cleanup some contacts names there later today evening.

apps/storage/dbmanager.cpp
```diff
@@ -53,7 +53,7 @@ DBManager::~DBManager()
     // Use timeout of 500ms to guaranty we won't block
     workerThread.wait(500);
     gDbManager = 0;
-    foreach (QString connectionName, QSqlDatabase::connectionNames()) {
+    for (const QString &connectionName : QSqlDatabase::connectionNames()) {
```
llewelld, line 56: I've been trying to learn about this from @chriadam ([useful ref](https://youtu.be/Ov7s0GgBbOQ?t=667)), so I hope you'll bear with me on this, it's a bit of a learning exercise for me. As I understand it using C++ `for` here (as opposed to Qt's `foreach`) on a Qt container could result in a deep copy, because the value returned by `QSqlDatabase::connectionNames()` is not `const` (even though the value pulled from it is). So it would be better to either leave this using the Qt macro, or alternatively cast the returned `QStringList` to `const` before the `for` loop.

I think this is the only case where this could happen (there are some other potentials, but elsewhere it looks like the lists are never handed out, so this wouldn't apply).

As I say, I'm still learning about this, so happy to be corrected.

okodron: Yes, if you want to be 100% sure it's better to `const` cast the returned `QStringList` object as otherwise it can detach during the for-based-loop ([1], [2] & [3]):
```suggestion
    const auto names = QSqlDatabase::connectionNames();
    for (const QString &connectionName : names) {
```

[1] https://stackoverflow.com/questions/35811053/using-c11-range-based-for-loop-correctly-in-qt
[2] https://www.dvratil.cz/2015/06/qt-containers-and-c11-range-based-loops/
[3] https://www.kdab.com/goodbye-q_foreach/

apps/core/settingmanager.cpp
```diff
@@ -196,7 +195,8 @@ void SettingManager::handleObserve(const QString &message, const QVariant &data)
             }
 
             // Remove uninstalled OpenSearch configs
-            foreach(QString searchName, registeredSearches) {
+            const QStringList removeUninstalled = registeredSearches;
```
rainemak, line 198: const stringlist as well.

common/opensearchconfigs.cpp
```diff
@@ -26,11 +26,12 @@ const StringMap OpenSearchConfigs::parseOpenSearchConfigs()
 {
     StringMap configs;
 
-    foreach(QString openSearchPath, m_openSearchPathList) {
+    for (const QString &openSearchPath : m_openSearchPathList) {
         QDir configDir(openSearchPath);
         configDir.setSorting(QDir::Name);
 
-        foreach (QString fileName, configDir.entryList(QStringList("*.xml"))) {
+        const QStringList configFiles = configDir.entryList(QStringList("*.xml"));
```
rainemak, line 33: and same here

rainemak:
Rebased and fixed findings.

okodron (approved):

llewelld (approved):
LGTM

PR #824, opened 2021-04-12
Remove clearance of general.useragent.override user pref

llewelld:
Clearing the general.useragent.override user pref was a temporary
addition intended to be removed on the next upgrade.

Reverts 426e60bed917d46e2191f6f1b15ebe22dacfd681


rainemak (approved):

llewelld:
Rebased (badly, but got there eventually).

PR #826, opened 2021-04-16
[browser] Implement microphone permission settings. Fixes JB#53948

d-grigorev:


apps/browser/qml/pages/PermissionPage.qml
```diff
@@ -60,24 +60,37 @@ Page {
                                         title: qsTrId("sailfish_browser-ti-geolocation"),
                                         type: "geolocation",
                                         capability: PermissionManager.Prompt,
-                                        iconSource: "image://theme/icon-m-browser-location"
+                                        iconSource: "image://theme/icon-m-browser-location",
+                                        isSensitiveData: true
```
atatarov, line 64:
```suggestion
                                        sensitiveData: true
```
It is preferable to not to use the suffix in qml for boolean

apps/browser/qml/pages/SitePermissionPage.qml
```diff
@@ -43,6 +43,7 @@ Page {
         "geolocation": PermissionManager.Prompt,
         "popup": _getPopupCapability(),
         "cookie": _getCookieCapability()
```
atatarov, line 45:
```suggestion
        "cookie": _getCookieCapability(),
```

apps/browser/qml/pages/PermissionPage.qml
d-grigorev: Edited. Thanks.

apps/browser/qml/pages/SitePermissionPage.qml
d-grigorev: Fixed, thank you.

atatarov (approved):
It looks good to me. So, lets wait for the others PR's
https://git.sailfishos.org/mer-core/gecko-dev/merge_requests/247
https://git.sailfishos.org/mer-core/embedlite-components/merge_requests/120
https://github.com/sailfishos/sailfish-components-webview/pull/98

okodron (approved):

llewelld (approved):
LGTM.

llewelld (commented):
Should this be on top of the `next` branch? If so, please rebase it there.

d-grigorev:
> Should this be on top of the `next` branch? If so, please rebase it there.

Done.

llewelld (changes requested):
This still works great. The only thing I've subsequently noticed is that the permissions also has to be handled in [CertificateInfo.qml](https://github.com/sailfishos/sailfish-browser/blob/0aac4d7ad343b4c90b4dd4f59f9b72493b4dfd10/apps/browser/qml/pages/components/CertificateInfo.qml#L174) to avoid breaking list of current permissions for the page.

I think it would be sensible to make the change at the same time as the other changes.

d-grigorev:
> This still works great. The only thing I've subsequently noticed is that the permissions also has to be handled in [CertificateInfo.qml](https://github.com/sailfishos/sailfish-browser/blob/0aac4d7ad343b4c90b4dd4f59f9b72493b4dfd10/apps/browser/qml/pages/components/CertificateInfo.qml#L174) to avoid breaking list of current permissions for the page.
> 
> I think it would be sensible to make the change at the same time as the other changes.

Didn't know about that nice page. Fixed, thanks.

llewelld:
> Didn't know about that nice page. Fixed, thanks.

Yeah, I only noticed by accident myself. Thanks for making the change.

apps/browser/qml/pages/components/CertificateInfo.qml
```diff
@@ -179,6 +179,8 @@ SilicaFlickable {
                                         return "image://theme/icon-m-browser-cookies" + (blocked ? "-template" : "")
                                     } else if (model.type === "popup") {
                                         return "image://theme/icon-m-browser-popup" + (blocked ? "-template" : "")
+                                    } else if (model.type === "microphone") {
+                                        return "image://theme/icon-m-browser-microphone" + (blocked ? "-template" : "")
```
llewelld, line 183: It looks like you actually need to handle the AlwaysAsk as a special case here, If it's set to AlwaysAsk, I'd suggest not showing the icon at all.

tests/manual/testwebrtc.html
```diff
@@ -0,0 +1,81 @@
...
+      .then(
+        (stream) => { event.target.className = "button success"; playStream(stream); },
+        (_)      => { event.target.className = "button error"; }
+      )
+      .catch(stopPlayback)
+    }
+  );
+}
+</script>
+</body>
```
llewelld, line 81: Really nice test page, by the way.

apps/browser/qml/pages/components/CertificateInfo.qml
d-grigorev: Hmm. This is strange. From my opinion selecting "Always ask" on this page should  remove the exception, but it's not. I will look at this, thanks.

d-grigorev: Btw, the same is happening for geolocation, so it needs some rework. Also, the PermissionExceptionsPage.qml gets weird, as it cannot show "Always ask".  I guess it is better to show (and do) "Delete exception" instead of "Always ask", what do you think?

llewelld: Yes, I'd agree. Either have "Always ask" as the text but actually delete the exception, or change the wording to "Delete exception" as you suggest as well.

Just to ensure we've not missed anything: if there were the option to set "Always allow" or "Always deny" as global permissions, then having an "Always ask" exception would make sense. However, it looks to me like it's not possible to set "Always allow/deny" as global options, so just deleting the exception is a suitable approach. Does that fit with your understanding too?

Since this isn't an issue that's exclusive to the microphone, you could record this as a separate issue.

atatarov: There are two case:
1. The expire session
2. The permanent permission
When we set the global permisson "block all request", the expire session will ignore, but the permanent permission will be as a exception. So, we should show both, when you tapping on cert info and change permission for this site, the permanent state will set(anyway).
If you want to filter "session permissions" in the case when we allow/deny all request, you should change filterproxymodel and it should depends on WebEngineSettings, it would be a difficult.

atatarov: I think we already have the solution:
1. The user gets request for permission, selects allow/block(expire session). 
2. The user go to global settings and tap on "Block new requests" 
I think the "new requests" is a key info for our behavior. The new requests will be blocked, the expire session already confirmed before

d-grigorev: @atatarov, Here's an another case, not related to this particular PR:
1. Allow some site to  use the mic, check "Remember for this site"
2. Go to certificate info page
3. Observe microphone icon
4. Tap on microphone icon, the page with exceptions for this site will open
5. Select "Always ask" for the microphone
6. Go to the Settings -> Permissions, tap "Microphone" then "Show exceptions". The exceptions page will open. Observe a site URI with an empty decision on this: nor Allow nor Block is shown, just an URI.
7. Go to certificate info.
8. Microphone icon is still there

This is due to "Always ask" stored as permanent exception. From my understanding, the permanent exception should be removed on step 5. The location permission has the same behavior.

I'm not sure this should be solved within this task, I will file an another bug for this.

Regarding session permissions. I'm deliberately not setting session permissions, because I not sure that it will work as user expects. I think the browser must always ask for user's decision when "Always ask" policy is set.

d-grigorev: Or. We may legalize "Always ask" as a permanent exception. For exemple, user sets "Always block" as default policy, then selects a site and make and exception "Always ask" for it. What do you think? It is much easier to implement and have a real use case. What do you think?

atatarov: So, let move this question to separate task, where will be the global permissions

d-grigorev: See https://bz.jollamobile.com/show_bug.cgi?id=54610

llewelld (approved):
With the permission issue report as a separate bug, this now LTGM. Thanks Denis.

apps/browser/qml/pages/components/CertificateInfo.qml
llewelld: Okay, this is in the new bug now, so I'll resolve this conversation. Thanks both.

PR #830, opened 2021-04-23
Ensure BookmarkFilterModel::filterAcceptsRow() can be called out-of-order

llewelld:
To be added on top of https://github.com/sailfishos/sailfish-browser/pull/798, so will be WIP until that has been merged.

A small glitch meant that favicons downloaded when a favourite was added
could trigger a `dataChanged()` signal while the model was being
constructed, which interacted badly with the accept count.

This change ensures that `filterAcceptRow()` calls can be received in any
order and still respond correctly.

### DIFF
--- a/apps/browser/bookmarks/bookmarkfiltermodel.cpp
+++ b/apps/browser/bookmarks/bookmarkfiltermodel.cpp
@@ -26,11 +26,16 @@ int BookmarkFilterModel::getIndex(int currentIndex)
 
 bool BookmarkFilterModel::filterAcceptsRow(int sourceRow, const QModelIndex &sourceParent) const
 {
+    extendCheckPosition(sourceRow, sourceParent);
+    return ((sourceRow < m_maxSourceModelPosition) && filterAcceptsRowUnbounded(sourceRow, sourceParent));
+}
+
+bool BookmarkFilterModel::filterAcceptsRowUnbounded(int sourceRow, const QModelIndex &sourceParent) const
+{
     QModelIndex index = sourceModel()->index(sourceRow, 0, sourceParent);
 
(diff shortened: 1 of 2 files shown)


llewelld:
https://github.com/sailfishos/sailfish-browser/pull/798 has now been merged in and I've rebased this on top of it, so I'm removing the WIP prefix.

llewelld:
Rebased.

apps/browser/bookmarks/bookmarkfiltermodel.cpp
```diff
@@ -25,12 +25,17 @@ int BookmarkFilterModel::getIndex(int currentIndex)
 }
 
 bool BookmarkFilterModel::filterAcceptsRow(int sourceRow, const QModelIndex &sourceParent) const
+{
+    extendCheckPosition(sourceRow, sourceParent);
+    return ((sourceRow < m_maxPosition) && filterAcceptsRowUnbounded(sourceRow, sourceParent));
```
rainemak, line 30: Slightly counter intuitive naming as you're comparing positions as per variable names and acceptable position should be less than maxCount or <= maxPosition.

rainemak: Maybe it this should be called ```m_maxSourceModelPosition```.... Feel free to figure a better name as well.

rainemak (approved):
Seems to be working. Might be worth adjusting that one variable name or adding a comment there. Added a name proposal.

Your call to update the member variable name as it's quite tricky name.

Approving already.

okodron (approved):

apps/browser/bookmarks/bookmarkfiltermodel.cpp
llewelld: Thanks for your suggestion of `m_maxSourceModelPosition`; it makes good sense to me as well and I've switched to using that.

PR #831, opened 2021-04-26
[browser] Adapt to sharing UI and API changes. JB#53987 OMP#JOLLA-75

blammit:
The sharing UI is now launched out-of-process.

### DIFF
--- a/apps/browser/qml/pages/components/Overlay.qml
+++ b/apps/browser/qml/pages/components/Overlay.qml
@@ -14,6 +14,7 @@ import Sailfish.Silica.private 1.0 as Private
 import Sailfish.Browser 1.0
 import Sailfish.Policy 1.0
 import Sailfish.WebView.Controls 1.0
+import Sailfish.WebView.Popups 1.0
 import Sailfish.WebEngine 1.0
 import com.jolla.settings.system 1.0
 import "." as Browser
@@ -284,9 +285,8 @@ Shared.Background {
                         onCall: {
                             Qt.openUrlExternally("tel:" + controller.text)
                         }
-
                         onShare: {
-                            pageStack.animatorPush("Sailfish.WebView.Popups.ShareTextPage", {"text" : controller.text })
+                            webShareAction.shareText(controller.text)
                         }
                         onSearch: {
                             // Open new tab with the search uri.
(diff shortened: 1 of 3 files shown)


blammit:
Note: associated API changes need to be merged before this can be merged.

apps/browser/qml/pages/components/Overlay.qml
```diff
@@ -273,7 +278,16 @@ Shared.Background {
...
                         onShare: {
-                            pageStack.animatorPush("Sailfish.WebView.Popups.ShareTextPage", {"text" : controller.text })
+                            var content = {
+                                "name": controller.text,
+                                "data": controller.text,
+                                "type": "text/plain",
+                                // also some non-standard fields for Twitter/Facebook status sharing:
+                                "status": controller.text,
+                            }
+                            shareAction.capabilitiesFilter = "text/plain"
```
atatarov, line 288: it looks like this at least these three lines can be as a function

atatarov (approved):

apps/browser/qml/pages/components/Overlay.qml
```diff
@@ -476,10 +493,12 @@ Shared.Background {
...
-                    pageStack.animatorPush("Sailfish.WebView.Popups.ShareLinkPage", {
-                                               "link": webView.url,
-                                               "linkTitle": webView.title
-                                           })
+                    var content = {
+                        "type": "text/x-url",
+                        "linkTitle": webView.title,
+                        "status": webView.url,
+                    }
+                    _shareContent(content, content.type)
```
chriadam, line 501: the old page had a header: "Share link" etc.  Should that (translated) UI context hint be provided as some content data, so that the sharing dialog can display it?  Or is that already handled somehow?

blammit: I checked this with Martin and indeed the header should be kept. I've restored this now, in this MR.

pvuorela:
Should target the 'next' branch.

pvuorela:
There's also BookmarkItem.qml and HistoryList.qml that appear to use ShareLinkPage.

pvuorela (approved):
LGTM

okodron (approved):

PR #833, opened 2021-04-29
Add login management models

llewelld:
Adds models to allow management of logins (usernames and passwords)
through the frontend.

PR #834, opened 2021-04-29
Add login management settings page

llewelld:
Adds a page accessible from the settings for managing stored logins
(usernames and passwords).

This is still WIP, but it may already be useful for testing the models.

### DIFF
--- a/apps/browser/qml/pages/EditLoginPage.qml
+++ b/apps/browser/qml/pages/EditLoginPage.qml
@@ -1,0 +1,76 @@
+/****************************************************************************
+**
+** Copyright (c) 2021 Open Mobile Platform LLC.
+**
+****************************************************************************/
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.0
+import Sailfish.Silica 1.0
+import Sailfish.Browser 1.0
+
+Dialog {
+    id: page
+    property int uid: -1
+    property LoginModel loginModel
+    property alias hostname: hostnameField.text
+    property alias username: usernameField.text
+    property alias password: passwordField.text
+
+    canAccept: loginModel.canModify(uid, username, password)
+    onAcceptBlocked: usernameField.errorHighlight = true
+    onAccepted: loginModel.modify(uid, username, password);
+
+    SilicaFlickable {
+        anchors.fill: parent
+        contentHeight: column.height
+
+        Column {
+            id: column
+            width: parent.width
+
+            DialogHeader {
+                //% "Edit login"
+                title: qsTrId("sailfish_browser-me-login_edit_login")
+            }
+
+            Label {
+                id: hostnameField
+                x: Theme.horizontalPageMargin
+                width: parent.width - 2 * Theme.horizontalPageMargin
+                height: Theme.itemSizeSmall
+                //% "Hostname"
+                text: qsTrId("sailfish_browser-me-login_edit_hostname")
+                color: Theme.highlightColor
+            }
+
+            TextField {
+                id: usernameField
+                //% "Username"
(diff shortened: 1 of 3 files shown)


llewelld:
Rebased.

llewelld:
This is no longer WIP. However, some points to note:
1. The icon used for the button on the Settings page isn't yet available, so the one used is a placeholder.
2. The favicons in the login list are also placeholders. There's a separate task to integrate them properly.

apps/browser/settings/declarativeloginmodel.h
```diff
@@ -0,0 +1,61 @@
...
+#include <QQmlParserStatus>
+
+#include "logininfo.h"
+
+class DeclarativeLoginModel : public QAbstractListModel, public QQmlParserStatus
+{
+    Q_OBJECT
+    Q_INTERFACES(QQmlParserStatus)
+
+    Q_PROPERTY(int count READ count NOTIFY countChanged)
```
atatarov, line 24:
```suggestion
    Q_PROPERTY(int count READ rowCount NOTIFY countChanged)
```
I'm not sure if it's ok for you, but you can use 'rowCount()' instead additional function

apps/browser/settings/declarativeloginmodel.cpp
```diff
@@ -0,0 +1,155 @@
...
+}
+
+void DeclarativeLoginModel::handleRecvObserve(const QString &message, const QVariant &data)
+{
+    if (message == ALL_LOGINS) {
+        setLoginList(qvariant_cast<QVariantList>(data));
+    }
+}
+
+void DeclarativeLoginModel::setLoginList(const QVariantList &data)
```
atatarov, line 101: I think if it(https://github.com/sailfishos/sailfish-components-webview/pull/87) approved this code could be used
Taked from https://github.com/sailfishos/sailfish-browser/blob/master/apps/history/declarativehistorymodel.cpp#L113

apps/browser/settings/declarativeloginmodel.cpp
```diff
@@ -0,0 +1,155 @@
...
+}
+
+void DeclarativeLoginModel::requestLogins()
+{
+    QVariantMap data;
+    data.insert(QStringLiteral("action"), "getall");
+    SailfishOS::WebEngine::instance()->notifyObservers(LOGINS_ACTION, QVariant(data));
+}
+
+void DeclarativeLoginModel::remove(int index)
```
atatarov, line 122: If the user open context menu two times (while 5 seconds) and call remove, the second index will invalid or to point on other element. It could be avoided by searching the LoginInfo in m_loginList.
So, I see the 'getIndex()' still needed in this model, but it shouldn't use for removing elemenets from list. What do you think?

apps/browser/settings/declarativeloginmodel.h
llewelld: Thanks, nice idea. I've made this change.

apps/browser/settings/declarativeloginmodel.cpp
llewelld: Yes, a good point, thanks for spotting this. Unfortunately the LoginInfo items aren't necessarily unique, so it's not possible to find the right one just by searching on those. I've added a `uid` role to fix this.

apps/browser/settings/declarativeloginmodel.cpp
llewelld: Thanks, that's not a bad idea. However, this case is slightly different to the history model, because while the history model is persistent, this login model is transient, which means that `DeclarativeLoginModel::setLoginList()` will only ever be called on an empty model. As such, although the changes found in `DeclarativeHistoryModel::updateModel()` wouldn't do any harm, they aren't really needed here.

If future changes mean this model becomes persistent, then I think we should make this change.

llewelld:
Rebased on top of changes made to #833.

llewelld:
Rebased.

mkenttala (approved):

atatarov (approved):

llewelld:
Rebased.

PR #836, opened 2021-05-04
Add favicons to the login management settings page

llewelld:
Stores favicons for logins and shows them on the login management page in the browser settings.

Builds on top of #834.

To be used alongisede https://github.com/sailfishos/sailfish-components-webview/pull/106 .

PR #838, opened 2021-05-06
Add device lock gate for password access

llewelld:
The Logins list in Settings allows access to the user's stored account passwords, either by copying the password or editing it. This change introduces a device lock gate, so that the user has to perform device authentication (e.g. PIN, fingerprint) before the passwords will be shown.

Once authenticated, the user won't be asked again until the Logins list page has been exited.

Builds on top of #836, so only the latest commit (a07b351e0f4bb853475766e82fe6bae15884212c) contains changes from that.

### DIFF
--- a/apps/browser/qml/pages/LoginsPage.qml
+++ b/apps/browser/qml/pages/LoginsPage.qml
@@ -33,6 +33,21 @@ Page {
         return hostname
     }
 
+    function copyUsername(username) {
+        Clipboard.text = username
+        //% "Username copied"
+        notification.text = qsTrId("sailfish_browser-me-login_copied_username")
+        notification.show()
+    }
+
+    // Should only be called with SecureAction to avoid leaking passwords
+    function copyPassword(password) {
+        Clipboard.text = password
+        //% "Password copied"
+        notification.text = qsTrId("sailfish_browser-me-login_copied_password")
+        notification.show()
+    }
+
     SilicaListView {
         id: view
 
@@ -42,8 +57,8 @@ Page {
         header: Column {
             width: parent.width
             PageHeader {
-                //% "Logins and passwords"
-                title: qsTrId("sailfish_browser-he-logins")
+                //% "Passwords"
+                title: qsTrId("sailfish_browser-he-passwords")
             }
             SearchField {
                 width: parent.width
@@ -109,33 +124,57 @@ Page {
                 }
             }
 
-            menu: ContextMenu {
-                MenuItem {
-                    //% "Copy password"
-                    text: qsTrId("sailfish_browser-me-login_copy_password")
-                    onClicked: {
-                        Clipboard.text = model.password
-                        notification.show()
+            menu: Component {
+                ContextMenu {
+                    id: contextMenu
+                    MenuItem {
+                        visible: !secureAction.available
+                        //% "Copy username"
+                        text: qsTrId("sailfish_browser-me-login_copy_username")
+                        onClicked: copyUsername(model.username);
                     }
-                }
-                MenuItem {
-                    //% "Edit"
-                    text: qsTrId("sailfish_browser-me-login_edit")
-                    onClicked: {
-                        var page = pageStack.animatorPush("EditLoginPage.qml", {
-                                                              loginModel: loginModel,
-                                                              uid: model.uid,
-                                                              hostname: model.hostname,
-                                                              username: model.username,
-                                                              password: model.password
-                                                          })
+                    MenuItem {
+                        visible: secureAction.available
+                        //% "Copy"
+                        text: qsTrId("sailfish_browser-me-login_copy")
+                        onClicked: {
+                            // Hide existing menu items
+                            var content = contextMenu._contentColumn
+                            for (var i = 0; i < content.children.length; i++) {
+                                content.children[i].visible = false
+                            }
+                            // Block menu closing
+                            contextMenu.closeOnActivation = false
+                            // Reset highlight
+                            contextMenu._setHighlightedItem(null)
+                            // Add sub-menu menu items
+                            copyOptions.createObject(contextMenu, {
+                                                         menu: contextMenu,
+                                                         username: model.username,
+                                                         password: model.password
+                                                     })
+                        }
                     }
-                }
-                MenuItem {
-                    //% "Delete"
-                    text: qsTrId("sailfish_browser-me-login_delete")
-                    onClicked: {
-                        remove(model.uid);
+                    MenuItem {
+                        //% "Edit"
+                        text: qsTrId("sailfish_browser-me-login_edit")
+                        onClicked: {
+                            var page = pageStack.animatorPush("EditLoginPage.qml", {
+                                                                  loginModel: loginModel,
+                                                                  secureAction: secureAction,
+                                                                  uid: model.uid,
+                                                                  hostname: model.hostname,
+                                                                  username: model.username,
+                                                                  password: model.password
+                                                              })
+                        }
+                    }
+                    MenuItem {
+                        //% "Delete"
+                        text: qsTrId("sailfish_browser-me-login_delete")
+                        onClicked: {
+                            remove(model.uid);
+                        }
                     }
                 }
             }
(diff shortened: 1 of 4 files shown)


llewelld:
Rebased.

llewelld:
Rebased.

llewelld:
Rebased.

llewelld:
Rebased on top of upstream/next after merging in the prior PRs. This should be a lot easier to follow now.

llewelld:
Rebased to include the merge commit.

apps/browser/qml/pages/components/SecureAction.qml
```diff
@@ -0,0 +1,33 @@
...
+import com.jolla.settings.system 1.0
+import org.nemomobile.devicelock 1.0
+
+DeviceLockQuery {
+    id: deviceLockQuery
+    property bool requireAuthentication: true
+    property string message
+
+    function perform(resolve) {
+        if (requireAuthentication && _availableMethods !== Authenticator.NoAuthentication) {
```
rainemak, line 21: There are few sources possibly causing this misunderstanding but this should be more like SecurityCodeSettings.set instead of checking available method. To clarify, available methods are for device lock automatic locking setting. More specificly, Authenticator.NoAuthentication is available method when automatic locking is set to "Not in use". That still means that device lock code is set if SecurityCodeSettings.set.

apps/browser/qml/pages/LoginsPage.qml
```diff
@@ -33,6 +33,11 @@ Page {
         return hostname
     }
 
+    function copyPassword(password) {
+        Clipboard.text = password
```
rainemak, line 37: Not completely sure if we should have an extra clipboard property for sensitive data so that vkb would not show. That said, that would be a new task not related to this one per se.

apps/browser/qml/pages/components/SecureAction.qml
```diff
@@ -0,0 +1,33 @@
...
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.0
+import com.jolla.settings.system 1.0
+import org.nemomobile.devicelock 1.0
+
+DeviceLockQuery {
```
rainemak, line 15: Looks that there's some weird flickering when this is fading away. Hmm, is there page (dialog) replace maybe happening?

apps/browser/qml/pages/components/SecureAction.qml
llewelld: Thanks for clarifying this, it's a really good point. I've updated the code to use `SecurityCodeSettings.set` instead.

apps/browser/qml/pages/LoginsPage.qml
llewelld: It's a really good idea in my opinion.

apps/browser/qml/pages/components/SecureAction.qml
llewelld: Yes, I get this too. I'm at a bit of a loss right now to understand why. Both cancel and accept use the same `pageStack.pop()` with `backNavigation` set to `false`, which should perform a fade back to the previous page. But the glitch when accepting is really obvious. Allowing back navigation solves the problem because it animates as a slide rather than a fade, but this doesn't seem to be a sensible approach (for multiple reasons).

 I'll persevere to try to find a solution.

apps/browser/qml/pages/components/SecureAction.qml
adenexter: You shouldn't need to check at all, there is some sort of option to force a prompt but the default behavior is to authenticate if no code is set.

apps/browser/qml/pages/components/SecureAction.qml
adenexter: There's probably an absence of page background duct tape for that transition.

If you're not invested in the having the security prompt inline in the application*, then you can use the device lock API directly and it will display a system dialog instead which won't have any page stack issues. 

* I have thoughts on the security credentials of asking the user to type their security code into some app.

apps/browser/qml/pages/components/SecureAction.qml
adenexter: Sorry, the requestPermission() function on Authenticator will still display a confirmation prompt if no security methods are enabled, it's authenticating a challenge code that will auto grant unless told otherwise.

Checking the availableMethods should be correct though. Those are the available methods for authentication, unlocking the device is done through the DeviceLock object and available methods for that aren't exposed through the API because that and the device lock state are managed internally in the daemon. If Authenticator.availableMethods changed based on the automatic locking setting, that would be a bug.

apps/browser/qml/pages/components/SecureAction.qml
adenexter: Some privileged group misunderstanding aside, this appears to work.

```
Authenticator {
    id: deviceLockQuery
    property bool requireAuthentication: true
    property string message
    property var _resolve

    function perform(resolve) {
        if (requireAuthentication && availableMethods !== Authenticator.NoAuthentication) {
            _resolve = resolve

            deviceLockQuery.requestPermission(message, {})
        } else {
            resolve()
        }
    }

    onPermissionGranted: {
        if (_resolve) {
            requireAuthentication = false
            _resolve()
        }
    }
}
```

apps/browser/qml/pages/components/SecureAction.qml
llewelld: I've switched to using the `Authenticator` code you suggested [below](https://github.com/sailfishos/sailfish-browser/pull/838#discussion_r647931330) @adenexter, which uses `availableMethods`.

apps/browser/qml/pages/components/SecureAction.qml
llewelld: Thanks Andrew, this has been really instructive. I've switched to using this code. If the browser is run without privileges, or there are no `availableMethods` for some other reason, then the password access methods are now disabled, which I think should have already been the case.

This also deals with the transition issues. Thanks again.

It's a separate issue, but the "full screen devicelock" approach is used in the WLAN editor (Settings > WLAN > Edit entry > unhide password). Perhaps that would also be better as an Authenticator rather than a DeviceLockQuery. It doesn't have the transition issues of course, but I feel like this "enter pin to show password" pattern is something we should be using in other places too, such as Accounts.

llewelld:
I've updated the PR with Andrews nice suggestion to use an Authenticator. I've also made various other design changes following a discussion with Martin, so these changes could all do with checking again.

apps/browser/qml/pages/components/SecureAction.qml
adenexter: Settings is a bit of a funny case because it has a number of settings pages that use inline security dialogs, including the actual device lock settings, and fingerprint settings. So do you have internal consistency in settings, or consistency in password fields, or flesh out the system dialogs so they can fully replace the inline ones?

adenexter:
LGTM

rainemak:
> LGTM

These LGTMs get lost easily in the middle. Please use as well https://github.com/sailfishos/sailfish-browser/pull/838/files && "Review changes"

rainemak:
@llewelld please resolve discussions that have been fixed already.

adenexter:
> 
> 
> > LGTM
> 
> These LGTMs get lost easily in the middle. Please use as well https://github.com/sailfishos/sailfish-browser/pull/838/files && "Review changes"

Thanks, I was looking for the button but on the files changed page did not occur to me. I am going to distractedly merge something instead of approving one day.

adenexter (approved):

apps/browser/qml/pages/LoginsPage.qml
llewelld: I looked into this further. Unfortunately there's no current capability for flagging sensitive data in the clipboard. It wouldn't be hard to enforce in the keyboard, the implementation difficulty is in how to lett the keyboard know. QClipboard doesn't have any feature to flag sensitive content. A `sensitive` flag could be sent directly to the keyboard rather than via the clipboard when the sensitive contents is set, it just requires some work in choosing the best way to communicate it to the keyboard.

This is just background info. I'll resolve this conversation for now and maybe we can return to it in the future.

apps/browser/qml/pages/components/SecureAction.qml
llewelld: Yes, I can see the difficulty. I have my own opinions about how it should all work of course, but best left for another time. I'll resolve this conversation now, as I think we covered what's needed for this PR. Hopefully we can revisit the wider issue in the future.

llewelld:
For info, this will introduce the following new strings:
```
//% "Username copied"
qsTrId("sailfish_browser-me-login_copied_username")

//% "Passwords"
qsTrId("settings_browser-la-passwords")

//% "Copy username"
qsTrId("sailfish_browser-me-login_copy_username")

//% "Copy"
qsTrId("sailfish_browser-me-login_copy")

//% "Unlock access to browser passwords"
qsTrId("sailfish_browser-me-login_unlock_password_access")

//% "Passwords"
qsTrId("settings_browser-la-passwords")
```

sledges (approved):
:+1: from L10n

xlise:
LGTM

rainemak (approved):
Just stamping I count that @adenexter checked this already. I did check the previous version on the device.

llewelld:
Rebased.

PR #839, opened 2021-05-10
Add unit tests for the browser login and passwords page

llewelld:
Builds on #838.

Current WIP implementation gets the existing webview tests to pass, I'll add the login unit tests on top in the same PR.

### DIFF
--- a/apps/browser/settings/declarativeloginmodel.cpp
+++ b/apps/browser/settings/declarativeloginmodel.cpp
@@ -57,6 +57,11 @@ void DeclarativeLoginModel::modify(int uid, const QString &username, const QStri
         return;
     }
 
+    if (!canModify(uid, username, password)) {
+        qWarning() << "Can't modify login entry to one that already exists";
+        return;
+    }
+
     LoginInfo newLogin = oldLogin;
     newLogin.setUsername(username);
     newLogin.setPassword(password);
@@ -119,6 +124,8 @@ void DeclarativeLoginModel::handleRecvObserve(const QString &message, const QVar
 {
     if (message == ALL_LOGINS) {
         setLoginList(qvariant_cast<QVariantList>(data));
+        // We're only interested in receiving the message once
+        SailfishOS::WebEngine::instance()->removeObserver(ALL_LOGINS);
     }
 }
 
@@ -135,6 +142,7 @@ void DeclarativeLoginModel::setLoginList(const QVariantList &data)
         m_logins.append(UidLoginInfo(m_nextUid, LoginInfo(varMap)));
         ++m_nextUid;
     }
+
     endResetModel();
     emit countChanged();
 
@@ -171,7 +179,7 @@ void DeclarativeLoginModel::remove(int uid)
 
     // Update the index (every item above has its position decremented)
     for (int pos = index; pos < m_logins.count(); ++pos) {
-        m_index.insert(m_logins.at(pos).first, index);
+        m_index.insert(m_logins.at(pos).first, pos);
     }
 
     endRemoveRows();
--- a/tests/auto/tests.xml
+++ b/tests/auto/tests.xml
@@ -16,7 +16,7 @@
            <case manual="false" name="dbmanager">
                <step>cd /opt/tests/sailfish-browser/auto/ &amp;&amp; ./tst_dbmanager</step>
            </case>
-           <case manual="false" timeout="300" name="webview">
+           <case manual="false" name="webview">
                <step>cd /opt/tests/sailfish-browser/auto/ &amp;&amp; LOW_MEMORY_DISABLED=1 ./tst_webview -platform wayland-egl</step>
            </case>
            <case manual="false" name="desktopbookmarkwriter">
--- a/tests/auto/tests.xml
+++ b/tests/auto/tests.xml
@@ -31,6 +31,9 @@
            <case manual="false" name="webpagefactory">
                <step>cd /opt/tests/sailfish-browser/auto/ &amp;&amp; ./tst_webpagefactory</step>
            </case>
+           <case manual="false" name="logins">
+               <step>cd /opt/tests/sailfish-browser/auto/ &amp;&amp; LOW_MEMORY_DISABLED=1 ./tst_logins -platform wayland-egl</step>
+           </case>
            <post_steps>
                <step>/usr/sbin/mcetool -jdisabled -Doff</step>
            </post_steps>
--- a/tests/auto/tst_logins/tst_logins.qml
+++ b/tests/auto/tst_logins/tst_logins.qml
@@ -1,0 +1,63 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ * Copyright (c) 2021 Open Mobile Platform LLC.
+ */
+
+import QtQuick 2.1
+import Sailfish.Silica 1.0
+import Sailfish.Browser 1.0
+
+ApplicationWindow {
+    id: window
+
+    property alias webView: webView
+    property alias tabModel: webView.tabModel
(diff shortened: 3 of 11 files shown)


llewelld:
Relies on https://git.sailfishos.org/mer-core/qtmozembed/merge_requests/92

llewelld:
Rebased, removing WIP.

llewelld:
Rebased.

llewelld:
Rebased.

apps/core/faviconmanager.h
```diff
@@ -0,0 +1,53 @@
...
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef FAVICONMANAGER_H
+#define FAVICONMANAGER_H
+
+#include <QObject>
+#include <QList>
+#include <QPointer>
```
okodron, line 16: Not necessary here

apps/browser/settings/declarativeloginmodel.cpp
```diff
@@ -102,13 +110,16 @@ QHash<int, QByteArray> DeclarativeLoginModel::roleNames() const
...
     roles[PasswordRole] = "password";
+    roles[FavIconRole] = "favicon";
     return roles;
 }
 
 void DeclarativeLoginModel::handleRecvObserve(const QString &message, const QVariant &data)
 {
     if (message == ALL_LOGINS) {
         setLoginList(qvariant_cast<QVariantList>(data));
+        // We're only intersted in receiving the message once
```
llewelld, line 121:
```suggestion
        // We're only interested in receiving the message once
```

okodron:
Should be rebased.

apps/core/faviconmanager.h
llewelld: Thanks; fixed now.

apps/browser/settings/declarativeloginmodel.cpp
llewelld: Fixed.

llewelld:
Rebased on to of jb54078/#838.

okodron (approved):

llewelld:
Rebased on top of the #838 changes.

llewelld:
Rebased on top of the (now merged) changes in #838.

tests/auto/tst_logins/tst_logins.qml
```diff
@@ -0,0 +1,68 @@
+/****************************************************************************
+**
+** Copyright (c) 2014 Jolla Ltd.
+** Contact: Raine Makelainen <raine.makelainen@jolla.com>
```
rainemak, line 4: Drop

tests/auto/tst_logins/tst_logins.qml
```diff
@@ -0,0 +1,68 @@
+/****************************************************************************
+**
+** Copyright (c) 2014 Jolla Ltd.
```
rainemak, line 3: Update

apps/browser/settings/declarativeloginmodel.cpp
```diff
@@ -119,6 +124,8 @@ void DeclarativeLoginModel::handleRecvObserve(const QString &message, const QVar
 {
     if (message == ALL_LOGINS) {
         setLoginList(qvariant_cast<QVariantList>(data));
+        // We're only interested in receiving the message once
+        SailfishOS::WebEngine::instance()->removeObserver(ALL_LOGINS);
```
rainemak, line 128: Asking maybe obvious things, but...
- What happens to new login after first set returned? I.e. will to still be visible?
- Is model created each time again when entering to the login settings page?

tests/auto/tst_logins/tst_logins.qml
llewelld: Thanks; I've just pushed fixes to all of the copyright headers.

tests/auto/tst_logins/tst_logins.qml
llewelld: Also updated.

tests/auto/tst_webview/tst_webview.cpp
```diff
@@ -1,9 +1,10 @@
...
-**
-** Copyright (c) 2014 Jolla Ltd.
-** Contact: Raine Makelainen <raine.makelainen@jolla.com>
-**
-****************************************************************************/
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ * Copyright (c) 2014-2021 Jolla Ltd.
```
rainemak, line 5:
```suggestion
 * Copyright (c) 2014 - 2021 Jolla Ltd.
```

apps/browser/settings/declarativeloginmodel.cpp
llewelld: Not obvious, and worth considering.

What you say is correct: the model is recreated each time the user enters the page. We could definitely keep it persistent and perform updates on change notifications from embedlite-components. But I wasn't convinced it was worth keeping it in memory the whole time, plus it avoided the worry of the two (UI model and gecko's model) getting out of sync. I also figured there wouldn't be any situation in which the model could be updated while the page was open (since any change would usually be triggered by some user interaction).

In this respect it differs from the history and bookmark models, but those two have to be kept in memory the whole time as they're used by the toolbar.

If you think it's better to keep it updated, I'm happy to make the change.

llewelld:
Squashed into three commits.

tests/auto/tst_logins/components/LoginsTestHelper.js
```diff
@@ -0,0 +1,83 @@
...
+
+  __pwmgr : null, // Password Manager service
+  get _pwmgr() {
+    if (!this.__pwmgr)
+      this.__pwmgr = Cc["@mozilla.org/login-manager;1"].
+                     getService(Ci.nsILoginManager);
+    return this.__pwmgr;
+  },
+
+  observe: function(aSubject, aTopic, aData) {
```
rainemak, line 41: I'm failing to understand need for this? Why are you not using the LoginsHelper.js of the embedlite-components? The embedlite-components part if essential bit of this whole thing and now there's some "extra" that we need to remember to fix always when we do changes.

If the issue is that "add" is missing from LoginsHelper.js, then I'd rather add that. Profile must be in case different from the normal browser use. This means that this is never tampering the passwords saved by the browser.

tests/auto/tests.xml
```diff
@@ -31,6 +31,9 @@
            <case manual="false" name="webpagefactory">
                <step>cd /opt/tests/sailfish-browser/auto/ &amp;&amp; ./tst_webpagefactory</step>
            </case>
+           <case manual="false" timeout="300" name="logins">
```
rainemak, line 34: I'd rather remove other timeouts as well.

tests/auto/tst_logins/components/LoginsTestHelper.js
llewelld: As discussed elsewhere, this was just about keeping the `add()` and `removeAll()` actions out of the non-test codebase where they're not used. However, I agree that it'll be more succinct and easier to maintain if they're kept together in the same place.

To this end I've moved what was the content of LoginsTestHelper.js into LoginsHelper,js, so this now depends on this change to embedlite-components: https://github.com/sailfishos/embedlite-components/pull/3

tests/auto/tests.xml
llewelld: This is now fixed so the tests cases run in under 90 seconds; I've removed the extended timeout.

llewelld:
Rebased.

rainemak (approved):
Much better! Let's keep on improving test cases as we go.

okodron (approved):

llewelld:
Rebased.

PR #843, opened 2021-05-20
Start browser always on normal browsing mode

rainemak:


rainemak:
I think @okodron that this should pretty much cover what we discussed earlier.

mkenttala (approved):

okodron:
> I think @okodron that this should pretty much cover what we discussed earlier.

Yes, now it works the same way as FF does.

okodron (approved):

PR #848, opened 2021-05-28
Avoid duplicating ConfigurationValue

llewelld:
If two or more ConfigurationValue instances are duplicated simultaneously, the values are not correctly updated for sandboxed apps. There is no problem if only a single instance exists at a time.

This change ensures the ConfigurationValue for the home page is shared rather than duplicated when the folder picker is pushed.

This doesn't address the underlying Sailjail issue, however even once addressed, I believe this change is appropriate anyway (since it shouldn't be necessary to use DBus to send data within the same app).

### DIFF
--- a/apps/browser/qml/pages/SettingsPage.qml
+++ b/apps/browser/qml/pages/SettingsPage.qml
@@ -82,7 +82,10 @@ Page {
                         textFormat: Text.StyledText
                         color: highlighted ? Theme.highlightColor : Theme.primaryColor
                         text: Theme.highlightText(title, site, Theme.highlightColor)
-                        onClicked: pageStack.animatorPush(Qt.resolvedUrl("components/AddHomePageDialog.qml"))
+                        onClicked: {
+                            pageStack.animatorPush(Qt.resolvedUrl("components/AddHomePageDialog.qml"),
+                                                   {homePageConfig: homePageConfig})
+                        }
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


rainemak:
LGTM.

rainemak (approved):

adenexter:
If duplication is an issue, it might be an idea to collect everything under /apps/sailfish-browser/settings or even /apps/sailfish-browser into a ConfigurationGroup singleton.

adenexter:
It's odd that this would be sandbox related though, my recollection is there's a single dconf object backing everything.

okodron (approved):

llewelld:
> It's odd that this would be sandbox related though, my recollection is there's a single dconf object backing everything.

It's a very specific issue. Writes to the the dconf file work fine (I can see the file updated outside the sandbox). DBus signals sent to dconf client instances are received fine. However, when the client receives the signal, it reverts the value back to what it was when the instance was created, rather than to the updated value. It only happens when the app is sandboxed, so I thought this might be because the sandboxed client is reading a copy of the dconf file internal to the sandbox that doesn't get updated, but I can't find evidence in the code or config of this. So, it's confusing to me too.

llewelld:
Rebased and fixed conflict.

llewelld:
> If duplication is an issue, it might be an idea to collect everything under /apps/sailfish-browser/settings or even /apps/sailfish-browser into a ConfigurationGroup singleton.

Makes sense, I'll follow this up separately.

PR #852, opened 2021-05-31
[browser] Show saved logins placeholder only after model populated. Fixes JB#54479 OMP#JOLLA-172

rainemak:


apps/browser/settings/declarativeloginmodel.cpp
```diff
@@ -127,6 +133,11 @@ void DeclarativeLoginModel::setLoginList(const QVariantList &data)
     }
     endResetModel();
     emit countChanged();
+
+    if (!m_populated) {
```
atatarov, line 137: I hope that now we do not assume the possibility of re-updating the model and we do not need to update "popuplated" more than once

llewelld:
I didn't get a chance to test it, but the code looks just fine to me.

llewelld (commented):
LGTM.

llewelld (approved):

apps/browser/settings/declarativeloginmodel.cpp
llewelld: It's not clear from this PR alone, but I've make this assumption in other parts of the code too, given the model gets recreated every time you enter the page.

In the unit test PR for the login page (under review and to be applied alongside this) the observer gets deregistered so that it won't update (see line 122 of declarativeloginmode.cpp here https://github.com/sailfishos/sailfish-browser/pull/839).

atatarov (approved):

okodron (approved):

PR #808, opened 2021-03-17
[sailfish-browser] New site pop-up menu. Contributes to JB#51692

aroditelev:
The menu should be displayed by pressing a button, hidden by tap in a free area:
Screenshot_20210316_009

PR #853, opened 2021-06-01
[browser] Ensure toolbar icons are aligned. Contributes to JB#51981 OMP#JOLLA-163

chriadam:


apps/browser/qml/pages/components/OverlayListItem.qml
```diff
@@ -24,7 +24,7 @@ BackgroundItem {
     Item {
         id: iconContainer
 
-        width: iconWidth
+        width: iconWidth + horizontalOffset
```
xlise, line 27: The indent from the left edge will increase by `horizontalOffset + horizontalOffset/2`  so and supposed?

chriadam: It will increase, yes, to match the indent used by the addTabButton in the SecondaryBar
See: https://github.com/sailfishos/sailfish-browser/blob/master/apps/browser/qml/pages/components/SecondaryBar.qml#L83

Or, have I made some mistake (I don't have tablet to test on, so I might be missing something in that case?)

llewelld (commented):
I think the original reason for this change (aligning the `SecondaryBar` items with the tab icon on the left hand side of the toolbar) has been superseded by the popup menu changes now merged into devel (#808). From what I can tell the `OverlayListItem` is now only used in this pop up menu. 

Would this be a sensible change to make anyway, or is it no longer needed?

chriadam:
> I think the original reason for this change (aligning the `SecondaryBar` items with the tab icon on the left hand side of the toolbar) has been superseded by the popup menu changes now merged into devel (#808). From what I can tell the `OverlayListItem` is now only used in this pop up menu.

You're right.  From a quick grep of the source code, SecondaryBar.qml is no longer used, and should probably be removed now?  There is only one reference to it (in a comment, which I guess is now wrong and should be changed to mention the pop up menu instead) as far as I can see, and no uses of it.

> Would this be a sensible change to make anyway, or is it no longer needed?

I think it is no longer needed, but perhaps @rainemak could take a quick look to make sure I'm not misunderstanding.

llewelld:
I agree with you Chris about the secondary bar not being needed any more. I should have mentioned #859 which slips in some changes to remove it (not obvious from the PR title I'm afraid, as at the time it seemed like housekeeping); my bad and I apologise.

chriadam:
No longer needed due to other changes.

PR #854, opened 2021-06-01
Align items on Settings page

llewelld:
Ensures icons and glassitems are vertically aligned at their centres and that the text is vertically aligned on the left.

### DIFF
--- a/apps/browser/qml/pages/SettingsPage.qml
+++ b/apps/browser/qml/pages/SettingsPage.qml
@@ -22,6 +22,8 @@ import "components"
 Page {
     id: page
 
+    readonly property int _textSwitchIconCenter: Math.round((permissionIcon.width - Theme.itemSizeExtraSmall) / 2.0)
+
     function removeProtocolTypeFromUri(uri) {
         if (uri.length === 0)
             return uri
@@ -57,7 +59,7 @@ Page {
                 label: qsTrId("settings_browser-la-home_page")
                 //% "Default"
                 value: homePageConfig.value === "about:blank" ? qsTrId("sailfish_browser-la-home_page_default") : removeProtocolTypeFromUri(homePageConfig.value)
-                leftMargin: 2*Theme.horizontalPageMargin + homePageIcon.width
+                leftMargin: Theme.horizontalPageMargin + homePageIcon.width + Theme.paddingMedium
                 contentHeight: Theme.itemSizeMedium
 
                 Icon {
@@ -97,7 +99,7 @@ Page {
                 //: Label for combobox that sets search engine used in browser
                 //% "Search with"
                 label: qsTrId("settings_browser-la-search_with")
-                leftMargin: 2*Theme.horizontalPageMargin + searchIcon.width
+                leftMargin: Theme.horizontalPageMargin + searchIcon.width + Theme.paddingMedium
                 contentHeight: Theme.itemSizeMedium
 
                 Icon {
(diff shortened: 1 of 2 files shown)


llewelld:
Screenshot showing the updated alignment.

BrowserSettingsAligned2

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -22,6 +22,8 @@ import "components"
 Page {
     id: page
 
+    readonly property int _textSwitchGlassCentre: Math.round((permissionIcon.width - Theme.itemSizeExtraSmall) / 2.0)
```
chriadam, line 25: maybe _textSwitchIconCentre -- err, I mean _textSwitchIconCenter as I think we are supposed to use US english -- would be a bit clearer, as "glass" may be used as either an adjective or a noun

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -152,6 +154,9 @@ Page {
                 description: qsTrId("settings_browser-la-close_all_tabs_description")
                 checked: closeAllTabsConfig.value
                 enabled: AccessPolicy.browserEnabled
+                // Margins adjusted to align with other items on the page
+                leftMargin: Theme.horizontalPageMargin + Theme.paddingLarge + _textSwitchGlassCentre
```
chriadam, line 158: sorry, where does this paddingLarge come from?

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -152,6 +154,9 @@ Page {
...
+                _label.anchors.leftMargin: Theme.paddingMedium + _textSwitchGlassCentre
```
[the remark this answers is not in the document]
llewelld: It was apparently added a couple of years back to allow the wrapping to be changed, but a lucky addition as far as this is concerned.

apps/browser/qml/pages/SettingsPage.qml
llewelld: Thanks; I've changed it as you suggested. The naming is awkward because it doesn't really have a practical purpose on its own, it  just happens to be useful for the calculation. I don't want it to give it a name that might lead someone to try to use it. `_textSwitchIconCenter` is suitable vague though ;)

apps/browser/qml/pages/SettingsPage.qml
llewelld: It's from the TextSwitch code. The left margin for the toggle item is set there as:
```
68:             left: parent.left; leftMargin: root.leftMargin - Theme.paddingLarge
```

So we have natural centre of the ComboBox icons (fixed):
`Theme.horizontalPageMargin + (permissionIcon.width) / 2)`

Centre for the TextSwitch Glass items (variable):
`root.leftMargin - Theme.paddingLarge + (Theme.itemSizeExtraSmall / 2)`

Equating and rearranging:
`root.leftMargin = Theme.horizontalPageMargin + Theme.paddingLarge + (permissionIcon.width) / 2) - (Theme.itemSizeExtraSmall / 2)`

apps/browser/qml/pages/components/BrowserListItem.qml
```diff
@@ -47,6 +47,8 @@ ListItem {
 
             width: parent.width - icon.width
             anchors.verticalCenter: parent.verticalCenter
+            leftMargin: Theme.paddingMedium
+            rightMargin: 0
```
atatarov, line 51: does it make any sense?

llewelld: It's a good question. I noticed that the `ValueButton` is inside a `Row` with the following properties:
```
        width: parent.width - 2*Theme.horizontalPageMargin
        x: Theme.horizontalPageMargin
```
So there is already a right margin of `Theme.horizontalPageMargin` applied to the item. If I'm understanding this code correctly, setting `rightMargin: 0` should leave it with a right margin of `Theme.horizontalPageMargin`. Please correct me if this is wrong or not what's needed.

rainemak: This is correct analysis as far as I see. ValueButton itself has a Column which is anchorized rightMargin bound to parent right.

atatarov (approved):

rainemak (approved):
LGTM. Let's merge this.

llewelld:
Rebased.

PR #855, opened 2021-06-04
[sailfish-browser] Expand the popup menu as it is dragged up. Fixes JB#54530 OMP#JOLLA-185

adenexter:


adenexter:
This works as is, but fillMode options on Background all do the wrong thing and do some aspect ratio scaling where none was needed here. I've worked around it by assigning an invalid value to the property which effectively chooses a no scaling option but that should be given an official name in Silica.

apps/browser/qml/pages/components/PopUpMenu.qml
```diff
@@ -7,90 +7,177 @@
...
+
+                        x: contentLoader.x
+                        y: Math.max(popUpMenu.margin, contentLoader.y - menuFlickable.contentY)
+                        z: -1
+                        width: footerLoader.width
+                        height: footerLoader.y - y
+
+                        color: Qt.tint(
+                                    popUpMenu.palette.colorScheme === Theme.LightOnDark ? "black" : "white",
+                                    Theme.rgba(popUpMenu.palette.primaryColor, Theme.opacityFaint))
```
adenexter, line 107: I was a little unclear as to the purpose of the ColorOverlay item, and whether it was supposed to tint the entire window or just adjust the background color. If it is the latter then adjusting the background color like this is much more efficient, and if it's the former then a Rectangle with the color Theme.rgba(popUpMenu.palette.primaryColor, Theme.opacityFaint) should achieve the same effect as well.

aroditelev: @adenexter ```ColorOverlay```  just adjust the background color

adenexter: Thanks, this should be okay as is then.

rainemak (changes requested):
- Please align to footer of the popup menu with the toolbar height -- toolbar top and popup menu footer top should be vertically aligned
- Clicks go now through now from the footer to the actual popup menu content. For example in between of footer items and good example is the disabled forward button

adenexter:
The alignment is fixed, the margins above and below the icons in landscape are getting a skinny but it's high enough to fit them still.
Click through is also fixed, in part by making the icon buttons a little bigger and removing the non-interactive space between them, and then putting a MouseArea behind it all because the forward button can be disabled. It did make the menu slightly wider by 8 reference pixels but the forward button and the menu icons above still seem to line up pretty well.

rainemak (approved):
LGTM for this. If there are more issues, let's tackle them separately.

@aroditelev or @atatarov please review as well.

atatarov (approved):

aroditelev:
Looks nice.  LGTM

rainemak:
Rebase before merging.

PR #856, opened 2021-06-04
[browser] Add popup menu for tabs view. Contributes to JB#50860

xlise:


apps/browser/qml/pages/components/TabView.qml
```diff
@@ -203,10 +200,89 @@ Item {
...
+        id: popupMenu
+
+        signal closeAllTabs
+
+        anchors {
+            right: parent.right
+            rightMargin: Theme.paddingMedium
+            bottom: parent.bottom
+            bottomMargin: Theme.paddingMedium
+        }
```
adenexter, line 222: After rebasing this on next you'll need to change this so the menu item fills the page. The margins will be taken care of internally.

apps/browser/qml/pages/components/PopUpMenu.qml
```diff
@@ -28,7 +28,7 @@ Item {
     signal closed
 
     width: Math.max(Theme.paddingLarge * widthRatio, footer.implicitWidth)
-    height: Math.min(parent.height - Theme.paddingLarge * 2, _menuItemHeight)
+    height: Math.min(parent.height - Theme.paddingLarge * 2, _menuItemHeight, container.contentHeight + footer.height)
```
adenexter, line 31: Rebase conflict. You should be able to drop the change after.

adenexter (commented):
You'll need to make a couple of changes to adapt to some changes made to PopUpMenu. Also I've spotted a small regression in the menu itself which results in it being open initially on the tabs page. This line here https://github.com/sailfishos/sailfish-browser/pull/857/files#diff-d71470f19a8849ea7dad86bf1d4483867a944c9adf6e15a366b107f4b624c99fR35 will fix that.

apps/browser/qml/pages/components/PopUpMenu.qml
xlise: Rebased. done

apps/browser/qml/pages/components/TabView.qml
xlise: done

apps/browser/qml/pages/components/TabGridView.qml
```diff
@@ -41,6 +41,24 @@ SilicaGridView {
...
+    function closeAllTabs() {
+        remorsePopup = Remorse.popupAction(
+                    tabGridView,
+                    //% "Closed all tabs"
+                    qsTrId("sailfish_browser-closed-all-tabs"),
+                    function() {
+                        tabGridView.closeAll()
+                        remorsePopup = null
+                    })
+        closingAllTabs = true
```
llewelld, line 53: The remorse timer creates some complexity, because it's possible to switch between normal and private browsing while the remorse is still running. This can mean that the user selects "Close all tabs" in private mode, but it's the normal browsing tabs which get closed. These are the steps I followed:
1. Open tab view.
2. Select batman to show the private tabs.
3. Select "Close all tabs" from the menu.
4. While the remorse timer is still running, open the menu again and select "New tab".

In this case, the model switches to normal browsing mid-remorse and the wrong set of tabs gets closed. I'm not sure what the solution is, but one way might be to block the UI from switching views while the remorse is running.

apps/browser/qml/pages/components/TabView.qml
```diff
@@ -203,10 +200,84 @@ Item {
...
+
+        anchors.fill: parent
+        active: false
+
+        menuItem: Component {
+            Item {
+                id: menuItem_
+                readonly property int iconWidth: Theme.iconSizeMedium + Theme.paddingLarge
+                readonly property int verticalPadding: 3 * Theme.paddingSmall
+                height: content.height + verticalPadding * 2
```
llewelld, line 225: Because the `PopupMenu` already has a margin at the bottom set to `Theme.paddingMedium`, you end up with a larger margin at the bottom than at the top. Personally I think it looks better with the same margin at the top and bottom; perhaps it could be balanced unless this is intentional?

apps/browser/qml/pages/components/TabView.qml
```diff
@@ -203,10 +200,84 @@ Item {
...
+                        }
+
+                        OverlayListItem {
+                            height: Theme.itemSizeSmall
+                            iconWidth: menuItem_.iconWidth
+                            iconSource: "image://theme/icon-m-incognito-new"
+                            //% "New private tab"
+                            text: qsTrId("sailfish_browser-la-new_private_tab")
+                            onClicked: {
+                                tabView.privateMode = true
```
llewelld, line 255: Because of the switch between modes, and the page sliding back, there can be a strange sliding effect if "New private tab" is selected while in normal mode, or vice versa (the tabs slide in one direction while the page slides in the other). Could the mode switch be delayed to avoid this (or something similar)? I appreciate this may not be straightforward to implement, so just a suggestion.

apps/browser/qml/pages/components/TabGridView.qml
```diff
@@ -15,7 +15,7 @@ import org.nemomobile.configuration 1.0
 import "." as Browser
```
llewelld, line 15: I think `org.nemomobile.configuration` and `"."` aren't needed and can be removed.

apps/browser/qml/pages/components/TabView.qml
```diff
@@ -17,8 +17,8 @@ import org.nemomobile.configuration 1.0
 import "." as Browser
 import "../../shared" as Shared
```
llewelld, line 18: I think these aren't needed either.
```suggestion
```

apps/browser/qml/pages/components/TabView.qml
```diff
@@ -80,32 +75,34 @@ Item {
...
-                color: Theme.colorScheme == Theme.LightOnDark ? "black" : "white"
-                width: parent.width
-                height: Math.max(0, _pulleyMenuHeight - y)
-                y: Math.max(0, _pulleyMenuHeight - _tabView.contentY + _tabView.originY)
+            Connections {
+                target: popupMenu
+                onCloseAllTabs: {
+                    if (isCurrentItem)
+                        _tabView.closeAllTabs()
+                }
```
llewelld, line 105: I notice that the `popupMenu.closeAllTabs` signal goes into the `tabGridView`, then comes out again and gets signalled all the way out to `Overlay`. I've not tried this, so just a suggestion, but it might be worth checking whether this can be simplified by moving the `TabGridView.closeAllTabs()` method out of `TabGridView` and into `TabView`. From what I can tell, it's only needed inside `TabGridView` in order to set the `closingAllTabs` property there.

llewelld (changes requested):
Re-use of the `PopUpMenu` is working really nicely here. I've added some small suggestions, but the only real issue I see is how to deal with closing all tabs while also switching between modes while the remorse timer is running.

apps/browser/qml/pages/components/TabView.qml
adenexter: The effect arises from the the (ab)use of Background in the popup menu to round the menu corners. I might have to have another go at decoupling the texture coordinates from the screen coordinates so that page movement isn't an issue.

It would still be a good idea to delay any action that ends up animating a page until after the menu closes. Or delay closing the menu until after the page is off screen as concurrent animations often end up looking messy.

llewelld: Ah, actually I wasn't referring to this effect. I had something simpler in mind: if the tab switches, while the page pops, then the tab pane is sliding from left to right (or right to left). At the same time the page is sliding right (back up the page stack). The result is that the tab ends up moving either double speed (right + right = double speed), or appears roughly static (left + right = static). It's not quite that simple because the transitions aren't linear, but hopefully you get the idea. The combined effect looks a little odd in my opinion, but it doesn't look broken.

What you describe also sounds like an important issue, but I'd like to distinguish the two to avoid confusion.

apps/browser/qml/pages/components/TabGridView.qml
xlise: done

apps/browser/qml/pages/components/TabView.qml
xlise: done

apps/browser/qml/pages/components/TabView.qml
xlise: dropped binding for TabView. done

apps/browser/qml/pages/components/TabGridView.qml
xlise: done

apps/browser/qml/pages/components/TabView.qml
xlise: done

apps/browser/qml/pages/components/TabView.qml
xlise: I tried to fix it, it stops working with this approach

xlise:
The changes are not final, because the popup menu does not work now

adenexter:
> The changes are not final, because the popup menu does not work now

What is the issue with the popup menu? I had to rebase the change for it to build due to a gmock dependency issue, but after that it appeared to work okay.

apps/browser/qml/pages/components/TabView.qml
```diff
@@ -203,10 +200,84 @@ Item {
...
+                                tabView.privateMode = true
+                                tabView.enterNewTabUrl()
+                            }
+                        }
+
+                        OverlayListItem {
+                            height: Theme.itemSizeSmall
+                            iconWidth: menuItem_.iconWidth
+                            iconSource: "image://theme/icon-m-tab-close"
+                            enabled: showCloseAllAction.value && webView.tabModel.count
```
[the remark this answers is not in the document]
llewelld: The disabled elements are now clearly distinguishable by being greyed out; thank you. I'll resolve this discussion.

apps/browser/qml/pages/components/TabView.qml
llewelld: The animations in and out of the tab view while also switching to a different mode (private/normal) is much cleaner now. Thank you.

apps/browser/qml/pages/components/TabView.qml
llewelld: There still seems to be something a bit strange going on here. Maybe due to the introduction of the handle the top item is now cut off along the top, See the screenshot, with the top margin highlighted red and comparison with the main browser toolbar menu. I think the highlight bar (when pressing an entry) should be the same height for all entries in the menu.

top-margin

apps/browser/qml/pages/components/TabGridView.qml
llewelld: Great, many thanks. Resolving.

apps/browser/qml/pages/components/TabView.qml
llewelld: Thank you also here.

apps/browser/qml/pages/components/TabGridView.qml
llewelld: This is definitely much improved, but I'm still experiencing some unexpected behaviour here, with the tab view getting confused about whether there are, or are not, tabs open. Here's the sequence I follow to hit the unexpected behaviour:

1. Open browser and clear all tabs from both normal and private modes.
2. Close and re-open browser.
3. Wait for the default homepage to load.
4. Open the tab view.
5. Select the "private mode" icon at the top of the screen.
6. Select the "normal mode" icon at the top of the screen.
7. Press on the "x" for the single open tab to close it.

At this point the tab count goes to zero and the "Push button plus to create a new tab" text appears in the background, but the tab isn't removed. It also seems to be possible to get the "private" tab icon confused this way (see the screenshot). I'm not sure exactly why this is happening I'm afraid.

confused-tabs

apps/browser/qml/pages/components/TabView.qml
llewelld: I was hoping it might be straightforward, but can appreciate I probably missed some underlying complexity. Thank you for trying it out.

llewelld:
> I had to rebase the change for it to build due to a gmock dependency issue, but after that it appeared to work okay.

I found the same: after rebasing it built and ran okay, and the popup menu is working nicely.

xlise:
> > The changes are not final, because the popup menu does not work now
> 
> What is the issue with the popup menu? I had to rebase the change for it to build due to a gmock dependency issue, but after that it appeared to work okay.

It did not help me, I fixed everything by reinstalling the system

apps/browser/qml/pages/components/TabGridView.qml
```diff
@@ -79,74 +97,31 @@ SilicaGridView {
...
             _forcedParent: parent
-            flickable: tabView
+            flickable: tabGridView
         }
     }
 
     ViewPlaceholder {
         x: -Theme.horizontalPageMargin
         width: parent.width + Theme.horizontalPageMargin
         enabled: !model.count || closingAllTabs
```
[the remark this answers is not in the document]
xlise: Fixed

apps/browser/qml/pages/components/TabView.qml
xlise: I completely agree, done

apps/browser/qml/pages/components/TabGridView.qml
xlise: Easier to reproduce
1) Launch browser
2) Wait for the home page to load
3) Switch to private mode
4) Go to normal mode
5) Remove tab

Not perceived 
```
        beginRemoveRows(QModelIndex(), index, index);
        m_tabs.removeAt(index);
        endRemoveRows();
```

xlise: I got around this, but this is a terrible solution (
    ```
    Connections {
        target: model
        // Force update GridView when deleting a tab
        onTabClosed: {
            var oldModel = tabGridView.model
            tabGridView.model = undefined
            tabGridView.model = oldModel
        }
    }
    ```

apps/browser/qml/pages/components/TabGridView.qml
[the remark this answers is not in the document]
llewelld: Thanks; working nicely.

apps/browser/qml/pages/components/TabView.qml
llewelld: Great; thank you, it looks well balanced now.

apps/browser/qml/pages/components/TabGridView.qml
llewelld: I don't see the underlying reason for this I'm afraid, but I can confirm that your solution works here too.

xlise: How is it possible to catch such an error, according to the previous instructions, I cannot

llewelld: Just to clarify: the error no longer happens after your fix (at least, I'm not able to find a way to make it happen).

apps/browser/qml/pages/components/TabsToolBar.qml
```diff
@@ -60,8 +61,8 @@ Item {
             verticalCenter: parent.verticalCenter
```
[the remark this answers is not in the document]
xlise: done

llewelld (approved):
LGTM, nice changes @xlise.

apps/browser/qml/pages/components/TabView.qml
```diff
@@ -203,10 +200,95 @@ Item {
...
+
+        signal closeAllTabs
+
+        anchors.fill: parent
+        active: false
+
+        menuItem: Component {
+            Item {
+                id: menuItem_
+                readonly property int iconWidth: Theme.iconSizeMedium + Theme.paddingLarge
```
atatarov, line 223: minor
```suggestion
                readonly property real iconWidth: Theme.iconSizeMedium + Theme.paddingLarge
```

atatarov (approved):

apps/browser/qml/pages/components/TabView.qml
xlise: I think a uniform spelling is better 
https://github.com/sailfishos/sailfish-browser/blob/next/apps/browser/qml/pages/components/PopUpMenuItem.qml#L16

PR #857, opened 2021-06-07
[sailfish-browser] Show/hide the popup menu by sliding from/to the bottom of the screen. Fixes JB#54531 OMP#JOLLA-186

adenexter:

PR #858, opened 2021-06-07
[sailfish-browser] Display a handle at the top of the popup menu. Fixes JB#54528 OMP#JOLLA-183

adenexter:


rainemak:
I think we should review the rest as one meaning that we should "ignore" https://github.com/sailfishos/sailfish-browser/pull/857

llewelld:
This is really nice behaviour. I'm experiencing a bit of a glitch though, when a page is added to or removed from the pageStack. The horizontal offset seems to mix badly with the shader positioning. These screenshots are from selecting the History item in the menu.

menuslide

adenexter:
> This is really nice behaviour. I'm experiencing a bit of a glitch though

I commented here first, https://github.com/sailfishos/sailfish-browser/pull/856#discussion_r649624624

I guess I will have to use a custom shader for the background after all in order to change the reference coordinates for the texure coordinates. It should be trivial but I just could not make it work the other day.

adenexter:
>  It should be trivial but I just could not make it work the other day.

Now I know why I couldn't make it work. The scene graph renderer pre-transforms geometry so it can batch together different items into a single render call. So my single item which I create with the geometry (0, 0, 768x1419) gets submitted to the GPU as (200, 32, 768x1419) or something worse if the item is rotated or otherwise transformed. All of which means I can't normalize the position using the item size to get texture coordinates which is what I was trying to do.

So I can add a normalized position coordinates to the Background item's geometry, which I already wanted for something before but also ran into odd issues with at the time. Or do a limited fix by counter transforming some items based on page coordinates, which I'm already realizing won't work because of animator transitions.

rainemak:
> > It should be trivial but I just could not make it work the other day.
> So I can add a normalized position coordinates to the Background item's geometry, which I already wanted for something before but also ran into odd issues with at the time. Or do a limited fix by counter transforming some items based on page coordinates, which I'm already realizing won't work because of animator transitions.

Idea / thought: How would it look like if we'd throw away that Background and ShaderEffectSource and just have a Rectangle with a rounded corners (~matching coloring) + a little bit transparency.

adenexter:
> Idea / thought: How would it look like if we'd throw away that Background and ShaderEffectSource and just have a Rectangle with a rounded corners (~matching coloring) + a little bit transparency.

The scroll bar and item press highlights won't be clipped that way. And the footer still has a Background item or a large rectangle clipped so only its bottom corners are rounded.

I've found my problem now, it requires a change to Silica but it's something I've wanted before and probably will again.

apps/browser/qml/pages/components/PopUpMenu.qml
```diff
@@ -30,74 +30,112 @@ SilicaControl {
...
+                    id: headerItem
+
+                    y: Math.max(0, footerLoader.y - height - (Screen.sizeCategory > Screen.Medium
+                                ? contentLoader.height
+                                : Math.min(contentLoader.height, Theme.paddingLarge * popUpMenu.heightRatio)))
+
+                    width: menuFlickable.width
+                    height: Theme.paddingLarge
+
+                    Rectangle {
```
rainemak, line 105: Let's raise this above actual content items. On landscape menu expands to the max position leaving only panning of the flickable. When panning menu downwards, you can see all menu items going above the handle

apps/browser/qml/pages/components/PopUpMenu.qml
```diff
@@ -30,74 +30,112 @@ SilicaControl {
...
         }
     }
 
-    FadeAnimation {
-        id: menuFadeAnimation
+    NumberAnimation {
+        id: menuAnimation
 
         running: false
+        duration: 2000
```
rainemak, line 58: This feels now like a leftover from the testing.

rainemak, line 58:
```suggestion
        duration: 200
```

adenexter: Done

apps/browser/qml/pages/components/PopUpMenu.qml
adenexter: Done if I've understood correctly.

adenexter:
> I've found my problem now, it requires a change to Silica but it's something I've wanted before and probably will again.

A fix for this has been merged to Silica.

apps/browser/qml/pages/components/PopUpMenu.qml
```diff
@@ -117,22 +140,42 @@ SilicaControl {
                             flickable: menuFlickable
                         }
                     },
+                    Rectangle {
```
llewelld, line 143: With the new variable dimensions of `menuFlickable` I'm seeing some strange behaviour from the `VerticalScrollDecorator`. In practice it seems the values you need going in to the decorator are:
```
flickable.Height: menuFlickable.height - menuItem.menuTop - (2 * popUpMenu.margin)
flickable.contentHeight: menuFlickable.contentHeight - contentLoader.y
flickable.contentY: menuFlickable.contentY + menuItem.menuTop - contentLoader.y + (2 * popUpMenu.margin)
```
In other words, giving:
```
_sizeRatio: ((menuFlickable.height - menuItem.menuTop - (2 * popUpMenu.margin)) - _headerSpacing) / (menuFlickable.contentHeight - contentLoader.y + _topMenuSpacing + _bottomMenuSpacing)
y: Math.max(0, Math.min((parent.height / (menuFlickable.height - menuItem.menuTop - (2 * popUpMenu.margin))) * (_headerSpacing + ((menuFlickable.contentY + menuIt    em.menuTop - contentLoader.y + (2 * popUpMenu.margin)) - flickable.originY + _topMenuSpacing) * _sizeRatio), (parent.height - height)))
```

llewelld (changes requested):
This is working nicely, no artefacts when the menu slides on/off the page in any direction. There's some strange behaviour from the scroll decorator though (a minor thing, see my comment elsewhere).

apps/browser/qml/pages/components/PopUpMenu.qml
adenexter: Done, in a simplified fashion. The page header and other margins in the source item are things we explicitly don't want as the presence of a dialog header in a Page containing the PopUpMenu should not have any effect on the menu's scroll decorator.

llewelld: Okay, perfect, this is working nicely now. Thank you.

llewelld (approved):
This is working nicely now for me on an XA2. The whole menu approach does now seem quite clear and usable (it's clear if there are more elements in the menu, there's good use of the screen space, good alignment with the other items on the toolbar, etc.).

rainemak (approved):
LGTM as well.

atatarov (approved):
Thank you! I guess we don't have the solution for the tablet, but it would be nice to have ability to close "popup menu" by swiping from top to bottom :)

rainemak:
> Thank you! I guess we don't have the solution for the tablet, but it would be nice to have ability to close "popup menu" by swiping from top to bottom :)

I think there are two improvements that we should consider but let's handle both of them separately

1) Dimming the content area when popup menu like we do elsewhere => this should help users to understand that tapping outside of the popup menu will dismiss it

2) Swipe to close gesture for the popup menu

okodron:
> > Thank you! I guess we don't have the solution for the tablet, but it would be nice to have ability to close "popup menu" by swiping from top to bottom :)
> 
> I think there are two improvements that we should consider but let's handle both of them separately
> 
>     1. Dimming the content area when popup menu like we do elsewhere => this should help users to understand that tapping outside of the popup menu will dismiss it
> 
>     2. Swipe to close gesture for the popup menu

Yes, good idea. Better to file a separate task about this so that it won't be forgotten.

aroditelev (approved):

PR #861, opened 2021-06-09
Remove close button from the toolbar

llewelld:
Rolls up a few changes:

1. Remove close button from the toolbar.
    
    With the popup menu replacing the toolbar, the button can no longer be
    pressed to close the current tab. This removes the button, the tab can
    still be closed from the tab grid.

2. Ensure tab number is visible in private mode.
    
    When in private mode the tab icon on the toolbar has inverted colours.
    This change inverts the text colour so it's still visible on top of the
    icon.

3. Fix bookmark edit dialog header.
    
    The missing header was causing the dialog fields to overlap with the
    action buttons at the top of the page. This adds a header to fix it.

I can split them into separate PRs if that would make it easier to handle.
Still needs an OMP bug ref associated with it.

### DIFF
--- a/apps/browser/qml/pages/components/BookmarkItem.qml
+++ b/apps/browser/qml/pages/components/BookmarkItem.qml
@@ -106,11 +106,14 @@ ListItem {
                                                   })
             }
             MenuItem {
-                //% "Edit"
+                // Defined in FavoriteContextMenu.qml
+                // "Edit"
                 text: qsTrId("sailfish_browser-me-edit")
                 onClicked: {
                     var page = pageStack.animatorPush(editDialog,
                                            {
+                                               //% "Edit bookmark"
+                                               "description": qsTrId("sailfish_browser-he-edit-bookmark"),
                                                "url": url,
                                                "title": title,
                                                "index": bookmarkFilterModel.getIndex(model.index)
--- a/apps/browser/qml/pages/components/ToolBar.qml
+++ b/apps/browser/qml/pages/components/ToolBar.qml
@@ -139,10 +139,26 @@ Column {
             Browser.TabButton {
                 id: tabs
 
-                icon.source: webView.privateMode ? "image://theme/icon-m-incognito" : "image://theme/icon-m-tabs"
+                icon.source: {
+                    if (webView.privateMode) {
+                        return webView.tabModel.count > 0 ? "image://theme/icon-m-incognito-selected"
+                                                          : "image://theme/icon-m-incognito"
+                    } else {
+                        return "image://theme/icon-m-tabs"
+                    }
+                }
+
+                label.color: {
+                    if (webView.privateMode) {
+                        return Theme.overlayBackgroundColor ? Theme.overlayBackgroundColor : "black"
+                    } else {
+                        return highlighted ? Theme.highlightColor : Theme.primaryColor
+                    }
+                }
+
                 opacity: findInPageActive ? 0.0 : 1.0
                 horizontalOffset: toolBarRow.horizontalOffset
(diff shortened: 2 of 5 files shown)


atatarov:
Thank you! It works nicely. There is a bug
"Browser "Add to apps grid" page items overlap header "
fix 54532, I think it could be pinned there also

atatarov (approved):

llewelld:
Rebased.

> Thank you! It works nicely. There is a bug "Browser "Add to apps grid" page items overlap header " fix 54532, I think it could be pinned there also

Thanks, and thanks for flagging up JB#54532. I'd forgotten about it completely (which is a bit of a worry, given that apparently I wrote it).

llewelld:
This PR will introduce one new string:
```
//% "Edit bookmark"
qsTrId("sailfish_browser-me-edit-bookmark"),
```
I'll try to see if there's some alternative to introducing it.

apps/browser/qml/pages/components/ToolBar.qml
```diff
@@ -139,10 +139,26 @@ Column {
...
+                    } else {
+                        return "image://theme/icon-m-tabs"
+                    }
+                }
+
+                label.color: {
+                    if (webView.privateMode) {
+                        return Theme.overlayBackgroundColor ? Theme.overlayBackgroundColor : "black"
+                    } else {
+                        return down ? Theme.highlightColor : Theme.primaryColor
```
rainemak, line 155:
```suggestion
                        return highlighted ? Theme.highlightColor : Theme.primaryColor
```

Whilst end result is the same, I find highlighted usage better when coloring.

rainemak (approved):
Added minor comment, please check that before merging. Approving already.

apps/browser/qml/pages/components/ToolBar.qml
llewelld: Thanks; I've made the change.

llewelld:
I've changed the new string:
```
//% "Edit bookmark"
qsTrId("sailfish_browser-me-edit-bookmark")
```
so that now it uses the following existing string instead:
```
//% "Edit"
text: qsTrId("sailfish_browser-me-edit")
```
It's not ideal, but the best that I could find. I suggest changing it to use a new string once the localization freeze is no longer in effect.

apps/browser/qml/pages/components/BookmarkItem.qml
```diff
@@ -111,6 +111,9 @@ ListItem {
                 onClicked: {
```
sledges, line 111: There is the previous redefinition of `sailfish_browser-me-edit` just above this line (number 109 at the time of writing), which says "Edit" instead of "Edit bookmark".

Another "Edit" with the same ID is defined here:
https://github.com/sailfishos/sailfish-browser/blob/3ed8aa667d13fd0820706e7782bb610877acf0aa/apps/browser/qml/pages/components/FavoriteContextMenu.qml#L48

sledges (changes requested):

apps/browser/qml/pages/components/BookmarkItem.qml
llewelld: Thanks for spotting this. I've removed the duplication and tried to make things a bit clearer.

llewelld:
I've reverted back to using the original (new) string, but also still now have the duplication removed. Thanks for your help @sledges. The new string is the following:
```
//% "Edit bookmark"
"description": qsTrId("sailfish_browser-he-edit-bookmark")
```

sledges (approved):
:+1:  from L10n

llewelld:
Rebased.

PR #862, opened 2021-06-10
[browser] Show favicon in history list. Contributes to JB#51735 OMP#JOLLA-166

chriadam:


chriadam:
Still WIP.

apps/browser/qml/pages/components/HistoryItem.qml
```diff
@@ -67,5 +76,5 @@ ListItem {
     }
 
     ListView.onRemove: animateRemoval()
-    onClicked: view.load(model.url, model.title)
+    onClicked: { console.log("XXXXXXXXXXx loading: " + model.url); view.load(model.url, model.title) }
```
chriadam, line 79: not sure if it's something that I broke, or whether it's broken on devel currently, but attempting to open a history item from the popup menu (rather than the url autocompletion list) currently fails.

apps/core/faviconmanager.cpp
```diff
@@ -193,12 +195,14 @@ void FaviconManager::grabIcon(const QString &type, DeclarativeWebPage *webPage,
         QObject::disconnect(*dataConn);
         if (dataFetcher->hasAcceptedTouchIcon()) {
             qCDebug(lcFavoritesLog) << "Storing favicon for" << type;
+qWarning() << "XXXXXXXXXXXXXXX got favicon result: " << dataFetcher->data() << " for url: " << webPage->url().toString() << " with type: " << type;
```
chriadam, line 198: interestingly, I never see this occurring.  In my testing (e.g. loading google.com / jolla.com / twitter.com / amazon.com they never return a favicon result here... which seems strange?

apps/core/faviconmanager.cpp
```diff
@@ -193,12 +195,14 @@ void FaviconManager::grabIcon(const QString &type, DeclarativeWebPage *webPage,
...
             add(type, webPage->url().toString(), dataFetcher->data(), true);
         }
         else {
             std::shared_ptr<QMetaObject::Connection> thumbConn = std::make_shared<QMetaObject::Connection>();
             *thumbConn = connect(webPage, &DeclarativeWebPage::thumbnailResult, [this, type, webPage, thumbConn](const QString &data) {
                 qCDebug(lcFavoritesLog) << "Storing thumbnail for" << type;
+qWarning() << "XXXXXXXXXXXXXXX got thumbnail result: " << data << " for url: " << webPage->url().toString() << " with type: " << type;
```
chriadam, line 205: instead, the result is always a thumbnail, which happens to be some ... completely nondescript square from somewhere in the page, rather than something useful.  Weird.

apps/shared/WebView.qml
```diff
@@ -211,6 +211,14 @@ WebContainer {
                 if (!modelUrl || modelUrl != url) {
                     tabModel.updateThumbnailPath(tabId, "")
                 }
+
+                // Update the favicon for history items.
+                if (modelUrl && modelUrl != url) {
+console.log("XXXXXXXXX url changed to: " + url + ", grabbing favicon..")
+                    FaviconManager.grabIcon("history", webPage,
```
chriadam, line 218: will this be persisted across reboots etc?  hopefully so?

apps/browser/qml/pages/components/HistoryItem.qml
chriadam: works on latest devel, so I must have broken it somehow.  curious.

apps/core/faviconmanager.cpp
llewelld: I get the same results as you. I did some really brief tests, moving your `grabIcon()` call from `WebView.onUrlChanged` and into `WebView.onLoadedChanged`. I had more luck in this case (see this debug output: [gecko.log](https://github.com/sailfishos/sailfish-browser/files/6629460/gecko.log)).

I'm not at all sure that `onLoadedChanged` is the correct place to put it, but I guess the important point is that the favicon url is read from the html file, so it can only be safely used after at least the main html file has been loaded by the browser.

apps/core/faviconmanager.cpp
llewelld: That's the default fallback in case there is no favicon. I think it takes a square from the top left corner of the page. It would be good to improve on that.

apps/core/faviconmanager.cpp
llewelld: If there's a way to solve this inside FaviconManager, rather than relying on `grabIcon` being called in the correct place, that would be really nice.

apps/shared/WebView.qml
llewelld: Yes, it should be. The favicons are saved to file each time a new one is added. There's scope for a lot of optimisation here, which will be even more relevant now this is being used for history and not just logins and favicons (since the latter are changed quite rarely, whereas history changes frequently).

apps/core/faviconmanager.cpp
chriadam: Thanks, I tried doing it with:

```
    if (!webPage->domContentLoaded()) {
        QObject *contextObj = new QObject(this);
        connect(webPage, &DeclarativeWebPage::domContentLoadedChanged,
                contextObj, [this, type, webPage, size, contextObj] {
                    this->grabIcon(type, webPage, size);
                    contextObj->deleteLater();
                }, Qt::UniqueConnection);
        return;
    }
```

and then calling grabIcon() in the onUrlChanged handler, but it didn't work (it appears to trigger too "soon" and we get some strange results like the favicon for the PREVIOUS site being stored for the next browse location).

For now, I think triggering grabIcon() from within `onLoadedChanged: if (loaded) ... ` makes the most sense.

apps/shared/WebView.qml
chriadam: Do you think there's value in adding a check inside grabIcon() which immediately returns if the type="history" and a favicon for that site has already been loaded, to avoid unnecessary favicon fetch+store?

apps/core/faviconmanager.cpp
chriadam: True.  I mean, I don't have any better suggestions (maybe choose the most common non-white/black colour from the page?  but that might be expensive to determine...)

apps/shared/WebView.qml
llewelld: I was pondering this too. The downside is that if, at some point in the future, a site without a favicon gets one, the icon won't be updated; or if there's a snapshot instead of a favicon, the icon won't update if the site changes it's design. But, personally, I think the positives of doing this outweigh the negatives.

I wouldn't restrict the behaviour to history items either. I don't see why it shouldn't apply in all cases (currently just this and the login icons).

apps/core/faviconmanager.cpp
llewelld: Yeah, it's tricky. I was thinking maybe grabbing a larger portion of the page and scaling it down. I'm not sure I'd trust myself to write a more complex heuristic based on page contents, but if you think something like that can work, I'm all for it.

apps/core/faviconmanager.cpp
llewelld: Thanks for trying. I agree, if there's no easy way to make it work universally, then better to implement something bespoke that will work fine for the history.

apps/core/faviconmanager.cpp
chriadam: A scaled down version would probably be far better, if it doesn't get totally mushed due to the low res.  Anyway, it's unrelated to this PR I guess.

apps/shared/WebView.qml
chriadam: I've now added this via a `if (!get(type, webPage->url().toString()).isEmpty()) return;` check in grabIcon()

chriadam:
No longer WIP.

apps/history/declarativehistorymodel.cpp
```diff
@@ -28,6 +31,7 @@ QHash<int, QByteArray> DeclarativeHistoryModel::roleNames() const
     roles[UrlRole] = "url";
     roles[TitleRole] = "title";
     roles[DateRole] = "date";
+    roles[FaviconRole] = "favicon";
     return roles;
 }
```
llewelld, line 37: Please excuse the odd comment placement (gitlab only lets you put comments near lines that have changed).

It would probably make sense to add a call to `FaviconManager::clear("history")` in the `DeclarativeHistoryModel::clear()` method, so that the icons get cleared from disc when the history is cleared.

apps/history/declarativehistorymodel.cpp
```diff
@@ -100,6 +104,8 @@ QVariant DeclarativeHistoryModel::data(const QModelIndex & index, int role) cons
         return url.title().isEmpty() ? url.url() : url.title();
```
llewelld, line 104: Similarly, sorry for the odd placement.

Would there be scope to also remove an individual icon when the history link is removed? It adds complexity, because the icons are shared between all the history items for the same domain, so some check would be needed to ensure there are no others still in the list. If you go down this route it might also make sense to make `FaviconManager::sanitiseHostname()` accessible for the comparison.

apps/core/faviconmanager.cpp
```diff
@@ -186,6 +188,10 @@ QString FaviconManager::get(const QString &type, const QString &hostname)
 
 void FaviconManager::grabIcon(const QString &type, DeclarativeWebPage *webPage, const QSize &size)
 {
+    if (!get(type, webPage->url().toString()).isEmpty()) {
+        return; // favicon was previously already loaded.
+    }
```
llewelld, line 193: A nice addition; I do think this makes sense in general.

apps/core/faviconmanager.h
```diff
@@ -14,8 +14,7 @@
 #include <QObject>
 #include <QMap>
 
-#include "declarativewebpage.h"
-
+class DeclarativeWebPage;
```
llewelld, line 17: Also a nice spot.

apps/history/declarativehistorymodel.cpp
```diff
@@ -100,6 +104,8 @@ QVariant DeclarativeHistoryModel::data(const QModelIndex & index, int role) cons
...
     case DateRole:
         return url.date();
+    case FaviconRole:
+        return FaviconManager::instance()->get("history", url.url());
```
llewelld, line 108: Should this be a `QStringLiteral("history")`? I always seem to misjudge the correct optimisation with this.

llewelld (changes requested):
This looks good and works nicely. My only real suggestion would be to clear the icons when the history is cleared (if for no other reason than privacy).

apps/history/declarativehistorymodel.cpp
chriadam: done

apps/history/declarativehistorymodel.cpp
chriadam: done, as sanitizedHostname()

apps/history/declarativehistorymodel.cpp
chriadam: done

apps/history/declarativehistorymodel.cpp
```diff
@@ -57,17 +62,26 @@ void DeclarativeHistoryModel::remove(int index)
...
-            return;
+        } else if (FaviconManager::sanitizedHostname(link.url()) == sanitizedUrl) {
+            // this site is still represented in history, so cannot clear icon
+            canClearFavicon = false;
         }
         index++;
     }
 
     // Not in model but remove from database anyway
     DBManager::instance()->removeHistoryEntry(url);
```
llewelld, line 79: In most cases `DBManager::instance()->removeHistoryEntry(url)` will now be called twice for the same entry (once here, and once in the call  to `remove(index)` on line 70. Does this matter?

Just a very small point, but the comment no longer makes sense, so probably better to remove it to avoid confusion.

apps/history/declarativehistorymodel.cpp
```diff
@@ -38,6 +42,7 @@ void DeclarativeHistoryModel::clear()
     m_links.clear();
     endResetModel();
     DBManager::instance()->clearHistory();
+    FaviconManager::instance()->clear(QStringLiteral("history"));
```
llewelld, line 45: Following testing I notice that `DeclarativeHistoryModel::clear()` [doesn't actually get called](https://github.com/sailfishos/sailfish-browser/blob/84fffac3ec4fe1e4ddbfbb1babee4b3b561ae32a/apps/core/settingmanager.cpp#L104) when the history is cleared.

For passwords the icons are [cleared directly](https://github.com/sailfishos/sailfish-browser/blob/84fffac3ec4fe1e4ddbfbb1babee4b3b561ae32a/apps/core/settingmanager.cpp#L125) in the `SettingsManager`, which might be one way to address this. But if something more complex is needed this could be lodged as a separate bug.

llewelld (approved):
I added a couple of comments, but this is already working nicely and I'd be happy for it to go in as it is already, so approving.

atatarov (approved):

apps/history/declarativehistorymodel.cpp
chriadam: I assumed the point of this was to catch the case where the url wasn't in the model (for some reason) but was in the database.

apps/history/declarativehistorymodel.cpp
chriadam: Interesting, nice catch.. I have created JB#54848 / OMP#JOLLA-230 to track this.  I guess that given that there are different data types (passwords, history items, favicons) and both in-memory and in-database copies of the data, there might be some complexities / side effects which need to be considered..

apps/history/declarativehistorymodel.cpp
llewelld: Yes, that was my assumption too. I was just being over picky here, because now this line will get called even if the entry *is* in the model. But still the code comment makes sense anyway to clarify things, so please ignore my suggestion.

apps/history/declarativehistorymodel.cpp
llewelld: Thanks for looking at this and for creating the bug.

chriadam:
Rebased, merging now.

PR #863, opened 2021-06-10
[browser] Avoid duplicating strings across subprojects. Fixes JB#54618

llewelld:
There are QML files shared by both the browser and captiveportal subprojects. If both projects run them through the translator there will be duplicate IDs, which unnecessarily increases work for the translators.

This change ensures the shared files are only parsed once.

### DIFF
--- a/apps/browser/browser.pro
+++ b/apps/browser/browser.pro
@@ -20,7 +20,10 @@ packagesExist(qdeclarative5-boostable) {
 }
 
 # Translations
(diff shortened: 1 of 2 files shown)


llewelld:
After this change, the resulting files **sailfish-browser_eng_en.qm** and **sailfish-captiveportal_eng_en.qm** respectively contain the following string ids. I don't see duplicates, and hopefully none are missing.

linguist-browser

linguist-captiveportal

atatarov (approved):

PR #866, opened 2021-06-17
[sailfish-browser] Verify web page URLs saved to the app grid have the http or https scheme. Fixes JB#54647 OMP#JOLLA-202

adenexter:


apps/browser/qml/pages/components/BookmarkEditDialog.qml
```diff
@@ -97,6 +96,10 @@ UserPromptDialog {
                 EnterKey.enabled: canAccept
                 EnterKey.iconSource: "image://theme/icon-m-enter-accept"
                 EnterKey.onClicked: accept()
+
```
rainemak, line 99: Let's have a custom error for the "acceptableInput" false. I.e. 

```
                description: {
                    if (errorHighlight) {
                        if (!acceptableInput) {
                            //% "URL scheme (%1) missing"
                            return qsTrId("sailfish_browser-la-missing_url_scheme").arg(scheme)
                        }
                        //% "URL is required"
                        return qsTrId("sailfish_browser-la-url_editor_error")
                    }
                    
                    return ""
                }

```

adenexter: Done.

apps/browser/qml/pages/components/BookmarkEditDialog.qml
```diff
@@ -83,20 +83,34 @@ UserPromptDialog {
...
-                //% "URL is required"
-                description: errorHighlight ? qsTrId("sailfish_browser-la-url_editor_error") : ""
+                description: {
+                    if (!errorHighlight) {
+                        return ""
+                    } else if (text.length > 0) {
+                        var scheme = root.url.match(/^(https?):\/\/.+/)
+
+                        //% "URL scheme (%1) missing"
+                        return qsTrId("sailfish_browser-la-missing_url_scheme").arg(scheme.length > 1 ? scheme[1] : "https")
```
llewelld, line 99: This bit is confusing me a bit, since `scheme[1]` will only ever match `http` or `https`, and the line won't be shown in these cases. So won't this line only ever say "URL scheme (https) missing"?

Did I miss a case?

apps/browser/qml/pages/components/BookmarkEditDialog.qml
```diff
@@ -83,20 +83,34 @@ UserPromptDialog {
                 id: urlField
                 text: url
 
-                acceptableInput: text.length > 0
```
llewelld, line 86: Currently if there's an error (i.e. no http:// or https:// prefix) I can still accept the dialogue and create the app icon. I think it'd make sense to block accept in this case (`acceptableInput` doesn't seem to be tied to the validator by default for me).

llewelld (changes requested):
At the moment I can still accept the page if there's an error, which I think is probably unintentional.

apps/browser/qml/pages/components/BookmarkEditDialog.qml
adenexter: Unless the binding has been overridden TextField.acceptableInput == TextInput.acceptableInput which is internally determined from https://doc.qt.io/qt-5/qml-qtquick-textinput.html#acceptableInput-prop, and the errorHighlight is derived from acceptableInput so if you're seeing the error message then the value is being updated based on the validator.

It's possible you were previously able to accept with a empty URL or the acceptableInput wasn't used in that determination.

apps/browser/qml/pages/components/BookmarkEditDialog.qml
adenexter: root.url is the unedited url supplied when the dialog is shown and not the user edited one that may have had the scheme stripped from it resulting in this error. 

If there's a case when the dialog is shown without a source URL then it will always show https.

pvuorela:
Just an option, but could alternatively automatically add the scheme, either one, if it's missing from user input. Wouldn't need an error state and string.

apps/browser/qml/pages/components/BookmarkEditDialog.qml
llewelld: Okay, thanks for clarifying, that makes sense.

adenexter:
Which scheme though?

apps/browser/qml/pages/components/BookmarkEditDialog.qml
adenexter: The value TextField.acceptableInput is changing but Dialog.canAccept isn't beyond evaluating to true during construction.  I can't see any assignments to canAccept inside the qml file, so I guessing something external is breaking the binding.

llewelld: In practice what I'm finding is that the dialog can be accepted even with an error showing. See here (Save button isn't dimmed, selecting it still creates the app shortcut, but the shortcut doesn't work):
acceptableInput

pvuorela:
Either :) Could argue for both, http is the traditional choice though these days could already assume everything is https.

apps/browser/qml/pages/components/BookmarkEditDialog.qml
llewelld, line 99: If `root.url` doesn't match the regex (e.g. attempting to bookmark a local file), `scheme` will be null which breaks the `scheme.length` check, so I guess there should be a null check in here too.

apps/browser/qml/pages/components/BookmarkEditDialog.qml
adenexter: AddToAppGridDialog overwrote canAccept. Fixed.

apps/browser/qml/pages/components/BookmarkEditDialog.qml
adenexter: Fixed.

llewelld:
Sorry for being awkward on something that looks like it should be an easy fix, but it seems there's a bit of a tangled web here. This page is also used for adding bookmarks. In the case of bookmarks the restriction to http/https should be relaxed I think, since while app grid icons don't work for anything other than http/https, the same restriction doesn't apply to bookmarks.

adenexter:
What works for bookmarks but not the app grid? file:// I suppose but if it's for a html file then that will work for the app grid too so it's debatable but that could be added to the generally accepted set too. And if there are others then maybe they should be added to the browser's .desktop?

llewelld:
There are various transports supported by gecko (file, data, about, chrome, view-source, ftp,...) some of which someone might plausibly bookmark. I'm not sure we can reasonably capture them all in the desktop file (if we can, then I'd have no objection), but it would be a shame to restrict them as favourites.

rainemak:
> Either :) Could argue for both, http is the traditional choice though these days could already assume everything is https.

Please do not automatically add the scheme

rainemak:
> There are various schemes supported by gecko (file, data, about, chrome, view-source, ftp,...) some of which someone might plausibly bookmark. I'm not sure we can reasonably capture them all in the desktop file (if we can, then I'd have no objection), but it would be a shame to restrict them as favourites.

If a bookmark is something else than http or https => when it shall be invalid from bookmark point of view.

apps/browser/qml/pages/components/BookmarkEditDialog.qml
llewelld: Thanks; working nicely now.

llewelld (approved):
Following @rainemak's [comment](https://github.com/sailfishos/sailfish-browser/pull/866#issuecomment-863907839) if we're only accepting http(s) for favourites then this is all working nicely from what I can tell, so approving.

rainemak (approved):

aroditelev (approved):

okodron (approved):

PR #873, opened 2021-06-23
[sailfish-browser] Add a drag to close gesture to the popup menu. Fixes JB#54770 OMP#JOLLA-208

adenexter:


adenexter:
This is the slightly clever option where a continuous drag will collapse the menu and keep on going to dismiss the menu. It's seamless but that lack of separate action to go from scrolling to closing might make accidentally closing a little too easy. The DragOverBounds boundsDirection works on both ends too so you can now overscroll the menu at the bottom, it should be possible to disguise that if not stop it but my limited efforts so far have resulted in binding loops.

The other option is to put the entire menu inside a MouseArea and drag using that when the Flickable is at its limits. In which case scrolling would stop when you reach the limits of the flickable and you would have to start a new gesture to close the menu.

adenexter:
I've suppressed the overscroll at the bottom and increased the close threshold so the menu has to be dragged a little beyond the bottom of the screen before it kicks in which reduces the instances of closing on short flicks.

aroditelev (approved):

rainemak (approved):
Please rebase before merging. Approving already.

PR #875, opened 2021-06-23
[user-agent] User agent override for vk.com. Fixes JB#54784

paveltkv:


okodron (approved):

rainemak (approved):
Didn't test on device as I don't have vk account but assume that case you were fixing works now. Hence, approving.

PR #877, opened 2021-06-24
[browser] Remove spacing between header and list view content in folder picker pages. Contributes to JB#54754

chriadam:


chriadam:
This one is no longer required, as the default header spacing has now been set to 0 in the sailfish-components-pickers PR

PR #879, opened 2021-06-25
[sailfish-browser] Redesign SettingsPage and PrivacySettingsPage. Contributes to JB#54798

Maves1:
Move "Do Not Track" TextSwitch to Settings Page. Add section headers and redesign "privacy" button

### DIFF
--- a/apps/browser/qml/pages/SettingsPage.qml
+++ b/apps/browser/qml/pages/SettingsPage.qml
@@ -14,7 +14,6 @@ import Sailfish.Silica 1.0
 import Sailfish.Browser 1.0
 import org.nemomobile.configuration 1.0
 import com.jolla.settings.system 1.0
-import Sailfish.Policy 1.0
 import Sailfish.WebEngine 1.0
 import Sailfish.Pickers 1.0
 import "components"
@@ -27,7 +26,7 @@ Page {
     function removeProtocolTypeFromUri(uri) {
         if (uri.length === 0)
             return uri
-       return uri.replace(/(^\w+:|^)\/\//, '')
+        return uri.replace(/(^\w+:|^)\/\//, '')
     }
 
     SilicaFlickable {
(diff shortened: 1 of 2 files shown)


okodron:
Please, change the commit message to `Contributes to JB#54798`

atatarov (approved):

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -27,7 +27,7 @@ Page {
     function removeProtocolTypeFromUri(uri) {
         if (uri.length === 0)
             return uri
-       return uri.replace(/(^\w+:|^)\/\//, '')
+        return uri.replace(/(^\w+:|^)\/\//, '')
     }
 
     SilicaFlickable {
```
llewelld, line 33: I guess it was hidden before because the page contents was shorter than the screen, but it's common practice to add some space to the bottom of the content so that it doesn't touch the bottom of the screen when scrolled. I.e. add something like this into the column around line 40:
```
        bottomPadding: Theme.paddingLarge
```

llewelld (commented):
This seems like a nice, clean change. Since several of the items have already lost the restriction of `AccessPolicy.browserEnabled`, now might be a good opportunity to remove references to it from the file completely.

Apart from this, and the addition of some padding at the bottom of the page, this looks good to me.

apps/browser/qml/pages/SettingsPage.qml
Maves1: Fixed :)
Added the padding and removed all references to AccessPolicy.browserEnabled

okodron: Should you drop `import Sailfish.Policy 1.0` as well?

llewelld: All of these look addressed now, so I'll resolve this conversation for clarity.

llewelld (approved):
There was some discussion about whether there should be an entire section just for the "Downloads Save destination". Leaving that to one side, all of the changes here look good to me. Thanks @Maves1 also for cleaning up the AccessPolicy code.

llewelld:
Needs rebasing before merging.

Maves1:
> Needs rebasing before merging.

Done

PR #880, opened 2021-06-25
[user-agent] User agent override for videomost.com to join conferences directly in the browser. Fixes JB#54802

paveltkv:


okodron:
Now it should work in every case.

okodron (approved):

mkenttala (approved):

rainemak (approved):
Didn't test on device but assuming that "like Chrome" produces expected results.

PR #881, opened 2021-06-28
[sailfish-browser] Use sandboxed booster. JB#54788 OMP#JOLLA-213

rainemak:
Instantiate browser booster on user session startup.

Modify exec lines in desktop and D-Bus autostart files so that
browser booster is used for launching.

Remove booster privilege configuration that is no longer needed.

Signed-off-by: Simo Piiroinen <simo.piiroinen@jolla.com>

### DIFF
--- a/data/70-browser.conf
+++ b/data/70-browser.conf
@@ -1,1 +1,1 @@
-BROWSER="invoker -s --type=generic /usr/bin/sailjail -p sailfish-browser.desktop /usr/bin/sailfish-browser"
+BROWSER="/usr/bin/invoker --type=browser,silica-qt5 -A -- /usr/bin/sailfish-browser"
--- a/50-sailfish-browser.conf
+++ b/50-sailfish-browser.conf
@@ -1,0 +1,2 @@
+[Unit]
+Wants=booster-browser@sailfish-browser.service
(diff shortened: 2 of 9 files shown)


Tomin1 (approved):

data/70-browser.conf
```diff
@@ -1 +1 @@
-BROWSER="invoker -s --type=generic /usr/bin/sailjail -p sailfish-browser.desktop /usr/bin/sailfish-browser"
+BROWSER="/usr/bin/invoker --type=browser,silica-qt5 -A -- /usr/bin/sailfish-browser"
```
izh1979, line 1: Why do we need to specify 2 booster types here?

dseight (approved):

data/70-browser.conf
rainemak: We don't. It should be enough like "--type=browser".

okodron (approved):

data/70-browser.conf
izh1979: I mean, either we need to use it everywhere or nowhere. Because, for example, we have also media booster. Are we using fallback there too?

Tomin1: We are using the same fallback also on camera which is using media booster.

rainemak:
Rebase

PR #882, opened 2021-06-28
[sailfish-browser] Add ability to open home page on first launch. Fixes JB#54801

aroditelev:


apps/browser/qml/pages/BrowserPage.qml
```diff
@@ -374,8 +375,9 @@ Page {
             // In case first use not done show the overlay immediately.
             if (url == "") {
                 bringToForeground(webView.chromeWindow)
-                if (!WebUtils.firstUseDone) {
-                    overlay.enterNewTabUrl(PageStackAction.Immediate)
+            // In case first use not done and the home page is set.
```
okodron, line 378: Indention
```suggestion
               // In case the first use is not done but the home page is set
```

apps/browser/qml/pages/BrowserPage.qml
```diff
@@ -389,6 +391,8 @@ Page {
             if (webView.tabModel.activateTab(url)) {
                 webView.releaseActiveTabOwnership()
             } else {
+            // In case first use not done and default home page is set and the site is opened by specifying a parameter to application.
```
okodron, line 394: Indention as well

aroditelev: Thanks!)

apps/browser/qml/pages/BrowserPage.qml
aroditelev: Fixed

apps/browser/qml/pages/BrowserPage.qml
```diff
@@ -374,8 +375,9 @@ Page {
             // In case first use not done show the overlay immediately.
             if (url == "") {
                 bringToForeground(webView.chromeWindow)
-                if (!WebUtils.firstUseDone) {
-                    overlay.enterNewTabUrl(PageStackAction.Immediate)
+                // In case the first use is not done but the home page is set.
+                if(!WebUtils.firstUseDone && WebUtils.homePage !== "about:blank" && WebUtils.homePage.length > 0) {
```
atatarov, line 379:
```suggestion
                if (!WebUtils.firstUseDone && WebUtils.homePage !== "about:blank" && WebUtils.homePage.length > 0) {
```

atatarov:
I hope this PR doesn't depend on https://github.com/sailfishos/sailfish-browser/pull/883

apps/browser/qml/pages/BrowserPage.qml
aroditelev: done

aroditelev:
> I hope this PR doesn't depend on #883

This pr won't solve the problem  #882

llewelld:
My understanding is that the original behaviour (on first launch, open the Start View, even if a home page has been configured) is intentional, so that the user is presented with a set of options on first launch. If this is the case then the original bug should be marked as invalid.

I'll explain this on the bug and we can have some further discussion about it there. My suggestion would be to leave this PR open until that discussion has concluded.

atatarov (approved):

rainemak:
As agreed. Closing as invalid.

PR #883, opened 2021-06-29
[browser] Fix initial currentIndex in ComboBox for choosing home page. Fixes JB#54820

Maves1:
Fix initially highlighted item in ComboBox for choosing home page

### DIFF
--- a/apps/browser/qml/pages/SettingsPage.qml
+++ b/apps/browser/qml/pages/SettingsPage.qml
@@ -48,14 +48,22 @@ Page {
             ComboBox {
                 id: homePage
 
+                readonly property bool _homePageBlank: homePageConfig.value === "about:blank"
+
                 width: parent.width
                 //: Label for home page text field
                 //% "Home Page"
                 label: qsTrId("settings_browser-la-home_page")
                 //% "Default"
-                value: homePageConfig.value === "about:blank" ? qsTrId("sailfish_browser-la-home_page_default") : removeProtocolTypeFromUri(homePageConfig.value)
+                value: _homePageBlank ? qsTrId("sailfish_browser-la-home_page_default") : removeProtocolTypeFromUri(homePageConfig.value)
                 leftMargin: Theme.horizontalPageMargin + homePageIcon.width + Theme.paddingMedium
                 contentHeight: Theme.itemSizeMedium
+
+                currentIndex: _homePageBlank
+                              ? 0 // For default home page (blank)
+                              : 1 // For custom home page
+
+                on_HomePageBlankChanged: currentIndex = _homePageBlank ? 0 : 1
 
                 Icon {
                     id: homePageIcon
@@ -72,9 +80,18 @@ Page {
                     }
                     MenuItem {
                         readonly property string site: removeProtocolTypeFromUri(homePageConfig.value)
-                        //: Instead of %1 site address will be displayed
-                        //% "Custom website %1"
-                        property string title: qsTrId("sailfish_browser-me-home_page_custom").arg(homePageConfig.value === "about:blank" ? "" : site)
+
+                        property string title: {
+                            if (homePage._homePageBlank || site.trim() === "") {
+                                //: Shown when site is empty
+                                //% "Custom website"
+                                return qsTrId("sailfish_browser-me-home_page_custom_empty")
+                            } else {
+                                //: Instead of %1 site address will be displayed
+                                //% "Custom website %1"
+                                qsTrId("sailfish_browser-me-home_page_custom").arg(site)
+                            }
+                        }
 
                         textFormat: Text.StyledText
                         color: highlighted ? Theme.highlightColor : Theme.primaryColor


atatarov (approved):

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -23,11 +23,12 @@ Page {
     id: page
 
     readonly property int _textSwitchIconCenter: Math.round((permissionIcon.width - Theme.itemSizeExtraSmall) / 2.0)
+    readonly property bool homePageBlank: homePageConfig.value === "about:blank"
```
llewelld, line 26: Since this is an internal property, it might make sense to also start with an underscore: `_homePageBlank`.

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -58,10 +59,14 @@ Page {
...
                 label: qsTrId("settings_browser-la-home_page")
                 //% "Default"
-                value: homePageConfig.value === "about:blank" ? qsTrId("sailfish_browser-la-home_page_default") : removeProtocolTypeFromUri(homePageConfig.value)
+                value: homePageBlank ? qsTrId("sailfish_browser-la-home_page_default") : removeProtocolTypeFromUri(homePageConfig.value)
                 leftMargin: Theme.horizontalPageMargin + homePageIcon.width + Theme.paddingMedium
                 contentHeight: Theme.itemSizeMedium
 
+                currentIndex: homePageBlank
+                              ? 0 // For default home page (blank)
+                              : 1 // For custom home page
```
llewelld, line 68: This works for the initial creation of the page, but after that the binding gets broken (because `currentIndex` is assigned to by the `ComboBox`). You could perhaps include a line like this, so that the index also gets updated if the value changes in the background:
```
onHomePageBlankChanged: currentIndex = homePageBlank ? 0 : 1
```
In practice this is a minor point as it's unlikely to happen in practice, but maybe it's nice to retain the behaviour.

llewelld (commented):
I've made some very minor comments. In practice this is working nicely already.

apps/browser/qml/pages/SettingsPage.qml
llewelld: I should have mentioned -- just in case it's helpful -- that I tested this with these:
```
dconf write /apps/sailfish-browser/settings/home_page '"https://example"'
dconf write /apps/sailfish-browser/settings/home_page '"about:blank'
```

atatarov: Thank you! Why the "homePageBlank" property has no binding with "currentIndex"?

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -79,7 +84,7 @@ Page {
                         readonly property string site: removeProtocolTypeFromUri(homePageConfig.value)
                         //: Instead of %1 site address will be displayed
                         //% "Custom website %1"
-                        property string title: qsTrId("sailfish_browser-me-home_page_custom").arg(homePageConfig.value === "about:blank" ? "" : site)
+                        property string title: qsTrId("sailfish_browser-me-home_page_custom").arg(homePageBlank ? "" : site)
```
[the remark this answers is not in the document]
atatarov: It's nice catch, thank you! I think the empty string as an argument will create an error like "Missing args"

apps/browser/qml/pages/SettingsPage.qml
llewelld: I've not checked in detail, so I could be wrong abut this, but I think it's because there's this line in `ComboBoxController.qml` which sets the index when the current item is changed:
```
    currentIndex = menuItemIndex
```
This assignment breaks the binding that you set on line 66.

atatarov: Done

llewelld: Working great now, thanks.

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -72,9 +80,18 @@ Page {
...
-                        property string title: qsTrId("sailfish_browser-me-home_page_custom").arg(homePageConfig.value === "about:blank" ? "" : site)
+
+                        property string title: {
+                            if (homePage._homePageBlank || site.trim() === "") {
+                                //: Shown when site is empty
+                                //% "Custom website"
+                                return qsTrId("sailfish_browser-me-home_page_custom_empty")
+                            } else {
+                                //: Instead of %1 site address will be displayed
+                                //% "Custom website %1"
```
llewelld, line 91: There was some discussion about changing the wording for these (taken from the notes):
 1. "Default home page" can be renamed to the "Start view"
 2. "Custom website" change "Web page" also update the actual dialog

But I think it was decided that these changes would be handled separately, so I'm just noting it here for reference.

llewelld (approved):
This is working really nicely as far as I can tell :+1:

apps/browser/qml/pages/SettingsPage.qml
okodron: Yes, let's create a separate MR for this.

PR #890, opened 2021-07-06
[sailfish-browser] Add to history list: call context menu, open in incognito tab, save to bookmarks. Fixes JB#54882

aroditelev:
Added a fix for opening context menu of the history list.
Also added menu items (open in incognito tab, add to bookmarks) in accordance with the design:
image

It should be noted that when saving to bookmarks, the standard icon is always used.

### DIFF
--- a/apps/browser/qml/pages/HistoryPage.qml
+++ b/apps/browser/qml/pages/HistoryPage.qml
@@ -1,6 +1,6 @@
 /****************************************************************************
 **
-** Copyright (c) 2020 Open Mobile Platform LLC.
+** Copyright (c) 2020 - 2021 Open Mobile Platform LLC.
 **
 ****************************************************************************/
 
@@ -20,7 +20,8 @@ Page {
     property var remorse
     readonly property bool pendingRemorse: remorse ? remorse.pending : false
 
-    signal loadPage(string url)
+    signal loadPage(string url, bool newTab)
+    signal saveBookmark(string url, string title)
 
     HistoryList {
         id: view
@@ -32,8 +33,10 @@ Page {
         onLoad: {
             view.focus = true
             pageStack.pop()
-            root.loadPage(url)
+            root.loadPage(url, newTab)
         }
+
+        onSaveBookmark: root.saveBookmark(url, title)
 
         Component.onCompleted: model.search("")
 
--- a/apps/browser/qml/pages/components/HistoryItem.qml
+++ b/apps/browser/qml/pages/components/HistoryItem.qml
@@ -26,8 +26,53 @@ ListItem {
 
     width: parent.width
     contentHeight: Math.max(Theme.itemSizeMedium, column.height + 2*Theme.paddingMedium)
+    menu: Component {
+        ContextMenu {
+            MenuItem {
+                //% "Open in normal mode tab"
+                text: webView.privateMode ? qsTrId("sailfish_browser-me-open-in-normal-mode-tab")
+                                          : //% "Open in private tab"
+                                            qsTrId("sailfish_browser-me-open-in-private-tab")
+                onClicked: view.load(model.url, model.title, true)
+            }
+            MenuItem {
+                //: Share link from browser history pulley menu
+                //% "Share"
+                text: qsTrId("sailfish_browser-me-share-link")
+                onClicked: webShareAction.shareLink(model.url, model.title)
+            }
+            MenuItem {
+                //% "Copy to clipboard"
+                text: qsTrId("sailfish_browser-me-copy-to-clipboard")
+                onClicked: {
+                    Clipboard.text = model.url
+                    //% "Link address copied"
+                    notification.text = qsTrId("sailfish_browser-me-link_address_copied")
+                    notification.show()
+                }
+            }
+            MenuItem {
+                //% "Add to bookmarks"
+                text: qsTrId("sailfish_browser-me-add-to-bookmarks")
+                onClicked: view.saveBookmark(model.url, model.title)
+            }
+            MenuItem {
+                //: Delete history entry
+                //% "Delete"
+                text: qsTrId("sailfish_browser-me-delete")
+                onClicked: historyDelegate.remove(model.url)
+            }
+        }
+    }
 
     ListView.onAdd: AddAnimation { target: root }
+
+    Notice {
+        id: notification
+
+        duration: Notice.Short
+        verticalOffset: -Theme.itemSizeMedium
+    }
 
     Row {
         id: row
@@ -67,5 +112,5 @@ ListItem {
     }
 
     ListView.onRemove: animateRemoval()
-    onClicked: view.load(model.url, model.title)
+    onClicked: view.load(model.url, model.title, false)
 }
--- a/apps/browser/qml/pages/components/HistoryList.qml
+++ b/apps/browser/qml/pages/components/HistoryList.qml
@@ -20,42 +20,20 @@ SilicaListView {
     property bool showDeleteButton
     property bool menuClosed
 
-    signal load(string url, string title)
+    signal load(string url, string title, bool newTab)
+    signal saveBookmark(string url, string title)
+
+    onLoad: if (newTab) webView.privateMode = !webView.privateMode
 
     // To prevent model to steal focus
     currentIndex: -1
 
     delegate: HistoryItem {
         id: historyDelegate
-        menu: contextMenuComponent
+
         search: view.search
         showDeleteButton: view.showDeleteButton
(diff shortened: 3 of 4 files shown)


apps/browser/qml/pages/components/HistoryContextMenu.qml
```diff
@@ -0,0 +1,45 @@
...
+        MenuItem {
+            //: Share link from browser history pulley menu
+            //% "Share"
+            text: qsTrId("sailfish_browser-me-share-link")
+            onClicked: webShareAction.shareLink(model.url, model.title)
+        }
+        MenuItem {
+            //% "Copy to clipboard"
+            text: qsTrId("sailfish_browser-me-copy-to-clipboard")
+            onClicked: Clipboard.text = model.url
```
llewelld, line 31: It would be good to add a simple notification to indicate the text has been copied. Something like [this](https://github.com/sailfishos/sailfish-browser/blob/fba7e57fd375e391d099f2b5179e52cc67440d35/apps/browser/qml/pages/LoginsPage.qml#L36).

apps/browser/qml/pages/components/HistoryList.qml
```diff
@@ -32,29 +34,8 @@ SilicaListView {
         showDeleteButton: view.showDeleteButton
         onMenuOpenChanged: view.menuClosed = !menuOpen
 
-        Component {
+        HistoryContextMenu {
```
llewelld, line 37: Could this be added directly inside `HistoryItem.qml`? It doesn't look like `HistoryItem` will ever need to be used without the context menu.

apps/browser/qml/pages/components/HistoryContextMenu.qml
```diff
@@ -0,0 +1,45 @@
...
+
+import QtQuick 2.1
+import Sailfish.Silica 1.0
+
+Component {
+    ContextMenu {
+        MenuItem {
+            //% "Open in private tab"
+            text: qsTrId("sailfish_browser-me-open-in-private-tab")
+            onClicked: view.openInPrivateTab(model.url, model.title)
```
llewelld, line 20: When the user is already in Private mode, selecting the item will open it in private mode. In this case, perhaps the menu could have "Open in normal mode" instead, to make it convenient for the user to switch out of private mode?

apps/browser/qml/pages/components/HistoryContextMenu.qml
```diff
@@ -0,0 +1,45 @@
...
+****************************************************************************/
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.1
+import Sailfish.Silica 1.0
+
+Component {
```
llewelld, line 15: What's the reason for wrapping this in a `Component`?

apps/browser/qml/pages/HistoryPage.qml
```diff
@@ -35,6 +37,14 @@ Page {
             root.loadPage(url)
         }
 
+        onOpenInPrivateTab: {
```
llewelld, line 40: Just a suggestion, but rather than duplicate the code, it might be worth checking whether it would be neater to extend `onLoad` with an optional `private` bool parameter (I've not tried it, so it might also end up complicating things: your judgement).

apps/browser/qml/pages/HistoryPage.qml
```diff
@@ -21,6 +21,8 @@ Page {
     readonly property bool pendingRemorse: remorse ? remorse.pending : false
 
     signal loadPage(string url)
+    signal saveBookmark(string url, string title)
+    signal openInPrivateTab(string url, string title)
```
llewelld, line 25: Header needs updating with "2021".

llewelld (commented):
A nice addition and nice implementation. I've made a few suggestions.

apps/browser/qml/pages/components/HistoryContextMenu.qml
aroditelev: Calling the context menu does not work without a wrapper:  ```[W] unknown:92 - file:///usr/lib/qt5/qml/Sailfish/Silica/private/ViewItem.qml:92: TypeError: Cannot call method 'connect' of undefined```

apps/browser/qml/pages/HistoryPage.qml
aroditelev: ```onLoad``` if only you add the functionality of opening a new tab to it ...

apps/browser/qml/pages/components/HistoryContextMenu.qml
llewelld: Thanks for clarifying. That's peculiar though as I don't get that behaviour here: if I remove the wrapper it works without any errors.

llewelld: Just to clarify, I was only curious, it's not a problem as long as there's a reason for it.

apps/browser/qml/pages/HistoryPage.qml
llewelld: I was thinking that the logic about whether to open a new tab might also be simplified if it were pushed up into Overlay.loadPage().

apps/browser/qml/pages/components/HistoryContextMenu.qml
aroditelev: If you remove the Component, then nothing will change, but if you use this code directly in `HistoryList.qml`, you will get the error above.

aroditelev: But the error is very strange, before everything worked in ` HistoryList.qml`

apps/browser/qml/pages/HistoryPage.qml
aroditelev: done

apps/browser/qml/pages/components/HistoryContextMenu.qml
aroditelev: Done

apps/browser/qml/pages/components/HistoryList.qml
aroditelev: Done

apps/browser/qml/pages/components/HistoryContextMenu.qml
aroditelev: Done

apps/browser/qml/pages/HistoryPage.qml
aroditelev: Done

apps/browser/qml/pages/components/HistoryItem.qml
```diff
@@ -26,9 +26,54 @@ ListItem {
...
+    Notice {
+        id: notification
+
+        duration: Notice.Short
+        verticalOffset: -Theme.itemSizeMedium
+    }
+
     Row {
         id: row
         x: Theme.horizontalPageMargin
```
llewelld, line 79: On line 115, you'll need to add an extra `false` parameter:
```
    onClicked: view.load(model.url, model.title, false)
```
to avoid this error:
```
[W] unknown:115 - file:///usr/share/sailfish-browser/pages/components/HistoryItem.qml:115: Error: Insufficient arguments
```
Although you can skip parameters in javascript, apparently you can't for signals.

apps/browser/qml/pages/components/HistoryList.qml
```diff
@@ -20,42 +20,20 @@ SilicaListView {
     property bool showDeleteButton
     property bool menuClosed
 
-    signal load(string url, string title)
+    signal load(string url, string title, bool newTab)
+    signal saveBookmark(string url, string title)
+
+    onLoad: webView.privateMode = !webView.privateMode
```
llewelld, line 26: You'll need a bit of logic here to avoid switching modes when the user presses on a history item (rather than selecting "new tab" from the menu).

llewelld (changes requested):
This is looking great, thank you. There are just a couple of small fixes needed.

apps/browser/qml/pages/components/HistoryList.qml
aroditelev: Thanks, did not notice), added a condition.

apps/browser/qml/pages/components/HistoryItem.qml
aroditelev: Done

llewelld (approved):
This all looks good to me. Thank you.

atatarov (approved):

PR #892, opened 2021-07-08
Activate multiple available locales. JB#54830 OMP#JOLLA-222

llewelld:
If the "intl.locale.requested" preference is unset, gecko will default
to only a single fallback locale. If it's set to an empty string, it
will instead set the locale based on the OS setting.

This adds the preference as an empty string to the default preferences.


okodron (approved):

rainemak (approved):

llewelld:
Rebased.

PR #893, opened 2021-07-09
[browser] Update contextMenu terminology in settings page for "home page". Contributes to JB#54904

atatarov:
Rename "default home page" to "start view", "custom website" to "web page" in contextMenu.
Remove trim() method from title of start view combobox menu, it had already trimmed when
sets to config value.

### DIFF
--- a/apps/browser/qml/pages/SettingsPage.qml
+++ b/apps/browser/qml/pages/SettingsPage.qml
@@ -54,14 +54,14 @@ Page {
                 //: Label for home page text field
                 //% "Home Page"
                 label: qsTrId("settings_browser-la-home_page")
-                //% "Default"
-                value: _homePageBlank ? qsTrId("sailfish_browser-la-home_page_default") : removeProtocolTypeFromUri(homePageConfig.value)
+                //% "Start view"
+                value: _homePageBlank ? qsTrId("sailfish_browser-la-start_view") : removeProtocolTypeFromUri(homePageConfig.value)
                 leftMargin: Theme.horizontalPageMargin + homePageIcon.width + Theme.paddingMedium
                 contentHeight: Theme.itemSizeMedium
 
                 currentIndex: _homePageBlank
-                              ? 0 // For default home page (blank)
-                              : 1 // For custom home page
+                              ? 0 // For start view (blank)
+                              : 1 // For web page
 
                 on_HomePageBlankChanged: currentIndex = _homePageBlank ? 0 : 1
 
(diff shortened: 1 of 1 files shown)


apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -51,17 +51,17 @@ Page {
...
                 width: parent.width
-                //: Label for home page text field
-                //% "Home Page"
-                label: qsTrId("settings_browser-la-home_page")
-                //% "Default"
-                value: _homePageBlank ? qsTrId("sailfish_browser-la-home_page_default") : removeProtocolTypeFromUri(homePageConfig.value)
+                //: Label for start view text field
+                //% "Start view"
+                label: qsTrId("settings_browser-la-start_view")
+                //% "Start view"
```
okodron, line 57: Not sure whether "Home page" should be renamed exactly to "Start view" which has 2 possible values: "Start view" and "Web page".

rainemak: Yeap, I think this should stay as "Home page" but I'd rather not use "page" word with "start view" that is the qml thingie. Reason being that it easily confuses people when there are "page" when referring to web pages and same for "views".

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -74,22 +74,22 @@ Page {
 
                 menu: ContextMenu {
                     MenuItem {
-                        //% "Default home page"
-                        text: qsTrId("sailfish_browser-me-home_page_default")
+                        //% "Start view"
+                        text: qsTrId("sailfish_browser-me-start_view")
```
rainemak, line 78: As summary just rename these two menu items

okodron (approved):

apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -60,8 +60,8 @@ Page {
                 contentHeight: Theme.itemSizeMedium
```
llewelld, line 60: For me it would be clearer if `Default` on line 57 (the text shown in the ComboBox selection when Start View is selected) were changed to `Start view`, so that the text matches the option selected by the user.

There may be some disagreement over this; I tried to figure out the intention from the earlier review discussion but this point remains ambiguous for me.

llewelld (commented):
All of the proposed changes are good and match what I would expect, but there is one additional change which I expected to see, as I've noted.

apps/browser/qml/pages/SettingsPage.qml
llewelld: Thanks for making this change, it looks nicer to me.

llewelld (approved):
This looks good to me, and makes the changes that I understood from the task.

There may be some remaining changes needed, but maybe we can merge these in already and make any further changes later if they're considered necessary?

PR #894, opened 2021-07-13
[browser] Remove favicons when clearing history. Contributes to JB#54864 OMP#JOLLA-237

chriadam:
SettingManager::clearHistory() calls
DBManager::clearHistory() calls
DBWorker::clearHistory() emits historyAvailable() which triggers
DeclarativeHistoryModel::historyAvailable() which calls
DeclarativeHistoryModel::updateModel().

Thus, DeclarativeHistoryModel::clear() is never invoked in this
codepath.  We cannot simply clear the favicon for any url which is
no longer represented in the history model within updateModel()
since historyAvailable() will be emitted with limited data (due to
the LIMIT in the query in the general case, or due to the WHERE
in the dynamic filter case).  As such, we need to ensure that we
manually clear the history favicons when DBManager::clearHistory()
is called.

### DIFF
--- a/apps/storage/dbmanager.cpp
+++ b/apps/storage/dbmanager.cpp
@@ -13,6 +13,7 @@
 #include <QMetaObject>
 
 #include "dbworker.h"
+#include "faviconmanager.h"
 
 static DBManager *gDbManager = 0;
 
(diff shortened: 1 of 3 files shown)


llewelld (approved):
Tested and these changes work nicely. Good that you checked out the details so thoroughly.

okodron (approved):

PR #896, opened 2021-07-21
Switch from NetworkManager to ConnectionHelper. JB#54831 OMP#JOLLA-223

llewelld:
Introduces the ConnectionHelper to allow the network selector to be
triggered when the "network-enable" notification is sent.

NetworkManager was previously used for determining the  online status.
ConnectionHelper exposes the same functionality (it's a very light
wrapper around NetworkManager in fact), so ConnectionHelper is now used
for this instead.

### DIFF
--- a/apps/shared/ResourceController.qml
+++ b/apps/shared/ResourceController.qml
@@ -17,6 +17,7 @@ import Sailfish.WebEngine 1.0
 import Nemo.DBus 2.0
 import MeeGo.Connman 0.2
 import org.nemomobile.policy 1.0
+import Nemo.Connectivity 1.0
 
 // QtObject cannot have children
 Item {
@@ -83,7 +84,10 @@ Item {
     Connections {
         target: WebEngine
 
-        onInitialized: networkManager.notifyOfflineStatus()
+        onInitialized: {
+          WebEngine.addObserver("network-enable")
+          connectionHelper.notifyOfflineStatus()
+        }
         onRecvObserve: {
             if (message === "media-decoder-info") {
                 if (data.state === "meta") {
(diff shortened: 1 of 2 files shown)


apps/shared/ResourceController.qml
```diff
@@ -95,19 +99,20 @@ Item {
                     _lastStateOwner = data.owner
                 }
                 calculateStatus()
+            } else if (message === "network-enable") {
```
chriadam, line 102: if this message is asynchronous, is it worth adding a check for `&& !connectionHelper.online` ?  or maybe attemptToConnectNetwork() already has such a check, internally?  From a quick check, it looks like it does have one for `m_netman->defaultRoute()->state() == "online"` but still does some signal emission etc stuff... I guess it's fine as-is but thought I'd raise it for discussion.

llewelld:
Rebased.

apps/shared/ResourceController.qml
llewelld: Thanks, it's a good point. As you mention, `ConnectionHelper` does do its own check with the fallback being that the connection dialog doesn't open (which is what we want here).

However, correct behaviour will rely on the two checks being equivalent. From digging through the `ConnectionHelper` code it's not 100% clear to me that `NetworkManager->defaultRoute()->state()` will always give exactly the same result as the `NetworkManager::state` property, but I think they must be pretty close, given that updating `NetworkManager::state` will synchronously cause the default route to be updated.

My preference would be to avoid complicating the code here and leave the check to `ConnectionHelper`. But I agree there's a risk, so open to changing it still.

atatarov (approved):

chriadam (approved):
LGTM, works nicely.

PR #898, opened 2021-08-10
[sailfish-browser] Don't add opensearches when using private mode. JB#55094 OMP#JOLLA-271

mkenttala:


rainemak (approved):
Looks good! Didn't test on device but this should work.

okodron (approved):

PR #899, opened 2021-08-19
[browser] Fix accessing the security code prompt from within a sandbox. JB#55160 OMP#JOLLA-307

adenexter:
Access the device lock Authenticator interface on the system bus instead of through the peer-to-peer connection used by the device lock QML plugin.


apps/core/secureaction.cpp
```diff
@@ -0,0 +1,254 @@
...
+public:
+    ServiceWatcher()
+        : QDBusServiceWatcher(deviceLockService, QDBusConnection::systemBus())
+    {
+        connect(this, &QDBusServiceWatcher::serviceOwnerChanged,
+                this, [this](const QString &, const QString &oldOwner, const QString &newOwner) {
+            if (newOwner.isEmpty() && m_available) {
+                m_available = false;
+                emit availableChanged(false);
+            } else if (oldOwner.isEmpty() && !m_available) {
```
chriadam, line 73: Maybe this can never happen in practice, but it looks like if oldOwner and newOwner are both empty and m_available is false, we will enter this branch and set m_available to true here, even though newOwner is still empty.

apps/core/secureaction.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+QString replyPath()
+{
+    static int counter = 0;
+
+    return QStringLiteral("/org/nemomobile/devicelock/client/sailfish_browser/SecureAction%1").arg(++counter);
+}
+
+enum class Method {
+    NoAuthentication    = 0x000,
```
chriadam, line 52: OOI is it intentionally 0x000 rather than 0x0000?

apps/core/secureaction.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+        m_authenticating = true;
+
+        m_response.reset(new QDBusPendingCallWatcher(systemBus.asyncCall(createAuthenticatorCall(
+                    QStringLiteral("RequestPermission"),
+                    { QVariant::fromValue(QDBusObjectPath(m_replyPath)), m_message, QVariantMap(), uint(Method::AllAvailable) }))));
+
+        connect(m_response.data(), &QDBusPendingCallWatcher::finished, this, [this](QDBusPendingCallWatcher *watcher) {
+            watcher->deleteLater();
+            m_response.take();
```
chriadam, line 215: I assume this is the same pointer as watcher?

apps/core/secureaction.h
```diff
@@ -0,0 +1,84 @@
...
+
+private:
+    SecureAction * const m_secureAction;
+};
+
+class SecureAction : public QObject
+{
+    Q_OBJECT
+    Q_PROPERTY(bool available READ isAvailable NOTIFY availableChanged)
+    Q_PROPERTY(QString message READ message WRITE setMessage NOTIFY messageChanged)
```
chriadam, line 47: to be API compat should a requireAuthentication property be added, also?

apps/core/secureaction.cpp
```diff
@@ -0,0 +1,254 @@
...
+            m_registered = systemBus.registerObject(m_replyPath, this);
+
+            if (!m_registered) {
+                qmlInfo(this) << "DBus adaptor registration error " << systemBus.lastError().name() << " " << systemBus.lastError().message();
+                return;
+            }
+        }
+
+        m_authenticating = true;
+
```
chriadam, line 208: What ensures that SecureActionAuthenticatorAdaptor::PermissionGranted() will be invoked by the deviceLockService as a result of this call?  Is there some connection to a dbus signal which I'm missing?

chriadam: Or is it just documented that the deviceLockService will invoke that method on the replyPath?  If so, maybe a comment could be added above to state this.

apps/core/secureaction.cpp
adenexter: It can't happen, if the old and new owners are the same the owner hasn't changed.

apps/core/secureaction.cpp
adenexter: Not intentional. It will be same in nemo-qml-plugin-devicelock.

apps/core/secureaction.cpp
adenexter: Yes

apps/core/secureaction.h
adenexter: Not needed, it's internal API and require authentication was a just a authentication hasn't been asked for yet flag. It starts true and is cleared when authentication is granted.

apps/core/secureaction.cpp
adenexter: Yes, the service will send a reply to the reply path object.

adenexter: Added a comment.

Tomin1 (approved):
LGTM

atatarov (approved):

PR #902, opened 2021-08-31
[browser] Do not reset model upon a tab close. Fixes JB#55345 OMP#JOLLA-333

rainemak:


mkenttala (approved):

apps/browser/qml/pages/components/TabGridView.qml
```diff
@@ -123,14 +123,4 @@ SilicaGridView {
     }
 
     onLoadedChanged: positionViewAtIndex(model.activeTabIndex, GridView.Center)
-
-    Connections {
```
chriadam, line 127: Why was this done in the past?  Why is it no longer required?  Neither the bug nor the commit message have any useful information about this..

rainemak: We missed this in the review when changes were landed to the tab model. This causes horrible flickering in the tab model when you have multiple tabs open and you close one as the model resets when a tab is closed. If there are errors like model data not being updated correctly, those should be fixed instead.

rainemak:
Amended commit message and bug description.

chriadam:
Thanks, LGTM

chriadam (approved):

okodron (approved):

PR #905, opened 2021-09-04
RFC: Bring back the close button in the toolbar

deepbluev7:
Closing annoying popups has become really hard with 4.2.0.19, because the close button wasn't clickable anymore. This happened for 3 reasons, the Animator sets the Overlay to disabled, the Flickable of the PopUpMenu covers the button and the InvertedMouseArea also stole the input from it. 0c03d773f2dd06a79612c45f0e9533950a39e41a fixes the Flickable covering the button, so we only needed to remove the `enabled: false` property change on the animator and make the InvertedMouseArea not steal inputs from it. That way the PopUpMenu still closes, when the close-tab button is pressed, but it also closes the tab, which imo is the right behaviour. The back button is now also clickable and closes the menu as well (and navigates back).

<del>
Because I thought the close button not working was an intentional change at first and the button was supposed to close the Menu now, I moved the share button to the menu and instead added a "close menu" button to the PopUpMenu. Since clicking outside and sliding down are the only ways to close the menu on next, this might be unnecessary.

There are a few small issues with this change though:

- The close button was later removed. I reverted that commit in this PR
- Background dimming was added. This makes the close button dimmed as well. Not sure if that is an issue and if it is, how it should be resolved.
- The menu doesn't render for me at all if I install this on 4.2.0.19, if I apply this patch on next. I think that is not my fault though, just that the shader the Background uses is maybe incompatible with 4.2.0.19? It works fine if I apply it on top of the 2.1.13 tag used in 4.2.0.19, but there the button is still covered by the Flickable and it sounds like useless duplication of effort to fix that there, if it is already fixed in next. EDIT: It seems like the shader does not work on my Xperia 10 II? Uncommenting the shader fixes it, but the menu is obviously scaled wrong in that case.
- If the Share change is acceptable, no idea if I did the localization stuff properly. I should probably find an existing share text instead and reuse its id?
</del>

I've since changed the approach to just add a close and back button to the menu footer as well as a close menu icon instead of the share button to prevent me from refreshing the page by accident. See https://github.com/sailfishos/sailfish-browser/pull/905#issuecomment-986977387 for more details.

For more discussion around the browser changes, see https://forum.sailfishos.org/t/browser-redesign-in-sailfish-4-2-feedback-thread/7867

I would appreciate some feedback on this change and if that would be an acceptable direction. It would also be nice to know, if I can somehow resolve the rendering issues so that I can properly test this. Thank you!

### DIFF
--- a/apps/browser/qml/pages/components/PopUpMenuFooter.qml
+++ b/apps/browser/qml/pages/components/PopUpMenuFooter.qml
@@ -27,6 +27,20 @@ Rectangle {
         height: root.height
 
         Shared.IconButton {
+            id: closeTabButton
+            width: Theme.itemSizeLarge
+            icon.source: "image://theme/icon-m-tab-close"
+            icon.opacity: enabled ? 1.0 : Theme.opacityLow
+            enabled: webView.tabModel.count > 0
+            onTapped: {
+                webView.tabModel.closeActiveTab()
+                if (webView.tabModel.count === 0) {
+                    overlay.startPage(PageStackAction.Animated)
+                }
+            }
+        }
+
+        Shared.IconButton {
             id: forwardButton
             width: Theme.itemSizeLarge
             icon.source: "image://theme/icon-m-forward"
@@ -34,18 +48,6 @@ Rectangle {
             enabled: webView.canGoForward
             onTapped: {
                 webView.goForward()
-                overlay.animator.showChrome()
-            }
-        }
-
-        Shared.IconButton {
-            id: shareButton
-            width: Theme.itemSizeLarge
-            icon.source: "image://theme/icon-m-share"
-            icon.opacity: enabled ? 1.0 : Theme.opacityLow
-            enabled: webView.contentItem
-            onTapped: {
-                overlay.toolBar.shareActivePage()
                 overlay.animator.showChrome()
             }
         }
--- a/apps/browser/qml/pages/components/Overlay.qml
+++ b/apps/browser/qml/pages/components/Overlay.qml
@@ -504,14 +504,6 @@ Shared.Background {
                 }
                 onShowChrome: overlayAnimator.showChrome()
 
-                onCloseActiveTab: {
-                    // Activates (loads) the tab next to the currect active.
-                    webView.tabModel.closeActiveTab()
-                    if (webView.tabModel.count === 0) {
-                        overlay.startPage(PageStackAction.Animated)
-                    }
-                }
-
                 onLoadPage: overlay.loadPage(url)
                 onEnterNewTabUrl: overlay.enterNewTabUrl()
                 onFindInPage: {
--- a/apps/browser/qml/pages/components/PopUpMenuItem.qml
+++ b/apps/browser/qml/pages/components/PopUpMenuItem.qml
@@ -94,6 +94,22 @@ Item {
 
             OverlayListItem {
                 height: Theme.itemSizeSmall
+                enabled: webView.contentItem
+                opacity: enabled ? 1.0 : 0.5
+                iconWidth: root.iconWidth
+                horizontalOffset: root.horizontalOffset
+                iconSource: "image://theme/icon-m-share"
+                //% "Share"
+                text: qsTrId("sailfish_browser-la-share")
+
+                onClicked: {
+                    overlay.toolBar.shareActivePage()
+                    overlay.animator.showChrome()
+                }
+            }
+
+            OverlayListItem {
+                height: Theme.itemSizeSmall
                 enabled: !DownloadManager.pdfPrinting
                 opacity: enabled ? 1.0 : 0.5
                 iconWidth: root.iconWidth
--- a/apps/browser/qml/pages/components/ToolBar.qml
+++ b/apps/browser/qml/pages/components/ToolBar.qml
@@ -57,7 +57,6 @@ Column {
     signal showSecondaryTools
     signal showInfoOverlay
     signal showChrome
-    signal closeActiveTab
     signal showCertDetail
 
     // Used from the PopUpMenu


deepbluev7:
Thank you for taking a look at this!

1. The problem is that not being able to close new pages, that open without the users consent or by mistake, take a lot more effort to close now. This is really annoying in day to day usage (and multiple people complained about it already). Imo this doesn't even look that much out of place, if one removes the dimming effect. It looks like the button is just part of an expanded bottom row. You basically have Close, Back, Forward, Refresh, Close Menu available to you in that case. The only other option I would see is adding the close button to the overlay on the right, but then it is a bit too easy to hit (I already refresh the page quite a lot by accident). Adding it to the overlay permanently would overload the overlay too much and leave too little place for the 3 other buttons, TLS indicator and URL.
2. Thanks, I will do that.
3. The rendering issues basically make the menu transparent. I think that is related to the shader being used having issues with the Xperia 10 II on Sailfish OS 4.2.0.19. This might be fixed in a newer build already.
4. I'm not sure I am following you. Are you suggesting to split up the code into a PR that moves the share menu, which then would get merged, but I should keep the close button as a local patch?

attah:
Any updates here?

I agree the menu state should not influence other actions at all, and presumably not having the MouseArea steal clicks would have them register on the webpage too, which is bad (pressing anywhere outside the menu should just close it and do nothing more). But the problem is that not being able to close tabs from when you are viewing them is such a major PITA it really outweighs all of the other improvements.

How about longpress and swipe (like for changing keyboard languages or picking accented letters) on the tab icon, and have "close tab" there?

deepbluev7:
I was a bit busy with university the last few weeks, so I couldn't address the feedback yet, but I just had the idea that you probably could just make the popup wider and place the buttons below it. You can still close the popup by swiping down on it, so that should be fine for one handed use and it would give us space for the back and close buttons. Longpress and swipe doesn't quite feel right, I would expect that to switch tabs or go back/forward.

attah:
Well, if there is going to be any space outside the menu (which i do think looks nice), i really think it should do exactly nothing if you press it. But swiping down should of course also work (but does not mean tapping outside is free for other uses). Widening the menu by one icon could be the solution indeed!

Sure, previous/next tab would be a more logical action on longpress+swipe, but any possibility to close is infinitely better than none... and having the entry-point on the tab button could be considered more logical than it being on the menu.

deepbluev7:
I just tried it, it does exactly nothing if you press it. (I just wasn't sure anymore, so I didn't want to claim it). Clicking outside just closes the menu. Clicking outside on the close button, closes the tab and the menu.

deepbluev7:
Okay, I changed a few things in this patch now:

- Share menu is still moved to the menu
- There is still a close menu button
- I've added a "close current tab" button and a back button to the menu footer

As a result the close button doesn't look weird anymore, it is a proper part of the menu now.Also the options further up are easier to reach with your left hand thumb now (since the menu is wider and you can't easily reach across with your thumb the further up a menu item is). You have 3 options to close the menu now. Either by tapping the same location you opened it with, dragging it down or tapping outside of it. The menu button might not be needed then, but I have accidentally refreshed the page a lot without it. Branch is based on the 4.4.0 branch now, so I can't actually build this, but I've applied the patch manually to my browser on my system.


Demo

deepbluev7:
Someone noticed I accidentally deleted the option to favourite a page, so I brought that back and removed the menu close button, since you can just swipe down:

Screenshot

attah:
Just installed it, and am now beating myself up for not doing so earlier.
This is brilliant!

rainemak:
This PR should alleviate / easy a bit: https://github.com/sailfishos/sailfish-browser/pull/976

The way how we organize & show tabs in browsing page doesn't help as we don't have means to sneak peek to tabs and quickly show them and switch between them. That said, I think that there are cases when this would beneficial as well. I'll discuss and check this with design. My apologies from my delayed response.

apps/browser/qml/pages/components/PopUpMenuFooter.qml
```diff
@@ -27,40 +27,40 @@ Rectangle {
...
 
         Shared.IconButton {
-            id: shareButton
+            id: backwardButton
             width: Theme.itemSizeLarge
-            icon.source: "image://theme/icon-m-share"
+            icon.source: "image://theme/icon-m-back"
             icon.opacity: enabled ? 1.0 : Theme.opacityLow
-            enabled: webView.contentItem
+            enabled: webView.canGoBack
```
rainemak, line 48: Back button should not be here. It's already there in the toolbar and visible if you can go back.

deepbluev7: So the idea behind that is, if you ever want to do advanced things with you tab, like close, go forward, etc, I often combine that with going backwards, reloading, etc (some pages are really annoying). So imo it is the most consistent to have all controls you need for that available in one menu section. Otherwise I need to remember to close the menu, when I hit forward one time too many and need to go back again. In my personal opinion the UX is best, if you have close, backward, forward and reload in one row.

rainemak: That advanced use is very much speculative for boarder audience. From UX point of this once again increases the menu width is not good as it reduces area that can be used for closing the menu.

Checked this with design we concluded following order for the bottom part of the menu:
"Close" / "Forward" /  "Favorite" / "Reload"

"Share" can be moved to be a menu item.

deepbluev7: Well, you can just swipe down the menu, but I can remove the back button, if it really is an issue. It does have the sideeffect of the menu being easier to use with your left thumb though.

rainemak: Yeap, it was raised please drop. If you're on top of this, could also rebase the PR. Thank you in advance!

deepbluev7: Done

rainemak: Thank you! Almost there. Please update also icon order to the menu:

"Close" / "Forward" / "Favorite" / "Reload"

Just swap "Favorite" and "Reload" items.

See: https://github.com/sailfishos/sailfish-browser/pull/905#discussion_r785692823

deepbluev7: Oops, I missed that, my bad!

rainemak:
@deepbluev7 thank you from your contribution! Works quite nicely. I'll merge this one.

One thing that I'm thinking. After tapping close button from menu, I'd expect that menu would be close. That's a matter of taste as some may like close multiple tabs but usually new focused item should be available right after expected action. With this you need to first close the menu to reach content by either tapping outside or dragging it closed. This is adds extra step. In case I'd like to contribute that as well, I'm all green. We can gather a bit of feedback from this behaviour before we conclude this.

That said, nice get you in board!

deepbluev7:
@rainemak Thank you for merging this! For me not closing the menu was actually intentional, because I found it confusing, if the menu closed, but the old behaviour can be restored by restoring the changes in Overlay.qml, if it turns out to be bad.

rainemak:
Like said, @deepbluev7 thank you for this one.

rainemak (commented):
LGTM

PR #906, opened 2021-09-06
[sailfish-browser] Do not call CloseEventFilter::cancelStopApplication. Contributes to JB#54813

paveltkv:
CloseEventFilter::cancelStopApplication stops the browser process from
closing, but does not restore the browser window. This makes it impossible
to close current process and start a new instance of the browser without
the device reboot.


paveltkv:
Another solution has been merged: https://github.com/sailfishos/sailfish-browser/pull/897

PR #908, opened 2021-09-13
[browser] Fix active tab display when opening tabs. Fixes JB#54794 OMP#JOLLA-214

xlise:


adenexter:
Ideally you'd change the highlightRangeMode so the GridView would scroll itself to keep the current item in view, but I can't find a combination of parameters that doesn't also change the scroll limits.

adenexter (approved):

aroditelev (approved):

rainemak:
@xlise could you rebase this one?

xlise:
> @xlise could you rebase this one?

rebased

xlise:
> Ideally you'd change the highlightRangeMode so the GridView would scroll itself to keep the current item in view, but I can't find a combination of parameters that doesn't also change the scroll limits.

Thanks for the advice

PR #909, opened 2021-09-16
[browser] Use combobox also for save destination. Fixes JB#55580

atatarov:
The BrowserListItem isn't good for this case, because this should be highlighted by default for currentIndex.
So, it's good to use ComboBox instead and move general code to BrowserComboBox for avoiding duplicated code.
Set save destination by default if the user reject to choose download folder.

### DIFF
--- a/apps/browser/qml/pages/SettingsPage.qml
+++ b/apps/browser/qml/pages/SettingsPage.qml
@@ -45,32 +45,24 @@ Page {
                 title: qsTrId("sailfish_browser-he-settings")
             }
 
-            ComboBox {
+            BrowserComboBox {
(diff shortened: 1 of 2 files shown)


aroditelev (approved):

rainemak:
@atatarov could you rebase this one?

rainemak:
Add BrowserComboBox for what? Would be more informative to say on the title that "[browser] Use combobox also save destination. Fixes JB#55580"

In to body would could elaborate that actual implementation more so that it becomes clear that you want to avoid duplicated code.

rainemak (approved):

PR #910, opened 2021-09-23
Use connectivity to establish online status. Fixes JB#55628 OMP#JOLLA-401

llewelld:
This partially reverts changes made in 8014ee097530c4c902. In that
change, ConnectionHelper was introduced to allow the "Select Wifi
network" dialog to be triggered. In order to simplify the code, the
online status check was also switched to using ConnectionHelper.

It turns out this interferes with the captive portal, since this relies
on the "ready" state to be considered equivalent to "online". The
ConnectionHelper didn't respect this.

This change therefore switches back the online status check to using the
previous NetworkManager approach.

Setting as draft until I've tested the change.

### DIFF
--- a/apps/shared/ResourceController.qml
+++ b/apps/shared/ResourceController.qml
@@ -1,7 +1,7 @@
 /****************************************************************************
 **
 ** Copyright (c) 2013 - 2019 Jolla Ltd.
-** Copyright (c) 2019 Open Mobile Platform LLC.
+** Copyright (c) 2019 - 2021 Open Mobile Platform LLC.
 ** Contact: Dmitry Rozhkov <dmitry.rozhkov@jolla.com>
 ** Contact: Raine Makelainen <raine.makelainen@jollamobile.com>
 **
@@ -107,17 +107,18 @@ Item {
 
     ConnectionHelper {
         id: connectionHelper
+        readonly property bool connected: status >= ConnectionHelper.Connected
 
         function notifyOfflineStatus() {
             if (WebEngine.initialized) {
                 WebEngine.notifyObservers("embed-network-link-status",
                                           {
-                                              "offline": !connectionHelper.online
+                                              "offline": !connectionHelper.connected
                                           })
             }
         }
 
-        onOnlineChanged: notifyOfflineStatus()
+        onConnectedChanged: notifyOfflineStatus()
     }
 
     DBusInterface {


llewelld:
I've tested the "error page" functionality and that still works as expected after this change, so I'm switching it from draft. The captive portal capability still needs to be checked though.

mkenttala (approved):

okodron (commented):
If you revert back to `NetworkManager` usage then `ConnectionHelper` seems to be not necessary here.
Or you can try to check its `state` for `online` and `ready`. However `ConnectionHelper` provides states as enums instead of strings.

llewelld:
> Or you can try to check its `state` for `online` and `ready`. However `ConnectionHelper` provides states as enums instead of strings.

That's a nice observation indeed; thank you. We need `ConnectionHelper` to open the dialogue, but we could be using the `status` enum provided there. I'll update the PR.

llewelld:
I've updated the PR so now it's all done using the `ConnectionHelper`.

apps/shared/ResourceController.qml
```diff
@@ -107,17 +107,19 @@ Item {
 
     ConnectionHelper {
         id: connectionHelper
+        readonly property bool connected: status === ConnectionHelper.Online
+                                          || status === ConnectionHelper.Connected
```
rainemak, line 111: Needless to say but "status >= ConnectionHelper.Connected" would do the same. I think this is better still.

rainemak:
Please update PR title as well.

rainemak (approved):
Approving already. Looks right, checked nemo-qml-plugin-connectivity as well.

https://github.com/sailfishos/nemo-qml-plugin-connectivity/blob/master/src/nemo-connectivity/connectionhelper.cpp#L67

apps/shared/ResourceController.qml
llewelld: Thanks; indeed. Updated.

apps/shared/ResourceController.qml
```diff
@@ -107,17 +107,18 @@ Item {
...
     ConnectionHelper {
         id: connectionHelper
+        readonly property bool connected: status >= ConnectionHelper.Connected
 
         function notifyOfflineStatus() {
             if (WebEngine.initialized) {
                 WebEngine.notifyObservers("embed-network-link-status",
                                           {
-                                              "offline": !connectionHelper.online
+                                              "offline": !connectionHelper.connected
```
okodron, line 116: Should you also take `connectionHelper.online` in account?

okodron (approved):

llewelld:
Rebased.

PR #912, opened 2021-09-24
[browser] Make SettingsManager refactoring. JB#55660

atatarov:


okodron (approved):

apps/browser/qml/pages/PrivacySettingsPage.qml
```diff
@@ -76,6 +76,12 @@ Page {
                     text: qsTrId("settings_browser-la-clear_bookmarks")
                     checked: true
                 }
+                TextSwitch {
```
aroditelev, line 79: need an empty string for consistency

apps/core/settingmanager.cpp
```diff
@@ -31,6 +31,7 @@ SettingManager::SettingManager(QObject *parent)
     m_clearCookiesConfItem = new MGConfItem("/apps/sailfish-browser/actions/clear_cookies", this);
     m_clearPasswordsConfItem = new MGConfItem("/apps/sailfish-browser/actions/clear_passwords", this);
     m_clearCacheConfItem = new MGConfItem("/apps/sailfish-browser/actions/clear_cache", this);
+    m_clearSitePermissionsConfItem = new MGConfItem("/apps/sailfish-browser/actions/clear_site_permissions", this);
```
rainemak, line 34: It's conceptually wrong that we keep on adding these confitems. These were here originally here because we had browser settings separated to the settings app. Hence, we needed a persistent mechanism to handle these kind of changes also when browser was not running. Now, however, all settings are part of the browser and accessible only when browser is running implying the there is no need to route setting change via confitem.

Now it would be probably good time create a PR that cleans up SettingsManager and exposes it to QML as singleton with needed invokables so that PrivacySettingsPage.qml could be changed like this (nb BookmarkManager should be exposed as well):

```
                        remorse = Remorse.popupAction(page, qsTrId("settings_browser-la-cleared_private_data"),
                                                 function() {
                                                     if (clearHistory.checked) {
                                                         SettingsManager.clearHistory()
                                                     }
                                                     if (clearCookies.checked) {
                                                         SettingsManager.clearCookies()
                                                     }
                                                     if (clearSavedPasswords.checked) {
                                                         SettingsManager.clearPasswords()
                                                     }
                                                     if (clearCache.checked) {
                                                         SettingsManager.clearCache()
                                                     }
                                                     if (clearBookmarks.checked) {
                                                         BookmarkManager.clearBookmarks()
                                                     }
                                                     if (clearSitePermissions.checked) {
                                                         SettingsManager.clearSitePermissions()
                                                     }
                                                 }
                        );
```

atatarov:
it depend on
https://github.com/sailfishos/qtmozembed/pull/10/files

rainemak:
Conceptually code ok. However, now you are mixing up refactoring and new functionality in one commit message. Could you still split this into two commits. One that does thta refactoring and on top of that "Add remove all permissions to PrivacySettingsPage"

atatarov:
> Conceptually code ok. However, now you are mixing up refactoring and new functionality in one commit message. Could you still split this into two commits. One that does thta refactoring and on top of that "Add remove all permissions to PrivacySettingsPage"

Done

apps/browser/main.cpp
```diff
@@ -54,6 +55,11 @@ static QObject *faviconmanager_factory(QQmlEngine *, QJSEngine *)
 {
     return FaviconManager::instance();
 }
+
+static QObject *bookmarkmanager_factory(QQmlEngine *, QJSEngine *)
+{
+    return BookmarkManager::instance();
```
rainemak, line 61: Fingers cross that this won't blow up durng browser shutdown. Afair Qml engine takes ownership of this but that could well be harmless if engine destroys this as BookmarkManager as only default destructor.

To avoid potential dangling pointer, you could add small fix for the BookmarkManager::instance() and change the singleton to be QPointer.

apps/core/declarativewebcontainer.cpp
```diff
@@ -935,23 +934,6 @@ void DeclarativeWebContainer::initialize()
...
-    if (clearTabs) {
-        m_model->clear();
-    }
-
-    // If data was cleared when initialized and we had tabs in previous
-    // session, reset tab model to unloaded state. DBManager emits
-    // tabsAvailable with empty list when tabs are cleared => tab model
-    // changes back to loaded and the initialize() slot gets called again.
-    if (m_settingManager->initialize() && (oldCount > 0) && clearTabs) {
-        m_model->setUnloaded();
```
rainemak, line 951: This can be dropped now from the model as it's not really in use. There's one test case but that's pretty useless of its own so let's just remove that tst_persistenttabmodel::setUnloaded() test as well.

rainemak (approved):

PR #913, opened 2021-09-27
[sailfish-browser] Fix adding link to bookmarks if opening history page via popup menu. Fixes JB#55694

aroditelev:
Add a call to function of saving bookmarks in the PopUpMenuItem element.
Add use of favicon when saving to bookmarks.

### DIFF
--- a/apps/browser/qml/pages/HistoryPage.qml
+++ b/apps/browser/qml/pages/HistoryPage.qml
@@ -21,7 +21,7 @@ Page {
     readonly property bool pendingRemorse: remorse ? remorse.pending : false
 
     signal loadPage(string url, bool newTab)
-    signal saveBookmark(string url, string title)
+    signal saveBookmark(string url, string title, string favicon)
 
     HistoryList {
         id: view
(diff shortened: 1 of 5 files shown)


atatarov (approved):

okodron (approved):

llewelld (approved):
I was able to reproduce the issue before applying the PR; afterwards the issue was fixed, so this is great. The code looks good to me too. Nice that the icon is also now added as well.

One thing I notice: via the history it's possible to add two identical bookmarks (this is blocked if you go via other routes). It doesn't relate to this PR, but might be worth considering for a fix at some future point.

aroditelev:
> I was able to reproduce the issue before applying the PR; afterwards the issue was fixed, so this is great. The code looks good to me too. Nice that the icon is also now added as well.
> 
> One thing I notice: via the history it's possible to add two identical bookmarks (this is blocked if you go via other routes). It doesn't relate to this PR, but might be worth considering for a fix at some future point.

I agree, you need to think over a mechanism that will notify the user that this bookmark already exists, or provide the user with an editing form. But for this I propose to create a separate task.

PR #937, opened 2021-11-05
Initialise SettingManager at start up. Fixes JB#56200 OMP#JOLLA-507

llewelld:
The initialisation is needed to connect up change signals from the Search Provider DConf value. Hence prior to this change, if the user switched Search Provider, it wasn't applied until the next browser restart. This change fixes this.

This is a small regression from 68b52fef7fd89c12ddfe423 which removed the code initialising the `SettingManager` (for valid reasons, but it left it uninitialised).

### DIFF
--- a/apps/core/declarativewebcontainer.cpp
+++ b/apps/core/declarativewebcontainer.cpp
@@ -21,6 +21,7 @@
 #include "logging.h"
 #include "declarativehistorymodel.h"
 #include "closeeventfilter.h"
+#include "settingmanager.h"
 
 #include <webengine.h>
 #include <QTimerEvent>
--- a/apps/core/declarativewebcontainer.h
+++ b/apps/core/declarativewebcontainer.h
@@ -10,8 +10,6 @@
 #ifndef DECLARATIVEWEBCONTAINER_H
 #define DECLARATIVEWEBCONTAINER_H
 
-#include "settingmanager.h"
-
 #include <qmozcontext.h>
 #include <qmozsecurity.h>
 #include <QtGui/QWindow>
--- a/apps/core/settingmanager.cpp
+++ b/apps/core/settingmanager.cpp
@@ -23,11 +23,12 @@ static SettingManager *gSingleton = 0;
 
 SettingManager::SettingManager(QObject *parent)
     : QObject(parent)
-    , m_initialized(false)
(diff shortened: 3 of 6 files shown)


apps/core/declarativewebcontainer.cpp
```diff
@@ -932,6 +933,8 @@ void DeclarativeWebContainer::initialize()
         return;
     }
 
+    SettingManager::instance()->initialize();
```
rainemak, line 936: I'd assume that this very call could be inside the SettingsManager

- EmbedLiteSearchEngine.js - profile-after-change => profile loaded
- SettingManager::handleObserve handles embed:search "init"
- At line settingsmanager.cpp:168 there call to setSearchEngine I think that could changed initialize()

Surely there must be somewhere SettingManager::instance() call and that we have in the browser.cpp

With this approach we're waiting for instance that tabsmodel is loaded which is unrelated to the initialization of the settingsmanager. What do you think?

llewelld: I agree doing it here isn't ideal, but I also wonder if triggering it on the search message is appropriate if `SettingManager` isn't really about search.

Perhaps we could drop `initialize()` entirely and just add the remaining `connect` into the constructor? There's nothing else happening in the initialize that's needed elsewhere and it streamlines things a bit.

rainemak: Yes I agree, keeping current setSearchEngine call in place in SettingManager::handleObserve + adding connection [1] to the SettingsManager constructor would make this even cleaner.

[1]
```
    connect(m_searchEngineConfItem, &MGConfItem::valueChanged,
            this, &SettingManager::setSearchEngine);
```

llewelld: I've moved the `connect` into the constructor and removed the `initialize()` code. I tested it on first startup (after deleting the browser settings folder), after a reboot (following oneshot initialisation), for the top menu search shortcut, and in the general case. They also seem to work as expected, although I had to fix https://github.com/sailfishos/embedlite-components/pull/69 to get the first-run case to work.

okodron (approved):

rainemak (approved):

llewelld:
Rebased.

PR #939, opened 2021-11-08
Update browser cover on tab changes. Fixes JB#56224 OMP#JOLLA-513

llewelld:
Adds a connection to ensure that the browser cover always gets updated whenever the number of tabs changes. Previously the switch from zero to one tabs (which triggers a cover change) wasn't being picked up, leaving the cover in an incorrect state.

### DIFF
--- a/apps/browser/qml/pages/components/Overlay.qml
+++ b/apps/browser/qml/pages/components/Overlay.qml
@@ -668,12 +668,6 @@ Shared.Background {
                     webView.tabModel.clear()
                     overlay.startPage()
                 }
-
-                Component.onCompleted: {
-                    window.setBrowserCover(webView.tabModel)
-                }
-
-                Component.onDestruction: window.setBrowserCover(webView.tabModel)
             }
         }
     }
--- a/apps/browser/qml/browser.qml
+++ b/apps/browser/qml/browser.qml
@@ -17,7 +17,7 @@ BrowserWindow {
     id: window
 
     function setBrowserCover(model) {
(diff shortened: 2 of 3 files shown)


okodron (approved):

apps/browser/qml/pages/components/Overlay.qml
```diff
@@ -70,6 +70,11 @@ Shared.Background {
         overlayAnimator.showChrome()
     }
 
+    Connections {
+        target: webView.tabModel
+        onCountChanged: window.setBrowserCover(webView.tabModel)
```
rainemak, line 75: I'd add "!WebUtils.firstUseDone" condition to the no tabs cover code path of the setBrowserCover so that when starting browser for the very first time we'd also have no tabs cover visible. You can simulate first time usage by closing all tabs and then removing .local/share/org.sailfishos/browser/.firstUseDone. When/if browser is started for the first with a url, the firstUseDone will be touched in place after toolbar hits the bottom state for the very first time. That may imply that you need to add to the "target: WebUtils" connection (BrowserPage line 374) onFirstUseDoneChanged signal handler as well.

I'd keep ```window.setBrowserCover(webView.tabModel)``` in the BrowserPage.qml Component.onCompleted so that initial status will be updated. Further, there's already tabModel connection in the BrowserPage approximately line 207.

llewelld:
Rebased.

rainemak (approved):
Now this should be good. Thank you for your update / fixes!

PR #946, opened 2021-11-16
Follow screen primary orientation when rotating the web window

jpetrell:


apps/core/declarativewebcontainer.cpp
```diff
@@ -865,7 +865,7 @@ void DeclarativeWebContainer::componentComplete()
 
 void DeclarativeWebContainer::updateContentOrientation(Qt::ScreenOrientation orientation)
 {
-    if (gForceLandscapeToPortrait) {
+    if (gForceLandscapeToPortrait || QGuiApplication::primaryScreen()->primaryOrientation() == Qt::LandscapeOrientation) {
```
inzanity, line 868: Is there some reason not to use screen()->primaryOrientation()?

jpetrell: No reason, did not realise the webcontainer was a window. Fixed. :slightly_smiling_face:

PR #947, opened 2021-11-16
Take primary screen orientation into account in the overlay input mask

jpetrell:


rainemak (approved):
Didn't test on device but change looks good.

PR #953, opened 2021-11-29
Create tab early and update resolved url only after it is resolved

rainemak:
See also: https://github.com/sailfishos/qtmozembed/pull/28

### DIFF
--- a/apps/qtmozembed/declarativewebpage.cpp
+++ b/apps/qtmozembed/declarativewebpage.cpp
@@ -55,7 +55,6 @@ DeclarativeWebPage::DeclarativeWebPage(QObject *parent)
     , m_userHasDraggedWhileLoading(false)
     , m_fullscreen(false)
     , m_forcedChrome(false)
-    , m_initialLoadHasHappened(false)
     , m_tabHistoryReady(false)
     , m_urlReady(false)
     , m_restoredCurrentLinkId(-1)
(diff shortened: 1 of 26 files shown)


apps/qtmozembed/declarativewebpage.cpp
```diff
@@ -137,18 +136,31 @@ void DeclarativeWebPage::setInitialTab(const Tab& tab)
...
+    bool urlReadyChanged = !m_urlReady;
+    if (urlReadyChanged && urlResolved) {
+        m_urlReady = true;
+        DBManager::instance()->getTabHistory(tabId());
+    }
 }
 
 void DeclarativeWebPage::onTabHistoryAvailable(const int& historyTabId, const QList<Link>& links, int currentLinkId)
 {
+    Q_UNUSED(historyTabId)
```
mkenttala, line 161: Unneeded Q_UNUSEDs

mkenttala (approved):

rainemak:
There's something wrong with the incoming urls. Need to check.

apps/qtmozembed/declarativewebpage.cpp
rainemak: Fixed

rainemak:
> There's something wrong with the incoming urls. Need to check.

As tabs can now be created guarded case when browser gets started with an incoming url so that it will be matched to the available tabs as well.

[Diff to previous is here](https://github.com/sailfishos/sailfish-browser/compare/e887b3a87063bf843f5390cad135789461f393c6..b8474d00cc4608235c0519985399e51cff98de3f)

rainemak:
And finally also unit tests are fixed. [Diff is here](https://github.com/sailfishos/sailfish-browser/compare/b8474d00cc4608235c0519985399e51cff98de3f..3c2f9e75d2fa34b388faada6ed43bce9413868cb) to name few major findings that I noticed whilst fixing unit tests

- External urls are now skipped immediately when seen so that those on end up to the tab model
- In the url matching upon activating a tab with url downgrading from https to http is guarded: url schemes must match, input url scheme must be empty or target url scheme is https)
- Requested url is cleared from tab also when requested url matches to resolved

rainemak:
Rebased

paveltkv (approved):

PR #954, opened 2021-11-30
Emulate touch events when mouse events come to the browser window

jpetrell:


apps/core/declarativewebcontainer.cpp
```diff
@@ -69,6 +71,7 @@ DeclarativeWebContainer::DeclarativeWebContainer(QWindow *parent)
     , m_activeTabRendered(false)
     , m_clearSurfaceTask(0)
     , m_closing(false)
+    , m_touchScreenDevice(new QTouchDevice)
```
rainemak, line 74: Newer destroyed? Could this be on stack instead?

jpetrell: Yikes. QTouchDevice is Q_GADGET and not QObject so added manual delete. Postponed the initialization to receiving of first mouse events so should not affect touch-driven browser usage.

rainemak (approved):
LGTM from my point of view. Added @okodron , @paveltkv and @atatarov to the reviewer list as if somebody could check as well.

Remember to rebase before merging.

PR #956, opened 2021-12-03
[sailfish-browser] Fix error reading tab thumbnails from cache. Fixes  JB#56298 OMP#JOLLA-533

adenexter:
QImage (Qt 5.6) writes directly to the target file when saving, so if an
Image item happens to try and read from it at the same file then that
may fail because the file is incomplete.

The concurrent read/write may happen because an existing tab is being
updated, or because saveToFile() was dispatched to a thread and the user
could have switched tabs in the interim leading to the data being
written to the thumbnail of a different tab.

### DIFF
--- a/apps/qtmozembed/declarativewebpage.cpp
+++ b/apps/qtmozembed/declarativewebpage.cpp
@@ -301,9 +301,12 @@ void DeclarativeWebPage::forceChrome(bool forcedChrome)
 
 void DeclarativeWebPage::grabResultReady()
 {
-    if (active()) {
-        QImage image = m_grabResult->image();
-        m_grabWritter.setFuture(QtConcurrent::run(this, &DeclarativeWebPage::saveToFile, image));
+    QImage image = m_grabResult->image();
+    if (!image.isNull() && active()) {
+        m_grabWritter.setFuture(QtConcurrent::run(
+                &DeclarativeWebPage::saveToFile,
+                image,
+                QStringLiteral("%1/tab-%2-thumb.jpg").arg(BrowserPaths::cacheLocation()).arg(tabId())));
     }
     m_grabResult.clear();
 }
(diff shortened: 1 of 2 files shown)


rainemak:
Getting following when entering to tabs view:
```
[W] unknown:139 - file:///usr/share/sailfish-browser/pages/components/TabItem.qml:139:13: QML Image: Cannot open: file:///home/defaultuser/.cache/org.sailfishos/browser/tab-1-thumb.jpg
[W] unknown:139 - file:///usr/share/sailfish-browser/pages/components/TabItem.qml:139:13: QML Image: Cannot open: file:///home/defaultuser/.cache/org.sailfishos/browser/tab-2-thumb.jpg
```
For each tab that I have.

rainemak (changes requested):
Getting following after creating a tab, open url, waiting url to be loaded and pressing tabs icon:

[W] unknown:139 - file:///usr/share/sailfish-browser/pages/components/TabItem.qml:139:13: QML Image: Cannot open: file:///home/defaultuser/.cache/org.sailfishos/browser/tab-1-thumb.jpg

adenexter:
> Getting following after creating a tab, open url, waiting url to be loaded and pressing tabs icon:
> 
I started browser from the launcher rather than the command line to test some later revisions for the first time in a while and forgot the booster restart dance.

The writes to QSaveFile are now committed before it is destroyed.

rainemak (approved):
Looks good now!

apps/qtmozembed/declarativewebpage.cpp
```diff
@@ -301,9 +301,12 @@ void DeclarativeWebPage::forceChrome(bool forcedChrome)
 
 void DeclarativeWebPage::grabResultReady()
 {
-    if (active()) {
-        QImage image = m_grabResult->image();
-        m_grabWritter.setFuture(QtConcurrent::run(this, &DeclarativeWebPage::saveToFile, image));
+    QImage image = m_grabResult->image();
```
okodron, line 304: Is it ok that now we always create QImage regardless of `active()`?

apps/qtmozembed/declarativewebpage.cpp
```diff
@@ -344,15 +347,20 @@ void DeclarativeWebPage::updateViewMargins()
...
+        return QString();
     }
 
-    // 75% quality jpg produces small and good enough capture.
-    QString path = QString("%1/tab-%2-thumb.jpg").arg(BrowserPaths::cacheLocation()).arg(tabId());
-    return !allBlack(image) && image.save(path, "jpg", 75) ? path : "";
+    QSaveFile saveFile(path);
+    if (image.save(&saveFile, "jpg", 75)) {
+        saveFile.commit();
+
```
okodron, line 359: Empty line just looks strange here.

apps/qtmozembed/declarativewebpage.cpp
adenexter: The image already exists at this point, this is just taking a copy of what is effectively a shared pointer to the actual image data.

okodron (approved):

apps/qtmozembed/declarativewebpage.cpp
okodron: Ok.

PR #958, opened 2021-12-07
[browser] Add test case for target="_blank" download triggering. JB#56631 OMP#JOLLA-569

rainemak:
The bug is not triggering always but at least with this it's relatively easy to get it reproduced and backtrace from it.


okodron (approved):

rainemak:
Rebased

mkenttala (approved):

PR #959, opened 2021-12-08
[sailfish-browser] Use a press and hold scheme in tab list to switch …

dcaliste:
…to close mode.

Coming from a discussion in the forum (https://forum.sailfishos.org/t/browser-redesign-in-sailfish-4-2-feedback-thread/7867/78 and later posts), it was proposed to use a press and hold behaviour to activate a close tab mode in the tab list view. Like in the home screen (Lipstick switcher).

I've investigated the possibility and created a PR to test the idea. It's using the same metrics (animation duration, scaling factor...) than in the Lipstick switcher.

Here is a screen shot :
close

Entering this mode would allow later to implement a tab reordering capability, like in the Lipstick switcher.

@rainemak or @jpetrell , what do you think ?

PS : @rainemak , I've removed the timer in `TabItem.qml` that you added in 2014 commit bab080154 when the tabs were using shader effects and textures. I think it's safe to remove it now that we're using images. Do you agree ?

### DIFF
--- a/apps/browser/qml/pages/components/TabGridView.qml
+++ b/apps/browser/qml/pages/components/TabGridView.qml
@@ -28,6 +28,8 @@ SilicaGridView {
                                         : parent.width < 2 * parent.height
                                           ? parent.width <= height ? 1 : 2 : 3
     readonly property real thumbnailWidth: (parent.width - Theme.horizontalPageMargin * 2 - (Theme.paddingLarge * (columns - 1))) / columns
+    readonly property real _tabScale: housekeeping ? 0.9 : 1.0
+    property bool housekeeping
 
     signal hide
     signal enterNewTabUrl
@@ -66,21 +68,69 @@ SilicaGridView {
     cellWidth: thumbnailWidth + Theme.paddingLarge
     cellHeight: thumbnailHeight + Theme.paddingLarge
 
-    delegate: TabItem {
-        id: tabItem
+    delegate: Item {
+        id: item
 
-        enabled: !closingAllTabs
-        opacity: enabled ? 1.0 : 0.0
+        opacity: !closingAllTabs ? 1.0 : 0.0
         Behavior on opacity { FadeAnimator {}}
 
         width: thumbnailWidth
         height: thumbnailHeight
 
+        scale: tabGridView._tabScale
+        Behavior on scale { SmoothedAnimation {duration: 200; velocity: -1} }
+
         GridView.onAdd: AddAnimation {
-            target: tabItem
+            target: item
         }
         GridView.onRemove: RemoveAnimation {
-            target: tabItem
+            target: item
+        }
+
+        TabItem {
+            id: tabItem
+            anchors.fill: parent
+            enabled: !closingAllTabs
+            showClose: !tabGridView.housekeeping
+
+            onClicked: {
+                if (tabGridView.housekeeping) {
+                    tabGridView.housekeeping = false
+                } else {
+                    activateTab(index)
+                }
+            }
+            onPressAndHold: tabGridView.housekeeping = true
+            onCloseClicked: closeTab(index)
+        }
+        Rectangle {
+            color: Theme.colorScheme === Theme.LightOnDark
+                ? closeButton.down ? Theme.highlightColor : Theme.primaryColor
+                : closeButton.down ? Theme.highlightDimmerColor : Theme.highlightBackgroundColor
+            width: Theme.iconSizeMedium
+            height: width
+            radius: width / 2.
+            anchors.horizontalCenter: parent.horizontalCenter
+            anchors.verticalCenter: parent.bottom
+            opacity: tabGridView.housekeeping ? 1.0 : 0.0
+            visible: opacity > 0.
+            Behavior on opacity { FadeAnimation { } }
+
+            IconButton {
+                id: closeButton
+                anchors.centerIn: parent
+                icon.color: Theme.colorScheme === Theme.LightOnDark
+                    ? Theme.highlightBackgroundColor : Theme.lightPrimaryColor
+                icon.highlightColor: Theme.colorScheme === Theme.LightOnDark
+                    ? Theme.highlightDimmerColor : Theme.darkSecondaryColor
+                icon.highlighted: down
+                icon.source: "image://theme/icon-close-app"
+                enabled: !closingAllTabs && tabGridView.housekeeping
+                onClicked: {
+                    tabItem.markClosed()
+                    closeTab(index)
+                }
+            }
         }
     }
 
@@ -90,7 +140,7 @@ SilicaGridView {
             z: -1
             width: tabGridView.width
             height: tabGridView.height
-            onClicked: hide()
+            onClicked: tabGridView.housekeeping = false
         }
     ]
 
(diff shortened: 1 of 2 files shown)


apps/browser/qml/pages/components/TabGridView.qml
```diff
@@ -28,6 +28,7 @@ SilicaGridView {
                                         : parent.width < 2 * parent.height
                                           ? parent.width <= height ? 1 : 2 : 3
     readonly property real thumbnailWidth: (parent.width - Theme.horizontalPageMargin * 2 - (Theme.paddingLarge * (columns - 1))) / columns
```
rainemak, line 30: Having `property bool housekeeping` would improve readability. This property could be then in bindings where you are know using _tabScale. For example this would make onClicked (and onPressAndHold) handler clear in the TabItem.

apps/browser/qml/pages/components/TabGridView.qml
```diff
@@ -66,21 +67,68 @@ SilicaGridView {
...
+            target: item
+        }
+
+        TabItem {
+            id: tabItem
+            anchors.fill: parent
+            enabled: !closingAllTabs
+
+            onClicked: {
+                if (tabGridView._tabScale == 1.0) {
```
rainemak:
```suggestion
                if (tabGridView.housekeeping) {
                    tabGridView.housekeeping = false
                } else {
                    activateTab(index)
                }
```
Usually reading positive conditions are more readable than going through negative (if not housekeeping). Hence, I'd change order as well.

apps/browser/qml/pages/components/TabGridView.qml
```diff
@@ -66,21 +67,68 @@ SilicaGridView {
...
+                    activateTab(index)
+                } else {
+                    tabGridView._tabScale = 1.0
+                }
+            }
+            onPressAndHold: tabGridView._tabScale = 0.9
+        }
+        Rectangle {
```
rainemak, line 106: This Rectangle & Image & MouseArea should really be an Rectangle & IconButton. IconButton will handle highlighting of the icon. The background Rectangle can be pretty much like already declared (small remark below).

apps/browser/qml/pages/components/TabGridView.qml
```diff
@@ -66,21 +67,68 @@ SilicaGridView {
...
+            color: Theme.colorScheme === Theme.LightOnDark
+                ? mouseArea.pressed ? Theme.highlightColor : Theme.primaryColor
+                : mouseArea.pressed ? Theme.highlightDimmerColor : Theme.highlightBackgroundColor
+            width: portrait ? Theme.iconSizeLarge : Theme.iconSizeMedium
+            height: width
+            radius: width / 2.
+            anchors.horizontalCenter: parent.horizontalCenter
+            anchors.verticalCenter: parent.bottom
+            opacity: tabGridView._tabScale < 1.0 ? 1.0 : 0.0
```
rainemak:
```suggestion
            opacity: tabGridView.housekeeping ? 0.0 : 1.0
```

apps/browser/qml/pages/components/TabGridView.qml
```diff
@@ -66,21 +67,68 @@ SilicaGridView {
...
+                anchors.centerIn: parent
+                source: "image://theme/icon-close-app?" + (Theme.colorScheme === Theme.LightOnDark
+                    ? mouseArea.pressed ? Theme.highlightDimmerColor : Theme.highlightBackgroundColor
+                    : mouseArea.pressed ? Theme.darkSecondaryColor : Theme.lightPrimaryColor)
+            }
+
+            MouseArea {
+                id: mouseArea
+                anchors.fill: parent
+                enabled: !closingAllTabs && tabGridView._tabScale < 1.0
```
rainemak: This enabled would be then enabling of the IconButton.

apps/browser/qml/pages/components/TabItem.qml
```diff
@@ -17,17 +17,20 @@ import Sailfish.Silica 1.0
...
-    property bool destroying
     property color highlightColor: Theme.colorScheme == Theme.LightOnDark
                                    ? Theme.highlightColor
                                    : Theme.highlightFromColor(Theme.highlightColor, Theme.LightOnDark)
     // In direction so that we can break this binding when closing a tab
     implicitWidth: width
     implicitHeight: height
 
-    enabled: !destroying
+    function closing() {
```
rainemak:
```suggestion
    function closing() {
```
Closing is a state/status whereas function name should be describing what this function does (verb). I would not call this "close" either as this is not really closing anything. That said, I'd propose "markClosed" as this is changing TabItem to itself final state.

rainemak (changes requested):
Almost there I'd say. Like discussed during the community meeting, people have gotten used to the close button that can be found from the toolbar menu. That's my favorite as well.

Maybe the biggest concern that I have is whether this actually improves situation or not but I'll to elaborate a way that should render good end results. I think we should leave close icon still to the TabItem header (where it was, at least for time being) as that would be a bit quicker way to close. Then this PR & commit would be more about introducing a housekeeping mode for tabs where closing is first action and re-organizing tab order could be implemented later.

apps/browser/qml/pages/components/TabGridView.qml
dcaliste: Indeed, good idea. It's more intentional with such a property.

apps/browser/qml/pages/components/TabGridView.qml
dcaliste: Yep, I've changed all use of `_tabScale` to use `housekeeping` instead and reverse order when necessary to avoid negative in tests.

apps/browser/qml/pages/components/TabGridView.qml
dcaliste: Good advice, thanks. I'm using an IconButton as done before in TabItem.qml.

apps/browser/qml/pages/components/TabGridView.qml
dcaliste: Done.

apps/browser/qml/pages/components/TabGridView.qml
dcaliste: Moved to it.

apps/browser/qml/pages/components/TabItem.qml
dcaliste: Ok, I've changed the name for your suggestion.

dcaliste:
Thank you @rainemak for your review and suggestions. I've updated the PR accordingly.

I'm a bit wondering though : do you prefer to have this code to TabItem.qml ? If so, I can do, no problem.

rainemak:
I guess rebase went wrong, maybe you rebased from master instead of next.

dcaliste:
Oops, stupid me. Corrected, now based on next, and not anymore 55 commit to push ;-)

apps/browser/qml/pages/components/TabItem.qml
```diff
@@ -110,31 +111,6 @@ BackgroundItem {
                     truncationMode: TruncationMode.Fade
                     color: down || activeTab ? root.highlightColor : Theme.primaryColor
                 }
-
-                IconButton {
```
rainemak: I'd keep this close button as well until housekeeping mode has also re-organizing tabs functionality.

dcaliste: Ok, thanks. I understand your point now. I'll update the PR after my vacations are over, second half of August.

dcaliste:
It's now rebased with all remarks addressed (if I didn't missed any) and the close icon back on TabItem.

PR #962, opened 2021-12-13
[browser] Add ability of using cutout API. Contributes to JB#56709

atatarov:


rainemak:
There's nothing OMP specific in the commit per se. Please drop "from omp side" from the commit message and merge commit message.

apps/shared/BrowserWindow.qml
```diff
@@ -58,4 +58,10 @@ ApplicationWindow {
             }
         }
     }
+
+    Component.onCompleted: {
+        if (window.hasOwnProperty("displayMode")) {
+            displayMode = ApplicationDisplayMode.FullPortrait
```
rainemak, line 64: Where is displayMode and ApplicationDisplayMode defined?

llewelld (approved):
I had a quick test and didn't notice any issues with this. So looks good to me. But I'd also await @rainemak's approval before merging.

okodron (approved):

apps/shared/BrowserWindow.qml
rainemak: For adding a comment that this is coming from Silica.

PR #964, opened 2021-12-15
[browser] Pass preference type argument when setting a pref. JB#56638 OMP#JOLLA-573

rainemak:
This commit also adds an int validator for int text fields. By default
int validator limits to min & max int values.

Similar error messages as in desktop Firefox (about:config) are shown
when int text fields reach limits.


rainemak:
See also https://github.com/sailfishos/qtmozembed/pull/33

mkenttala (approved):

okodron (approved):

apps/browser/qml/pages/components/ConfigDialog.qml
```diff
@@ -98,26 +99,50 @@ Dialog {
...
+
+                    label: {
+                        if (errorHighlight && digitsOnly) {
+                            if (text.charAt(0) == "-") {
+                                //% "Select a value that is no less than -2147483648"
+                                return qsTrId("sailfish_browser-la-min_int_warning")
+                            }
+
+                            //% "Select a value that is no more than 2147483647"
+                            return qsTrId("sailfish_browser-la-max_int_warning")
```
llewelld, line 118: In practice, the `errorHighlight` seems to cover more cases than just bounding (e.g. entering a decimal like "0.01" or paste a string "abcd"). It might be safer just to have error text "Please enter only integer values" or something like that.

apps/browser/qml/pages/components/ConfigDialog.qml
```diff
@@ -98,26 +99,50 @@ Dialog {
...
                     text: model.value
                     placeholderText: model.name
-                    inputMethodHints: (model.type == 64 ? Qt.ImhDigitsOnly : 0)
+                    inputMethodHints: (digitsOnly ? Qt.ImhDigitsOnly : 0)
                     width: parent.width
                     height: Theme.itemSizeMedium
 
+                    Component.onCompleted: {
+                        if (digitsOnly)
+                            validator = intValidator
```
llewelld, line 130: Braces, even for one-line conditionals (I understand this is our convention).

apps/browser/qml/pages/components/ConfigDialog.qml
```diff
@@ -98,26 +99,50 @@ Dialog {
...
+                    }
+
                     onTextChanged: {
                         if (text === model.value) return;
 
-                        if (model.type == 64) {
-                            changedConfigs[model.name] = parseInt(text);
+                        if (digitsOnly) {
+                            changedConfigs[model.name] = {
+                                value: parseInt(text),
```
llewelld, line 138: I wonder if it's worth forcing to base 10, to avoid zero prefixes getting interpreted as octals.

llewelld (commented):
Just some very minor comments.

apps/browser/qml/pages/components/ConfigDialog.qml
rainemak: Fine for me.

rainemak: Also description does work nicely if you're wondering.

apps/browser/qml/pages/components/ConfigDialog.qml
rainemak: Shouldn't harm

rainemak:
Resolved and fixed review findings / comments.

llewelld (approved):
Thanks, looks good to me.

llewelld:
Does @sledges need to be aware of the new string?

rainemak:
> Does @sledges need to be aware of the new string?

Not big deal if we don't translate it yet

PR #966, opened 2021-12-17
[browser] Add test case for opening tel uri into a new tab. JB#56648 OMP#JOLLA-577

rainemak:
This should be simple enough test case for this.


mkenttala (approved):

okodron (approved):

rainemak:
Updated test case to cover tel, sms, mailto, and geo with target="_blank".

PR #969, opened 2021-12-30
[browser] Clean tabs in reboot. Fixes JB#54782

mkenttala:


apps/browser/main.cpp
```diff
@@ -62,8 +64,22 @@ static QObject *bookmarkmanager_factory(QQmlEngine *, QJSEngine *)
 }
 }
 
+static void sighandler(int signal)
+{
+    Q_UNUSED(signal)
+    DeclarativeWebContainer::instance()->closeApplication();
```
adenexter, line 70: Virtually nothing is safe within an event handler so you can't call any functions except some system calls.

The normal procedure to create an eventfd which can be watched from the application main thread with a QSocketNotifier, and signaled by writing to it from the signal handler.

I'm not sure it's the best example but lipstick has an implementation of that:

https://github.com/sailfishos/lipstick/blob/master/src/homeapplication.cpp#L57
https://github.com/sailfishos/lipstick/blob/master/src/homeapplication.cpp#L214

Also have a look at the silica booster or something it extends, because I seem to remember something in there  installing a signal handler which call QCoreApplication::exit() when the application is boosted.  It probably doesn't matter because this will just override it anyway and circumstantially installed handlers aren't much good.

mkenttala: Tried to get socket notifier to work but the app got killed before the notify arrived. So now listening to dsme reboot state change instead.

rainemak (commented):
I think it's matter of taste whether proposed or existing behaviour is right. I'm fine with this as well. What's your opinion @llewelld ?

rainemak (approved):

PR #970, opened 2022-01-05
[browser] Fix painting of pages loaded with pushState. JB#56556

mkenttala:


apps/core/declarativewebcontainer.cpp
```diff
@@ -198,6 +198,8 @@ void DeclarativeWebContainer::setWebPage(DeclarativeWebPage *webPage, bool trigg
                     this, &DeclarativeWebContainer::updateActiveTabRendered, Qt::UniqueConnection);
             connect(m_webPage.data(), &QOpenGLWebPage::domContentLoadedChanged,
                     this, &DeclarativeWebContainer::updateActiveTabRendered, Qt::UniqueConnection);
+            connect(m_webPage.data(), &QOpenGLWebPage::firstPaint,
+                    this, &DeclarativeWebContainer::updateActiveTabRendered, Qt::UniqueConnection);
```
rainemak, line 202: Following block should be fixed...
```
 if (m_webPage->completed() && m_webPage->active() && m_webPage->domContentLoaded()) {
     m_webPage->update();
 }
```

to:
```
if (m_webPage->completed() && m_webPage->active() && (m_webPage->isPainted() || m_webPage->domContentLoaded())) {
    m_webPage->update();
}
```

apps/core/declarativewebcontainer.cpp
```diff
@@ -1040,7 +1042,7 @@ void DeclarativeWebContainer::updateLoading()
 
 void DeclarativeWebContainer::updateActiveTabRendered()
 {
-    if (m_webPage && !m_webPage->completed() && !m_webPage->domContentLoaded()) {
+    if (m_webPage && !m_webPage->completed() && !m_webPage->domContentLoaded() && !m_webPage->isPainted()) {
```
rainemak, line 1045:
```suggestion
    if (!m_webPage || !m_webPage->completed() || (!m_webPage->isPainted() && !m_webPage->domContentLoaded())) {
```
Either isPainted or domContentLoaded is enough for the update.

apps/core/declarativewebcontainer.cpp
```diff
@@ -1040,7 +1042,7 @@ void DeclarativeWebContainer::updateLoading()
 
 void DeclarativeWebContainer::updateActiveTabRendered()
 {
-    if (m_webPage && !m_webPage->completed() && !m_webPage->domContentLoaded()) {
+    if (!m_webPage || !(m_webPage->completed() || m_webPage->domContentLoaded() || m_webPage->isPainted())) {
```
rainemak, line 1045: In practise completed is always true as it requires that page initialized and it cannot be rendered without being initialized. Still prerequisite for these two other variables.

The latter part of the condition (in the brackets) evaluates always to false as inside brackets is always true. This return statement should be reached when we do not have webpage and neither domContentLoaded nor isPainted is true.

rainemak (approved):
LGTM. Didn't test this final version on device but looks good. Please rebase before merging if needed.

PR #976, opened 2022-01-13
[browser] Show back button to close a child tab. Fixes JB#56975

rainemak:
Let's give a test rid with this and gather internal feedback.

### DIFF
--- a/apps/browser/qml/pages/components/ToolBar.qml
+++ b/apps/browser/qml/pages/components/ToolBar.qml
@@ -202,8 +202,14 @@ Column {
             id: backIcon
             expandedWidth: toolBarRow.iconWidth
             icon.source: "image://theme/icon-m-back"
-            active: webView.canGoBack && !findInPageActive
-            onTapped: webView.goBack()
+            active: (webView.canGoBack || webView.contentItem.parentId > 0) && !findInPageActive
+            onTapped: {
+                if (webView.canGoBack) {
+                    webView.goBack()
+                } else {
+                    webView.tabModel.closeActiveTab()
+                }
+            }
         }
 
         Shared.ExpandingButton {


mkenttala (approved):

apps/browser/qml/pages/components/ToolBar.qml
```diff
@@ -202,8 +202,14 @@ Column {
             id: backIcon
             expandedWidth: toolBarRow.iconWidth
             icon.source: "image://theme/icon-m-back"
-            active: webView.canGoBack && !findInPageActive
-            onTapped: webView.goBack()
+            active: (webView.canGoBack || webView.contentItem.parentId > 0) && !findInPageActive
+            onTapped: {
+                if (webView.canGoBack) {
```
deepbluev7, line 207: The problem with this imo is, that if you hit back a few times (i.e. to get back to your search results), this will then just close your tab and there is no remorse timer to get it back. That would be very annoying. (Also the icon should change to match the action)

rainemak: This back navigation applies only for tabs that got created by JS i.e. parentId is greater than zero. A tab that got created by JS can be closed (window.close()) by the parent JS and well as by the child itself -- any time. Tabs that are root tabs (no parent) cannot be closed with back button over here or from the JS.

The same behaviour have been implemented to Android FF and mobile Safari.

llewelld (approved):
This LGTM!

apps/browser/qml/pages/components/ToolBar.qml
rainemak: We'll consider the icon separately and do some experiments with it.

rainemak:
@llewelld added one commit on top that takes icon-m-back-tab into use. There's fallback code path for the icon-m-back.

llewelld:
This still LGTM.

PR #981, opened 2022-01-31
[user-agent] Update preprocessed user agent overrides for m.vk.com. Fixes JB#57309

sergey-ivantsov:
I found that changing user-agent to desktop Firefox one helps to fix the problem of constantly reloading m.vk.com page (it starts reloading after you log in). Looks like it is a site problem and we need to adjust our user-agent to overcome it.
I will add the JB task number later when I get my Jolla account.


sergey-ivantsov:
Task in bugzilla: https://bz.jollamobile.com/show_bug.cgi?id=57309

mkenttala (approved):

PR #985, opened 2022-02-16
Stop embedding after views&window destoyed (tst_webview)

rainemak:
This PR contains several commits. Might be easier to review per commit.

Seems to exit now quite reliable but I think that exit occasionally segfaults still. I'll keep eye on the test case.


mkenttala (approved):

tests/auto/tst_webview/tst_webview.cpp
```diff
@@ -922,7 +906,70 @@ int main(int argc, char *argv[])
...
+        if (testcase->running) {
+            return;
+        }
+
+        contextDestroyed = true;
+        delete testcase;
+        testcase = nullptr;
+        QTest::waitForEvents();
+    });
+
```
rainemak, line 947: On extra line.

PR #986, opened 2022-02-17
[browser] Fix tab closing in app exit. Fixes JB#56247

mkenttala:


apps/core/declarativewebcontainer.cpp
```diff
@@ -678,6 +678,7 @@ bool DeclarativeWebContainer::eventFilter(QObject *obj, QEvent *event)
 {
     if (obj == m_chromeWindow) {
         if (event->type() == QEvent::Close) {
+            emit applicationClosing();
```
rainemak, line 681: This should after line 691.

mkenttala: Wanted to handle that tab closing as fast as possible before app exit but later seems to work just fine, fixed.

apps/core/declarativewebcontainer.cpp
```diff
@@ -690,6 +690,7 @@ bool DeclarativeWebContainer::eventFilter(QObject *obj, QEvent *event)
                     m_closeEventFilter->closeApplication();
                 }
             }
+            emit applicationClosing();
```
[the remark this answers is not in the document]
rainemak: But from user point of view as the window has been closed, it should be ok to commit tabs closing.

rainemak (approved):

PR #990, opened 2022-02-22
[browser] Fix popup input region. Fixes JB#57553

mkenttala:


apps/browser/qml/pages/BrowserPage.qml
```diff
@@ -66,7 +66,7 @@ Page {
         var mask = Qt.rect(0, 0,
                            portraitScreen ? Screen.width : Screen.height,
                            portraitScreen ? Screen.height : Screen.width)
-        if (webView.enabled && browserPage.active && !downloadPopup.visible) {
+        if (webView.enabled && browserPage.active && !webView.contentItem.popupOpener.active && !downloadPopup.visible) {
```
rainemak, line 69: I'd rather return `!webView.touchBlocked &&` back from sha1 ff92005de4c5cc828e037e98badbe8fb14c2155c and update condition of the touchBlocked to not include `selectionActive`.

As it is now this can cause errors due to many issues
1) What happens if close the last tab i.e. contentItem changes to null?
2) What if popupOpener is not yet created?

rainemak (approved):

PR #992, opened 2022-03-09
[browser] Fix icon for add to app grid. Fixes JB#57628

mkenttala:


llewelld:
This is a nice fix in the case where the site has a favIcon set (it worked for me, certainly).

Is it also worth covering the case where there's no favicon and taking a snapshot of the page in that case, as is done for the favourites grid? I.e. using grabThumbnail as happens in apps/browser/qml/pages/components/FavoriteGrid.qml ?

mkenttala:
Now grabs the icon from the page if favicon is empty.

llewelld:
This is working nicely, the icon appears even if there's no favicon now; great!

However, I also notice a strange side effect in that if there's no favicon for the page, pressing on the Favourite star in the toolbar footer now opens the "Add to app grid" page.

This only happens after "Add to app grid" has been selected at least once prior to this, so presumably it's because the `webPage.onThumbnailResult` signal is still connected (or something like that)?

mkenttala:
Disconnect added for the grab handler. Some problems with the scope of variables but seems ok now.

llewelld:
This works nicely now, thank you!

llewelld (approved):
LGTM.

PR #996, opened 2022-04-14
[browser] Do not raise overlay when starting with requested url. Fixes JB#56791

rainemak:


abranson (approved):
This fixes it for me! Thanks Raine!

PR #997, opened 2022-04-14
[browser] Remove browser application name from google domain overrides. Fixes JB#26304

rainemak:
This PR drops SailfishBrowser/1.0 from the google domain overrides
making it possible to use more user-friendly mobile version of the
gmail.com website.

Tips for testing PR
1) Build this PR
2) On device: rm $HOME/.local/share/org.sailfishos/browser/.mozilla/ua-update.json
3) Install new sailfish-browser package to the device
4) systemctl --user restart booster-browser@sailfish-browser.service

You can verify that ua-update.json looks the same as in this PR
> cat .local/share/org.sailfishos/browser/.mozilla/ua-update.json

Once this PR is merged, I will land this to the 78.8 over-the-air update channel.


llewelld (approved):
After this change Google switched me from "HTML" Gmail to the "App" version. The latter gives a much better Gmail experience, and with no problems downloading attachments.

So, nicely figured out, it's a really nice result.

PR #999, opened 2022-05-18
[ua] Add linux for facebook user-agent. Fixes JB#58153

rainemak:
The like button was not working without this.

Signed-off-by: Raine Makelainen <raine.makelainen@jolla.com>


llewelld (approved):
LGTM.

PR #1004, opened 2022-10-04
Encode URL string before copying to clipboard

nephros:
Proposed fix for: https://forum.sailfishos.org/t/4-4-0-72-browser-url-copy-does-not-encode-uri-string/13152

I believe it improves usability if strings pasted after copying from browser are valid URLs, not beautified strings.

### DIFF
--- a/apps/browser/qml/pages/components/ToolBar.qml
+++ b/apps/browser/qml/pages/components/ToolBar.qml
@@ -298,7 +298,9 @@ Column {
             onPressAndHold: {
                 var url = webView.url
                 if (url) {
-                    Clipboard.text = url
+                    // encode the string if it looks like it has query or fragment parts
+                    // FIXME: could be improved with *proper* matching.
+                    Clipboard.text = ( (url.indexOf('?') > -1) || (url.indexOf('#') > -1) ) ? encodeURI(url) : url
                     urlCopyNotice.show()
                 }
             }


Thaodan:
Changed target to next as PRs should go to that branch.

nephros:
rebased against `next` and squashed

nephros:
Not familiar with the process yet. Am I supposed to un-draft this myself or will it be done whe someone is ready to review?

apps/browser/qml/pages/components/ToolBar.qml
```diff
@@ -298,7 +298,7 @@ Column {
             onPressAndHold: {
                 var url = webView.url
                 if (url) {
-                    Clipboard.text = url
+                    Clipboard.text = encodeURI(url)
```
rainemak: In theory this should be applied only for fragment and query parts of the url, see whatwg [1]. For example space if not allowed for host part. Simple workaround would be to encode uri only if url is resolved. This would require a bit of changes to QtMozEmbed to expose isUrlResolved property (currently just public method) after that checking here whether url is resolved. Something like this:

```
                if (url) {
                    Clipboard.text = webView.contentItem.urlResolved ? encodeURI(url) : uri
                }
```

That said, this is quite a corner case. Would be mostly visible in slow networks and when entering search query and quickly long pressing url (e.g. Hello World). Your call @nephros, I'm happy with this as well.

In both cases, please just remove the draft status yourself.

[1] 
https://url.spec.whatwg.org/

nephros: >  This would require a bit of changes to QtMozEmbed to expose isUrlResolved property

I feel the copying to clipboard should not cause a trip to WebView and trying to resolve the URL.

Would it be acceptable to only encode if "?" or "#" are found in the original URL, i.e. it looks like it has a query and/or fragment?

Something like

```
Clipboard.text = ( url.indexOf("?") || url.indexOf("#") ) ? encodeURI(url) : url
```

with possibly a proper non-greedy RegExp test [1] instead of the simple indexOf.

[1] https://www.rfc-editor.org/rfc/rfc3986#appendix-B

rainemak: Trip to WebView is not bad at all as that would be just property getter and resolved as the page load more about changes need to QtMozEmbed I think. Your proposal is good enough, please update and let's go with that. This can be further improved if one likes.

rainemak (approved):
Thank you @nephros for your first browser contribution! This is a good starting point!

PR #1006, opened 2022-11-15
[browser] Delay external tab requests at startup. Fixes JB#57763

llewelld:
Makes two changes:

First, when the browser is started it loads the persistent tab model
asynchronously. Consequently if an external request to open a tab via
D-Bus triggered the browser to load (i.e. it wasn't already running),
the tab could be created before the tab model was loaded. This caused
the PID associated with the created tab to be wiped, and prevented the
external process that opened the tab from closing it again (used e.g. in
OAuth flows).

This change avoids this by delaying creation of tabs requested
externally until after the tab model has completed loading.

Second, when the browser runs inside the sandbox it's not able to check 
whether the sender is in the privileged group, so the check is guaranteed 
to fail.

This change removes the check.

It leaves it in for the case of dumpMemoryInfo() as this may be useful
when running the browser unsandboxed for debugging purposes, and without
the check there would be no other restrictions on processes calling it.

### DIFF
--- a/apps/browser/browserservice.cpp
+++ b/apps/browser/browserservice.cpp
@@ -25,7 +25,8 @@
 namespace {
 const auto ProcDir = QStringLiteral("/proc/%1");
 const auto ErrorPidIsNotPrivileged = QStringLiteral("PID %1 is not in privileged group");
-const auto ErrorPidIsNotOwnerServiceOrPrivileged = QStringLiteral("PID %1 is not the owner of '%2' or in privileged group");
+const auto ErrorPidIsNotOwnerService = QStringLiteral("PID %1 is not the owner of '%2'");
+const auto ErrorPidIsNotTabOwner = QStringLiteral("PID %1 is not the owner of the tab");
 const auto SailfishBrowserUiService = QStringLiteral("org.sailfishos.browser.ui");
 const auto TransferEngine = QStringLiteral("org.nemo.transferengine");
 }
@@ -91,6 +92,8 @@ void BrowserService::restartTransfer(int transferId)
 
 void BrowserService::dumpMemoryInfo(const QString &fileName)
 {
+    // The privileged check will always fail when the browser is sandboxed
+    // It can still be useful for debugging purposes running outside the sandbox
     if (!isPrivileged()) {
         return;
     }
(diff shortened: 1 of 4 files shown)


llewelld:
A note for reviewers: it may be easy to consider the two commits separately.

apps/browser/browserservice.cpp
```diff
@@ -201,40 +198,18 @@ void BrowserUIService::requestTab(int tabId, const QString &url)
 void BrowserUIService::closeTab(int tabId)
 {
     // Let privileged and the process that created the tab to also destroy it
```
Tomin1: I think you should adjust this comment too.

apps/core/declarativewebcontainer.cpp
```diff
@@ -528,13 +529,35 @@ int DeclarativeWebContainer::activateTab(int tabId, const QString &url)
...
+            // We assume that m_model->loaded() is now set to true
+            int activatedTab = requestTabWithOwner(tabId, url, ownerPid);
+            qCDebug(lcCoreLog) << "Delaying tab request created tabId:" << activatedTab;
+            m_reservedTabs = 0;
+            // Single-shot connection
+            QObject::disconnect(*connection);
+            delete connection;
+        });
+
+        // Return the predicted tabId
```
Tomin1: I'm just wondering which would make more sense: predicting `tabId` or delaying D-Bus call return until the id is known.

Tomin1 (commented):
Some initial thoughts on this. In principle this looks good, but I think that it would probably be best to set delayed reply to true and return the actual id once the tab can be actually opened. Also I wonder if we should get rid of that `IS_PRIVILEGED()` macro as it's used only once in the code now, but that's not important.

apps/browser/browserservice.cpp
llewelld: Thanks for noticing, I've changed it.

apps/core/declarativewebcontainer.cpp
llewelld: Thanks for the suggestion. I initially tried delaying but hit some difficulties, but I've had another go and I think this is nicer now. See the additional commit; what do you think?

Tomin1 (approved):
LGTM.

apps/core/declarativewebcontainer.cpp
```diff
@@ -540,6 +540,29 @@ int DeclarativeWebContainer::requestTabWithOwner(int tabId, const QString &url,
...
+{
+    // We should only create or activate tabs once the model has loaded
+    if (m_model->loaded()) {
+        // The tab model has already loaded, so we can go ahead and create the tab
+        int activatedTab = requestTabWithOwner(tabId, url, ownerPid);
+        emit requestTabWithOwnerAsyncResult(activatedTab, context);
+    } else {
+        // The model has yet to load, so queue creation of the tab
+        QMetaObject::Connection * const connection = new QMetaObject::Connection;
+        *connection = connect(m_model.data(), &DeclarativeTabModel::loadedChanged, [this, tabId, url, ownerPid, context, connection]() {
```
Tomin1, line 553: Oh, one comment: could and should this take `this` as context?
https://doc.qt.io/archives/qt-5.6/qobject.html#connect-5

llewelld: What's your concern in particular? The instance needs to be available to emit the signal.

Tomin1: Ah. Now I see what you mean. The model instance won't exist if `this` has been destroyed. Everything is fine here, nothing to see. :see_no_evil:

llewelld:
Rebased.

PR #1014, opened 2023-05-28
Enable dialog element

attah:
The dialog element is now standard, so this is one tiny thing that can be done to make the browser feel less dated.
It was apparently in testing/incubation for a very long time, but should work since Firefox 53, if enabled. See https://caniuse.com/dialog

Only tested with locally modified setting.


rainemak:
Nice small improvement. Would you like to add the pref to the https://github.com/sailfishos/gecko-dev/blob/master/embedding/embedlite/embedding.js (as a pref)? This way also WebView would benefit from this.

attah:
Done

rainemak:
> Done

Perfect! Thank you.

PR #1057, opened 2024-04-08
Remove transfer when privateMode is set

ilpianista:
Follow up of [transfer-engine!14](https://github.com/sailfishos/transfer-engine/pull/14), the goal here is to hide transfers made from a private tab.

I think that is OK to remove the transfer when it's completed, cancelled or failed.

### DIFF
--- a/apps/core/downloadmanager.cpp
+++ b/apps/core/downloadmanager.cpp
@@ -58,6 +58,7 @@ void DownloadManager::recvObserve(const QString message, const QVariant data)
     QString msg = dataMap.value(QStringLiteral("msg")).toString();
     QString targetPath = dataMap.value(QStringLiteral("targetPath")).toString();
     bool isSaveAsPdf = dataMap.value(QStringLiteral("saveAsPdf")).toBool();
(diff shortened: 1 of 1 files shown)


apps/core/downloadmanager.cpp
```diff
@@ -132,6 +133,9 @@ void DownloadManager::recvObserve(const QString message, const QVariant data)
         } else if (isMyApp(targetPath)) {
             finishMyAppDownload(targetPath);
         }
+        if (isPrivateMode) {
+            m_transferClient->clearTransfer(m_download2transferMap.value(downloadId));
+        }
```
rainemak, line 122: Small downloads disappear now really quickly. I doubt that people find downloaded files from the file system. Instead of immediately clearing, I think it would be nicer to gather a list of completed private downloadIds over here. Then upon clearing of PrivateTabModel (tab count reaching zero and destruction of the model) clear also private downloads from the DownloadManager. How do you see this?

Briefly chatted this with @pvuorela as well.

rainemak:
Small comment, could you please re-target this towards next.

PR #1093, opened 2025-02-03
UI fixes

pvuorela:
Bunch of UI glitches fixed in separate commits:
- On overlay avoid flicking favorites on top of the history button
- Avoid notch area on generic settings type UI. Only the webview and tab view to use really fullscreen mode.
- Avoid notch area on overlay. Small UI change: avoid special case (add extra display cutout avoidance padding on text field) on special case (no tabs && no private mode && portrait orientation). Instead now this case opens with the small gap on top. Simplifies the logic and makes the logic saner.
- More style adjustments and removal of some unnecessary ids, etc.

More details in commit messages.

### DIFF
--- a/apps/browser/qml/pages/components/ConfigWarning.qml
+++ b/apps/browser/qml/pages/components/ConfigWarning.qml
@@ -18,7 +18,8 @@ ConfirmDialog {
     property Item browserPage
 
     //: Warning of changing browser configurations.
-    //% "Changing these advanced settings can cause issues with stability, security and performance of Sailfish Browser. Continue ?"
+    //% "Changing these advanced settings can cause issues with stability, "
+    //% "security and performance of Sailfish Browser. Continue ?"
     text: qsTrId("sailfish_browser-la-config-warning")
     acceptDestination: Component {
         ConfigDialog {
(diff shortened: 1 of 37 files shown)


apps/browser/qml/pages/components/ConfigWarning.qml
```diff
@@ -18,7 +18,8 @@ ConfirmDialog {
     property Item browserPage
 
     //: Warning of changing browser configurations.
-    //% "Changing these advanced settings can cause issues with stability, security and performance of Sailfish Browser. Continue ?"
+    //% "Changing these advanced settings can cause issues with stability, "
+    //% "security and performance of Sailfish Browser. Continue ?"
```
direc85, line 22: "Continue ?" with a space here?

spiiroin (approved):

direc85 (approved):

apps/browser/qml/pages/components/ConfigWarning.qml
pvuorela: I didn't make a change here :) Yea, would be better without but at this point I'm maybe a bit wary of adjusting the source strings.

PR #1097, opened 2025-02-23
[browser] Allow non-http bookmark URLs

nephros:
Allows saving from the bookmark edit dialog also when the URI does not start with https.

The most common case is probably file://, which currently can be bookmarked, but not edited.  
But folks may want to bookmark things like app:, geo:, or dav: as well

### DIFF
--- a/apps/browser/qml/pages/components/BookmarkItem.qml
+++ b/apps/browser/qml/pages/components/BookmarkItem.qml
@@ -97,6 +97,7 @@ ListItem {
             }
             MenuItem {
                 text: qsTrId("sailfish_browser-me-add_to_launcher")
+                enabled: /^(file|https?):\/\/.+/.test(url) // See also BookmarkEditDialog.qml
                 onClicked: pageStack.animatorPush("AddToAppGridDialog.qml",
                                                   {
                                                       "url": url,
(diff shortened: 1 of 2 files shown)


pvuorela:
I'm not sure how useful bookmarking local files would really be, but indeed it's now inconsistent when such can be bookmarked but not edited. Either it should not be possible to bookmark or then the editing should allow file:// scheme too. 

For the other types, I'm even more skeptical. Browser itself doesn't handle the mentioned types so the bookmarks would just trigger something else to really handle them. The something else maybe even having its own bookmarking functionality.

One bigger problem is that bookmarks can be also added to the app launcher via the bookmark grid context menu. If that's done for types that the browser doesn't handle then the items can end up dead. Browser itself is now registered for html files and http/s schemes and I wouldn't like to add anything more there.

So I could maybe accept file:// bookmarks but then those should have the app launcher part disabled.

apps/browser/qml/pages/components/BookmarkItem.qml
```diff
@@ -97,6 +97,7 @@ ListItem {
             }
             MenuItem {
                 text: qsTrId("sailfish_browser-me-add_to_launcher")
+                enabled: /^(file|https?):\/\/.+/.test(url) // See also BookmarkEditDialog.qml
```
pvuorela: Maybe a bit misunderstanding, this shouldn't be allowed for generic file urls since those will not be opened with the browser. Exception being .html files. Simplest would be just to require http/https for now.

Furthermore the AddToAppGridDialog inherits BookmarkEditDialog so there should be some property for restricting the validators there to keep on allowing only what we support for the .desktop files as it used to be (with the possible exception of .html files)

PR #1100, opened 2025-03-04
Refresh shortcut

pvuorela:
Added simple handler for f5 to refresh. Fixed some uninitialized memory issues and made page grab api a bit easier to folllow. Different commits for each.

### DIFF
--- a/apps/core/declarativewebcontainer.h
+++ b/apps/core/declarativewebcontainer.h
@@ -258,7 +258,7 @@ private:
     QPointer<QQuickItem> m_rotationHandler;
     QPointer<DeclarativeWebPage> m_webPage;
     QPointer<QQuickView> m_chromeWindow;
-    QOpenGLContext *m_context;
+    QOpenGLContext *m_context = nullptr;
     QMutex m_contextMutex;
 
     QPointer<DeclarativeTabModel> m_model;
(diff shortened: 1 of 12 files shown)


apps/core/declarativewebcontainer.h
```diff
@@ -295,7 +295,7 @@ private slots:
     bool m_closing;
 
     QHash<int, uint> m_tabOwners;
-    DeclarativeHistoryModel *m_historyModel;
+    DeclarativeHistoryModel *m_historyModel = nullptr;
```
mlehtima, line 299: Should this be in the cpp file like the other variable initializations? In the cpp it might be also nice to have consistent 0 vs nullptr usage.

apps/history/declarativetabfiltermodel.h
```diff
@@ -34,7 +34,7 @@ class DeclarativeTabFilterModel : public QSortFilterProxyModel
     void countChanged();
 
 private:
-    bool m_showHidden;
+    bool m_showHidden = false;
```
mlehtima, line 37: Maybe this should also be in constructor?

mlehtima (changes requested):

apps/core/declarativewebcontainer.h
pvuorela: Cpp file initialization is the reason the bug here exists. Too easy to create a member variable and then just forget to add it to the init list.

apps/history/declarativetabfiltermodel.h
pvuorela: No, this is the only member variable of the class so it's now consistently having all default values in .h

spiiroin (approved):

apps/core/declarativewebcontainer.h
pvuorela: Added a separate commit for consistently doing the declarativewebcontainer POD initialization in the header. 

Actually doing that I noticed that it was also missing initialization for m_fromExternal sort of proving the point :)

direc85 (approved):
API names make a lot more sense now, as well as moving the member variable initialisation to header. F5 works as well.

mlehtima (approved):

PR #1101, opened 2025-03-05
Remove myapp special download handling

pvuorela:
None of this should be needed anymore. CPU_ARM has been also false for 64-bit devices, getting them refer to jolla tablet.


rainemak (approved):
LGTM

PR #1104, opened 2025-03-10
Fullscreen close button and captive portal fixes

pvuorela:
Depends on https://github.com/sailfishos/embedlite-components/pull/106

Some fixes on captive portal side and then separate commit for adding exit button for fullscreen mode. Captive portal can be perhaps easiest tested by running from commandline 'sailfish-captiveportal -captiveportal https://some_url'

Don't like the redundancy on main app pages but it's already there. For captive portal there's still some problem showing the text selection handles, but let's see if I have enough energy to dig into that.

### DIFF
--- a/apps/browser/main.cpp
+++ b/apps/browser/main.cpp
@@ -19,7 +19,6 @@
 #include <QDBusPendingCall>
 
 #include "browser.h"
-#include "browserappinfo.h"
 // Registered QML types
 #include "declarativebookmarkmodel.h"
 #include "bookmarkfiltermodel.h"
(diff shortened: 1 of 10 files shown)


mlehtima (approved):

direc85 (approved):
Works great, except for one thing: if I click the button, move the mouse outside it and then release, the button steals focus. This prevents e.g. Esc from exiting fullscreen - which is nice that it works too!

Edit: Long-pressing the button does the same - it disappears under the mouse, stealing focus.

pvuorela:
> Works great, except for one thing: if I click the button, move the mouse outside it and then release, the button steals focus. This prevents e.g. Esc from exiting fullscreen - which is nice that it works too!

Nicely spotted. That's sort of an existing 'feature' here. With two windows the mouse clicks will switch focus on the other one based on the window's input region. I'll perhaps leave adjustments to that to further exercise, here the esc handling can be remedied by just pressing mouse again.

PR #1105, opened 2025-03-13
[sailfish-browser] Fix some favorite item list glitches. JB#63270

pvuorela:
The FavoriteGrid (i.e. bookmarks) was having footer item applied too often:
- with keyboard open it was extracted from visible space, possibly hiding some bookmarks
- on landscape it was making the view needlessly scrollable, and not covering the bottom of the screen. Easily seen how scroll decorators moved only part of the view.
- With something written on search field it was making the filtered bookmark row scrollable over the search field while keeping history list in place.

Made footer to be used now only on either favorite grid or history list, not on both. For cosmetics, non-UI BookmarkModel model was moved out from the middle of UI parts.

### DIFF
--- a/apps/browser/qml/pages/components/FavoriteGrid.qml
+++ b/apps/browser/qml/pages/components/FavoriteGrid.qml
@@ -17,7 +17,8 @@ import Sailfish.Browser 1.0
 IconGridViewBase {
     id: favoriteGrid
 
-    property real menuHeight: 0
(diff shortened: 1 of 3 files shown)


mlehtima (approved):

pvuorela:
Added a oneliner commit for not dimming history button while dragging.

direc85 (approved):

PR #1108, opened 2025-03-14
[browser] Remove oneshot scripts which shouldn't be needed anymore. JB#57609

pvuorela:
The oneshot dir hasn't been touched in three years. Any migration type of handling should be done already everywhere.


direc85 (approved):
Yeah, these are old enough for removal.

PR #1110, opened 2025-03-28
[browser] Indicate page permission exception 'always ask' as different from deny. JB#63322

pvuorela:
It's confusing when these two permission states show the same icon. Just looking at the permission popup one might interpret that the page has permission to something without questions asked.

Adding now extra icon at the bottom corner similar to denied case. The 'icon-s-maybe' doesn't exactly match the 'icon-s-blocked' shape for the template variants but it should be good enough for now.

### DIFF
--- a/apps/browser/qml/pages/PermissionExceptionsPage.qml
+++ b/apps/browser/qml/pages/PermissionExceptionsPage.qml
@@ -22,7 +22,7 @@ Page {
     property string iconSource
 
     property var remorse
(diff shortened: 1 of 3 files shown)


pvuorela:
Added another commit for having the permission value shown in exception list for 'always ask' and allow to modify it symmetrically in page settings and exception list.

spiiroin (approved):

direc85 (approved):

PR #1111, opened 2025-03-31
[browser] Allow landscape content on cover. Fixes JB#35932

rainemak:
Avoid relayouting browser content. Now that landscape home can be used also with smaller screen sizes previous conditions make very little sense. Thus, use Orientation.LandscapeMask and Orientation.Portrait as default orientations.


pvuorela (approved):
The phone landscape switcher does use portrait shaped app items. Currently the relayouting makes the web page content look more proper there. Then again simplicity and avoiding state changes is also nice.

For the mentioned bug I wouldn't think this has much effect since on a tablet these changes shouldn't make any difference.

Not objecting much either so I'll just make this approved, up to you.

PR #1112, opened 2025-04-03
[browser] Fix url field not always visible after creating new tab. JB#63325

pvuorela:
When opening a new tab, the search field gets focused and its 'edited' property turned true. After loading a new page and reopening overlay the showHistoryButton property, depending on edited, was fluctuating, and as result the favoriteGrid header height changed, and opacityY changed. Finally resetUrl() turning edited false.

Avoiding here now unnecessary changes on 'edited' when nothing has really been manually edited.

The onFocusChanged part was added in commit 39bbe455638b4f / JB#29424 but I'm not able to reproduce the problem these days with the snippet removed.

### DIFF
--- a/apps/browser/qml/pages/components/Overlay.qml
+++ b/apps/browser/qml/pages/components/Overlay.qml
@@ -336,10 +336,13 @@ Shared.Background {
 
                 property bool edited
                 property bool enteringNewTabUrl
+                property bool _resetting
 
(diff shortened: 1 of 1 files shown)


spiiroin (approved):

direc85 (approved):
Reproduced on device, and confirming the fix works.

PR #1113, opened 2025-04-03
[browser] Add option to start browser in private browsing mode. Fixes JB#63344

rainemak:


apps/browser/qml/pages/SettingsPage.qml
```diff
@@ -155,6 +155,19 @@ Page {
                 onClicked: closeAllTabsConfig.value = !closeAllTabsConfig.value
             }
 
+            TextSwitch {
+                //: Start browser in private browsing mode.
```
pvuorela: Just duplicating the engineering english doesn't provide any value. Better just drop this.

pvuorela (approved):
I'm a bit skeptical how useful this is when the 'new private tab' is quite easily available after launching from the hamburger menu. Saves one tap when the toolbar can be directly tapped to open the overlay.

But I'll just approve. The settings isn't that crowded yet. One comment for pointless translation hint.

PR #1117, opened 2025-08-01
Allow PortraitInverted page orientation

ichthyosaurus:
Allow the browser to be opened in inverted portrait orientation. Both landscape orientations are allowed anyway, so I don't see any reason to block inverted portrait.

### DIFF
--- a/apps/shared/BrowserWindow.qml
+++ b/apps/shared/BrowserWindow.qml
@@ -26,8 +26,8 @@ ApplicationWindow {
     property QtObject webView
     property alias activityDisabledByMdm: mdmView.activity
 
-    allowedOrientations: defaultAllowedOrientations
-    _defaultPageOrientations: Orientation.LandscapeMask | Orientation.Portrait
+    allowedOrientations: Orientation.All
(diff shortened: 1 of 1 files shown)


pvuorela:
Quoting silica defaultAllowedOrientations documentation:

    The value depends on the device configuration. Typically a tablet device
    would include all orientations but a phone would exclude the inverted
    portrait orientation.

On a phone normally no apps are using the inverted portrait and browser shouldn't be a special case.

ichthyosaurus:
Ah well, I didn't see that in the docs. I use inverted portrait a lot when developing because I don't want to bend the USB cable too much when it's plugged into my laptop but the USB port is at the bottom of the phone:

cable

pvuorela:
If it's your development thing, maybe you could change the default orientation for all apps locally instead of changing the behavior for everyone on app one, https://forum.sailfishos.org/t/howto-enable-all-orientations-including-portrait-inverted-for-apps/4349

ichthyosaurus:
Nice, thank you for the link!

pvuorela:
I'll close this now. Hoping the dconf trick helps your case.

PR #1118, opened 2025-08-01
Use dialog instead of button to clear browser data

ichthyosaurus:
- Use dialog instead of button to clear browser data: This change makes the UI more Sailfish-y.
- Disable combo box when not deleting history: It's confusing to be able to change how much should be deleted when nothing should be deleted at all.
- ~~Only clear cookies and cache by default: Keep history, passwords, bookmarks, and site permissions by default. When deleting data for privacy, it's unexpected to delete passwords etc.~~
- When deleting data, the last used settings are remembered instead of always selecting defaults.

**Before:**

d1
d2

**After:**

d3
d4

### DIFF
--- a/apps/browser/qml/pages/PrivacySettingsPage.qml
+++ b/apps/browser/qml/pages/PrivacySettingsPage.qml
@@ -28,7 +28,7 @@ Dialog {
                || clearBookmarks.checked
                || clearSitePermissions.checked
     acceptDestination: Qt.resolvedUrl("components/PrivacySettingsConfirmDialog.qml")
-    acceptDestinationAction: PageStackAction.Replace
+    acceptDestinationAction: PageStackAction.Push
 
     onAcceptPendingChanged: {
         acceptDestinationInstance.historyEnabled = clearHistory.checked
--- a/apps/browser/qml/pages/PrivacySettingsPage.qml
+++ b/apps/browser/qml/pages/PrivacySettingsPage.qml
@@ -15,11 +15,31 @@ import Sailfish.Silica 1.0
 import Nemo.Configuration 1.0
 import Sailfish.Browser 1.0
 
-Page {
+Dialog {
     id: page
 
     property var remorse
     property var previousPage
+
+    canAccept: clearHistory.checked
+               || clearCookiesAndSiteData.checked
+               || clearSavedPasswords.checked
+               || clearCache.checked
+               || clearBookmarks.checked
+               || clearSitePermissions.checked
+    acceptDestination: Qt.resolvedUrl("components/PrivacySettingsConfirmDialog.qml")
+    acceptDestinationAction: PageStackAction.Replace
+
+    onAcceptPendingChanged: {
+        acceptDestinationInstance.historyEnabled = clearHistory.checked
+        acceptDestinationInstance.cookieAndSiteDataEnabled = clearCookiesAndSiteData.checked
+        acceptDestinationInstance.passwordsEnabled = clearSavedPasswords.checked
+        acceptDestinationInstance.cacheEnabled = clearCache.checked
+        acceptDestinationInstance.bookmarksEnabled = clearBookmarks.checked
+        acceptDestinationInstance.sitePermissionsEnabled = clearSitePermissions.checked
+        acceptDestinationInstance.historyPeriod = historyErasingComboBox.currentItem.period
+        acceptDestinationInstance.acceptDestination = previousPage
+    }
 
     SilicaFlickable {
         anchors.fill: parent
@@ -30,10 +50,13 @@ Page {
 
             width: parent.width
 
-            PageHeader {
+            DialogHeader {
                 //: Clear private data page header
                 //% "Clear private data"
                 title: qsTrId("settings_browser-ph-clear_private_data")
+
+                //% "Clear"
+                acceptText: qsTrId("sailfish_browser-he-clear")
             }
 
             Column {
@@ -125,38 +148,6 @@ Page {
                     //% "Site permissions"
                     text: qsTrId("settings_browser-la-clear_site_permissions")
                     checked: true
-                }
-
-                // Spacer between Button and switches
-                Item {
-                    width: parent.width
-                    height: Theme.paddingLarge
-                }
-
-                Button {
-                    //: Button for clearing selected private data items.
-                    //% "Clear"
-                    text: qsTrId("settings_browser-bt-clear")
-                    anchors.horizontalCenter: parent.horizontalCenter
-                    enabled: clearHistory.checked
-                             || clearCookiesAndSiteData.checked
-                             || clearSavedPasswords.checked
-                             || clearCache.checked
-                             || clearBookmarks.checked
-                             || clearSitePermissions.checked
-
-                    onClicked: {
-                        var page = pageStack.push(Qt.resolvedUrl("components/PrivacySettingsConfirmDialog.qml"), {
-                                                      historyEnabled: clearHistory.checked,
-                                                      cookieAndSiteDataEnabled: clearCookiesAndSiteData.checked,
-                                                      passwordsEnabled: clearSavedPasswords.checked,
-                                                      cacheEnabled: clearCache.checked,
-                                                      bookmarksEnabled: clearBookmarks.checked,
-                                                      sitePermissionsEnabled: clearSitePermissions.checked,
-                                                      historyPeriod: historyErasingComboBox.currentItem.period,
-                                                      acceptDestination: previousPage
-                                                  })
-                    }
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


pvuorela:
I'm a bit split whether to keep using the Page or Dialog. Indeed we often use dialogs on this kind of places but then again page gets a bit more leeway of adding different removal options and details in the page. E.g. maybe access to history for deleting individual sites.

One side-effect here is that now getting to confirmation page, then noticing that the deletion options are wrong, stepping back -> view is on the main settings view instead of the deletion options.

For the defaults, I'd maybe keep deleting last 24h on. For others could be indeed better if the app is not that eager to delete bookmarks or saved passwords etc.

ichthyosaurus:
Thanks for taking the time to review!

> but then again page gets a bit more leeway of adding different removal options and details in the page. E.g. maybe access to history for deleting individual sites.

How do you mean? You can still push a regular page or another dialog to the stack from a dialog.

> One side-effect here is that now getting to confirmation page, then noticing that the deletion options are wrong, stepping back -> view is on the main settings view instead of the deletion options.

Yes but that can be changed. I wasn't sure which is better so I stuck with this way. Like, getting to the confirmation page, then noticing that I don't want to delete anything and stepping back to abort should bring me to the main settings view. Should I change it?

> For the defaults, I'd maybe keep deleting last 24h on.

I'll turn it back on.

pvuorela:
> How do you mean? You can still push a regular page or another dialog to the stack from a dialog.

You can but I think it could be slightly confusing for the user. 

> Yes but that can be changed. I wasn't sure which is better so I stuck with this way. Like, getting to the confirmation page, then noticing that I don't want to delete anything and stepping back to abort should bring me to the main settings view. Should I change it?

I think backstepping from such a dialog should take to the previous view. It can very well be that the deletion options were set wrong.

ichthyosaurus:
> I think backstepping from such a dialog should take to the previous view. It can very well be that the deletion options were set wrong.

I changed it in [c6b7c5e](https://github.com/sailfishos/sailfish-browser/pull/1118/commits/c6b7c5e537d66c6bc23f42291ca7be34f878a8af) so that you can go back to the previous dialog.

> You can but I think it could be slightly confusing for the user.

Not necessarily, I'd say. I guess it depends on the situation. Maybe we can switch to a dialog now, and if it turns out that with new features this becomes confusing or impractical, you can revert to the old implementation with a page and a button?

ichthyosaurus:
What's the status with this one? I think especially with https://github.com/sailfishos/sailfish-browser/pull/1118/commits/25cd794bb468fd43137d9279cbcd96bacacec694 it makes sense to have this merged :).

apps/browser/qml/pages/PrivacySettingsPage.qml
```diff
@@ -15,25 +15,60 @@ import Sailfish.Silica 1.0
...
     property var previousPage
 
+    canAccept: clearHistory.checked
+               || clearCookiesAndSiteData.checked
+               || clearSavedPasswords.checked
+               || clearCache.checked
+               || clearBookmarks.checked
+               || clearSitePermissions.checked
+    acceptDestination: Qt.resolvedUrl("components/PrivacySettingsConfirmDialog.qml")
+    acceptDestinationAction: PageStackAction.Push
```
pvuorela: Should be already the default.

apps/browser/qml/pages/PrivacySettingsPage.qml
```diff
@@ -15,25 +15,60 @@ import Sailfish.Silica 1.0
...
+
+    onAcceptPendingChanged: {
```
pvuorela, line 32: We probably don't want these actions when property is changed to false. Could protect following code inside "if (acceptPending)"

apps/browser/qml/pages/PrivacySettingsPage.qml
```diff
@@ -15,25 +15,60 @@ import Sailfish.Silica 1.0
...
+        acceptDestinationInstance.cacheEnabled = clearCache.checked
+        acceptDestinationInstance.bookmarksEnabled = clearBookmarks.checked
+        acceptDestinationInstance.sitePermissionsEnabled = clearSitePermissions.checked
+        acceptDestinationInstance.historyPeriod = historyErasingComboBox.currentItem.period
+        acceptDestinationInstance.acceptDestination = previousPage
+    }
+
+    ConfigurationGroup {
+        id: config
+        path: "/apps/sailfish-browser/actions"
```
pvuorela: Per coding conventions have an empty line after the "id:". Same applies on some other places. I'd maybe also have explicitly removal type of container path for these instead of generic 'actions'.

apps/browser/qml/pages/PrivacySettingsPage.qml
```diff
@@ -45,15 +80,26 @@ Page {
...
                     text: qsTrId("settings_browser-la-clear_history")
-                    checked: true
 
                     //: Description for clearing history. This will clear history and tabs.
                     //% "Clears history and open tabs"
                     description: qsTrId("settings_browser-la-clear_history_description")
+
+                    checked: config.clear_history
+                    automaticCheck: false
+                    onClicked: config.clear_history = !config.clear_history
```
pvuorela, line 93: If dconf used, pondering should it rather save the selection there when accepting (or entering?) the confirmation page.

pvuorela (commented):
Was torn whether to have a dialog or not but with the polishing done here I'm ok this way too. And definitely the deletion defaults should be adjusted.

Saving the deletion options to dconf is not a bad idea, but for simplicity I'd maybe lean towards just doing better defaults. E.g. who wants to delete their bookmarks on a regular basis? Removing passwords I wouldn't either expect happening all the time. If those are used on some rare occasion, it could be still even strange to offer them the next time.

Although I do understand if you want to just delete cookies then needing to uncheck history items is extra work. Maybe we should just have everything disabled by default? Or by default the last 24h history, which on cookie case would require one tap to uncheck.

A couple small comments in the code. Think the commits could be squashed a bit too as the changes go back and forth.

ichthyosaurus:
Sorry for the delay, I'll get back to this as soon as I find a bit more free time.

apps/browser/qml/pages/PrivacySettingsPage.qml
ichthyosaurus: removed it in [5470ff0](https://github.com/sailfishos/sailfish-browser/pull/1118/commits/5470ff0cccd3cbe2224ca76a2661e3df442fbb4f)

apps/browser/qml/pages/PrivacySettingsPage.qml
ichthyosaurus: Fixed in [55c53b5](https://github.com/sailfishos/sailfish-browser/pull/1118/commits/55c53b5856e63877cd498041d549e4ad2dd81862)

apps/browser/qml/pages/PrivacySettingsPage.qml
ichthyosaurus: > Per coding conventions have an empty line after the "id:". Same applies on some other places

fixed in [48b5442](https://github.com/sailfishos/sailfish-browser/pull/1118/commits/48b5442b9f06cc9fb6224d2db3c01dee3cd732a2)

> I'd maybe also have explicitly removal type of container path for these instead of generic 'actions'.

I changed the folder to "privacy-cleanup" in [26919a7](https://github.com/sailfishos/sailfish-browser/pull/1118/commits/26919a7ee7efc4d7d865a4ec4fc82bbcf5ac1fee)

My thought process regarding "actions" was (ef4290c3076d97487acf2800e7f161f7d4a33093):

> This reuses config keys that were no longer in use (but not removed since 68b52fef7fd89c12ddfe423f1b36a075acff3a91. I figured it's better to reuse existing keys instead of polluting dconf further.

"privacy-cleanup" is definitely clearer than "actions".

apps/browser/qml/pages/PrivacySettingsPage.qml
ichthyosaurus: I'd rather keep it like it is because it makes the code easier to read if all relevant changes are in one place. I don't think it matters much from a user perspective.

ichthyosaurus:
> with the polishing done here I'm ok this way too

Thanks! :)

I addressed your comments above. I hope everything is ok now.

> Think the commits could be squashed a bit too as the changes go back and forth.

If it's ok for you I'd prefer to keep it a little untidy because I don't feel too comfortable with editing commits in git rebase -i. I lost my work one too many times :D.

pvuorela:
> If it's ok for you I'd prefer to keep it a little untidy because I don't feel too comfortable with editing commits in git rebase -i. I lost my work one too many times :D.

Well, adjusting the commits should be pretty standard procedure while developing. If you feel unsure you can always create local tags and branches you can compare and get back to if needed.

Having that said, at this point I'd rather also just get this in so maybe I'll just merge it as it is.

pvuorela (approved):
Seems to do the trick.

ichthyosaurus:
Thanks for merging :)

PR #1119, opened 2025-08-01
Fix menu icon in toolbar moving when switching to an "about:" page

ichthyosaurus:
As the title says: the right margin of the toolbar changed when viewing e.g. "about:blank#".

**Normal page:**

m1

**Before:**

m2

**After:**

m3


pvuorela (approved):
Seems to do the trick and indeed the visibility should be rather controlled via the 'active' property.

pvuorela:
Hmh, tried to switch the target to 'next' which is the real development head but this doesn't look ideal now with 57 commits.

Could you maybe rebase your changes on the next branch.

ichthyosaurus:
I've rebased it, and the other PRs too.

pvuorela:
Looking good, thanks!

PR #1120, opened 2025-08-01
Add menu button to quickly clear browsing data

ichthyosaurus:
This adds a menu button to quickly clear browsing data without having to open the settings page first.

**Unexpanded menu:**

b1

**Before:**

b2

**After:**

b3

### DIFF
--- a/apps/browser/qml/pages/components/PopUpMenuItem.qml
+++ b/apps/browser/qml/pages/components/PopUpMenuItem.qml
@@ -212,6 +212,21 @@ Item {
                 height: Theme.itemSizeSmall
                 iconWidth: root.iconWidth
                 horizontalOffset: root.horizontalOffset
+                //: The label for the button for accessing clear browsing data page
+                //% "Clear browsing data"
+                text: qsTrId("settings_browser-la-clear-browsing-data")
+                iconSource: "image://theme/icon-m-delete"
+
+                onClicked: {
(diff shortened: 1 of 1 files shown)


pvuorela:
The menu is already a bit crowded with items so there should be big enough need for adding something new. I'm not sure if removing browsing data has such a need. There is the private mode, and for clearing this saves only one tap. Shouldn't be that common need to go clearing the browsing data?

ichthyosaurus:
I often clear my browsing data because I want to get rid of cookies but I also want to keep the history (that's why I don't use private mode). (Also, there's no official way to always start in private mode but that's a different issue.)

That takes 7 taps and 2 swipes without #1118 and without #1120, or 6 taps and 2 swipes with #1120, or 2 taps and 3 swipes with #1118 and #1120.

pvuorela:
I guess you are still in minority in your way of clearing cookies. The shortcut you would get is a UI complexity addition for others. Small one but they accumulate over features. Out of the taps and swipes, this particular should be saving only one tap, I think.

Maybe easier savings on the clearing view, defaulting to everything feels indeed excessive. More on that on the other PR.

ichthyosaurus:
Should I close this or are you still considering it?

pvuorela:
Let's skip this one for now as it's only saving that one tap. Thanks for the suggestion though!

PR #1121, opened 2025-09-30
Remember Find-in-Page search terms

nephros:
Currently it's a bit annoying that the search field is cleared every time you do a on-page search (especially if nothing was found!).

I think the use cases are clear: I may want to do variations on my previous search term , I may have made a spelling mistake etc.


rainemak (approved):
This seems to be exactly the behavior that desktop Firefox has. I'm happy with this.

At first I thought that we should clear the ```lastFindText``` upon tab / url change but that's not really needed.

It is good as it is. :+1:

PR #1132, opened 2026-07-22
[browser] Add menu option to view page source

nephros:
I know the menu is a bit cluttered already.

Another place could be the certficate info page, but it doesn't quite fit...

### DIFF
--- a/apps/browser/qml/pages/components/PopUpMenuItem.qml
+++ b/apps/browser/qml/pages/components/PopUpMenuItem.qml
@@ -158,6 +158,24 @@ Item {
                     }
                 }
             }
+
+            OverlayListItem {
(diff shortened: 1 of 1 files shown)


nephros:
I could add some conditions to disable menu functions that don't quite apply
(like desktop version and add to app grid) but wanted to keep it simple for
now.

nephros:
Using this does not update the tab title. Not sure ho to do that.

pvuorela:
Yea, indeed the menu is already having too much content and it's not likely in general case that common to go looking at the html source. For comparison on Android Firefox it seems to require installing an extension even though otherwise they are having quite a bit of features available directly.

I don't have good immediate suggestions. I'm not entirely objecting on having this (it's cheap to implement), but then again it shouldn't clutter the menu nor settings.

PR #1133, opened 2026-07-24
[feat] Add RSS feed action to Browser menu. Fixes JB#18041

abranson:
Depends on sailfish-account-rss, plus Settings and Sailjail changes

### DIFF
--- a/apps/browser/qml/pages/components/PopUpMenuItem.qml
+++ b/apps/browser/qml/pages/components/PopUpMenuItem.qml
@@ -124,6 +124,24 @@ Item {
             }
 
             OverlayListItem {
+                visible: webView.contentItem && webView.contentItem.feedUrl
+                height: visible ? Theme.itemSizeSmall : 0
+                iconWidth: root.iconWidth
+                horizontalOffset: root.horizontalOffset
+                iconSource: "image://theme/icon-m-add"
+                //% "Add feed"
+                text: qsTrId("sailfish_browser-la-add_feed")
+
+                onClicked: {
+                    overlay.animator.showChrome()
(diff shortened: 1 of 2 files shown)


pvuorela:
Wondering if RSS is still a thing that we want supported in the broser. Definitely the popularity is not what it used to be and the other browser have generally dropped the support for what I gather.

If we want this supported, could it be done without browser? Dependency here to such optional external support can be problematic.

abranson:
I get asked about it a lot - it's still the best way to get news without being sucked into third-party services and algorithms. It doesn't really need the browser at all - it can be added from the accounts UI like an ICS calendar by pasting a URL. But this makes it easier.

This could even check for support before offering, with the new fileservice query method.

nephros:
> the other browser have generally dropped the support for what I gather

RSS support (as in parsing/display) in the Browser itself is really unnecessary.

BUT.

I have tested @abranson's RSS Account Integration I must say it's *tremendously* useful for several reasons, e.g.:

 - it lessens the need for dedicated account/libsocial/buteo integrations (e.g. all the various Fediverse stuff).

 - can be very useful in development and related tasks as you can subscribe to e.g. OBS events or things like Github/Codeberg notifications.

 - also the Forum has RSS support per-category, so you can e.g. subscribe to just "Announcements" or "Blog".

So even if Big Tech and Media and friends may have removed RSS support, many many other services still have it, and I believe these services are rather interesting to the current SFOS user demography.

For this reason, I think having the capability to add an "RSS Account" from the Browser would be useful.  
Whether that is done through an actual button, or some other way that allows a "Subscribe" action for a given URL (e.g. long-press on link?) I cant say.

PR #1135, opened 2026-08-01
[Browser] Add reload action to cover

nephros:
Oneliner change, reload current page/tab from cover.

### DIFF
--- a/apps/browser/qml/pages/BrowserPage.qml
+++ b/apps/browser/qml/pages/BrowserPage.qml
@@ -395,6 +395,10 @@ Page {
             iconSource: "image://theme/icon-cover-new"
             onTriggered: activateNewTabView()
         }
+        CoverAction {
+            iconSource: "image://theme/icon-cover-refresh"
+            onTriggered: webView.reload()
+        }
     }
 
     Connections {


pvuorela:
Back in ancient times we used to have this but then for sailfish 2.x we removed it in the bigger cover action overhaul where they got limited to one item (commit dc71d69cb). I can understand sometimes wanting to do this but I'm not sure is that enough to warrant special casing this application to have more actions than the rest.

nephros:
> Back in ancient times we used to have this but then for sailfish 2.x we removed it in the bigger cover action overhaul where they got limited to one item (commit [dc71d69](https://github.com/sailfishos/sailfish-browser/commit/dc71d69cb2d77d84fef861692710c7eab59b8d67)).

So is it a general design guideline to have only one action on the cover?

pvuorela:
> So is it a general design guideline to have only one action on the cover?

Yes.

nephros:
I see. Didn't know that :)

Would you accept a Setting (under Appearance) to switch this action? (Perhaps adding a 'Open new Private Tab' as an alternative as well?)

pvuorela:
> Would you accept a Setting (under Appearance) to switch this action? (Perhaps adding a 'Open new Private Tab' as an alternative as well?)

Sorry, but I'm slightly reluctant on adding more to choose from. Thinking many people sometimes wanting this might still find the new tab action desirable.

The private tab should already work with the existing action if the browser is in the private mode.

nephros:
Fine!  

No need to feel sorry, it's why I ask!

PR #1143, opened 2026-09-04
[browser] Upgrade Sailfish Browser from Gecko ESR115 to ESR153. Fixes JB#64963

abranson:
Upgrade Browser from ESR115 to ESR153, mostly involving the move to hosted tabs and Gecko’s Fission process model.
- Replace legacy page objects and queues with hosted logical-tab sessions and Qt Quick rendering for normal, private and captive-portal browsing.
- Adapt tab/history persistence and restoration; keep private sessions and thumbnails in memory.
- Route navigation, prompts, selection, fullscreen, downloads, thumbnails and gestures through hosted sessions.
- Handle content crashes, lifecycle and rotation; keep pages visible behind overlays.
- Keep context menus and select pickers attached to the originating tab/document, including cancellation on navigation.
- Update stack dependencies and add migration, session, navigation and popup regression tests.


apps/browser/bookmarks/declarativebookmarkmodel.h
```diff
@@ -26,6 +26,7 @@ class DeclarativeBookmarkModel : public QAbstractListModel
     Q_PROPERTY(bool activeUrlBookmarked READ activeUrlBookmarked NOTIFY activeUrlBookmarkedChanged FINAL)
 public:
     DeclarativeBookmarkModel(QObject *parent = 0);
+    ~DeclarativeBookmarkModel() override;
```
pvuorela, line 29: I think override on dtor is a bit silly. It's not really overriding anything but adding something to be executed, and should be obvious enough how ctor/dtors work.

apps/browser/settings/searchenginemodel.h
```diff
@@ -37,9 +35,8 @@ class SearchEngineModel : public QAbstractListModel, public QQmlParserStatus
     };
 
     explicit SearchEngineModel(QObject *parent = 0);
-    virtual ~SearchEngineModel();
+    ~SearchEngineModel() override;
```
pvuorela, line 38: Same, won't comment any further ones.

apps/browser/bookmarks/declarativebookmarkmodel.h
abranson: Afaik this should always be done when overriding all virtual functions, including destructors, as it's important to catch at compile-time if they become non-virtual in the future. Aside from these two, there's only one other in SecureAction, and that's been there since it was written.

pvuorela: There is no right rule for that but rather different explicit or implicit coding conventions, and great chance for getting into bikeshedding. 

We don't have any explicit convention so far and didn't find anything in Qt side either. But in practice we haven't really done this anywhere even if override keyword is otherwise used a lot. Also unlikely that QObject ever gets rid of the virtual dtor :)

Googling quickly for usage elsewhere, there's been some warnings on such usage https://learn.microsoft.com/en-us/cpp/code-quality/c26443 

Having said that, I do acknowledge that some people like it.

